PR #307, opened 2021-07-04
Drop the windows 20h2 docker image

matt-richardson:
As we're not using it, and having trouble with docker on - it's not stable

# Background

We started adding support for docker images for windows 20H2, but 20H2 is causing more pain than it's worth. It never made it out the door.
Dropping it to reduce complexity.

# Review

A simple sanity check of the idea will suffice.

# Checks


cailyoung (approved):
I was initially concerned that we might be breaking a composite build by removing a build target, but I checked, and each individual OS builds a specific tag in the compose file - so this is OK to remove this way.

PR #312, opened 2021-08-20
Remove build_id files from rpm

rain-on:
# Background
Specifically on Fedora/RHEL targets, the OctoCLI RPM clashes with Tentacle's RPM, as they both contain a symbolic link under /user/lib/build-id with the same name, but different targets (OctoCLI and Tentacle respectively).

This allows the RPM system to associate debug info/symbols stored on the system with a specific version of a specific app.

It appears the dotnet compiler works in such a way that these build-ids can collide.

This collision inhibits the install of the latter app (First in best dressed).

It may be resolved by setting `rpm --replacefiles` - however, it brings into question the value of the BuildId if conflicts are possible.

Sets a flag on fpm, to prevent the underlying rpm utility from installing a Build Id link, this may impact the ability to link debug information installed on the host with the executable which produced them - however this loss doesn't appear to impact our existing workflows.

# Results

## Before
image
Package contents with `usr/lib`

## After
image
Package content with no `usr/lib`

Following installation of both Octo and Tentacle into a Fedora core docker image you can see:
* Octo and Tentacle have matching Build-ID as extracted from the executable image (implying both have installed successfully)
* A single entry in the build-id directory points to Octo.exe
image


tothegills (approved):
Google says it's ok. Easy enough to roll forward out of if it causes problems.

PR #319, opened 2021-08-31
Ensure Tentacle binary is linked on new container

matt-richardson:
Even if the "i am configured" semaphore exists

# Background

If the `.configuredSemaphore` file exists, the Tentacle container wont attempt to reconfigure.

In https://github.com/OctopusDeploy/OctopusTentacle/pull/290, we changed it from completely aborting to "ignore reconfiguration and continue". 

However, in the wild (ie, on my nas), the new container, I found that sometimes it still would be unable to launch, as the binary had not been linked into the expected location.

This PR changes the order, so that if the symbolic link does not exist, it creates that before checking if the semaphore exists.

# Testing

```
docker run `
  --interactive `
  --detach `
  --name OctopusTentacle10996 `
  --publish 10996:10996 `
  --env "ListeningPort=10996" `
  --env "ServerApiKey=API-XXXX" `
  --env "TargetEnvironment=Home" `
  --env "TargetRole=container-server" `
  --env "ServerUrl=https://myoctopus.octopus.app" `
  --env "ACCEPT_EULA=Y" `
  --env "DISABLE_DIND=Y" `
  --restart always `
  -v C:\temp\testing-link-tentacle-binary-order:/etc/octopus `
  docker.packages.octopushq.com/octopusdeploy/tentacle:6.1.940-mattr-link-tentacle-binary-order
```
Then I connected to the container, installed `ps` (via `apt-get install procps`) and killed the tentacle process. Docker restarted the container, _interestingly with the same id_.

I then stopped the container, deleted it, and re-ran the command above - it successfully re-launched.

I believe that this second step was what was missed in #290.

# Review

A sanity check on the ideas and a quick :eyes: over the code will be sufficient.

# Checks

### DIFF
--- a/docker/linux/scripts/configure-tentacle.sh
+++ b/docker/linux/scripts/configure-tentacle.sh
@@ -15,14 +15,14 @@ alreadyConfiguredSemaphore="$configurationDirectory/.configuredSemaphore"
 mkdir -p $configurationDirectory
 mkdir -p $applicationsDirectory
 
+if [ ! -f /usr/bin/tentacle ]; then
+	ln -s /opt/octopus/tentacle/Tentacle /usr/bin/tentacle
+fi
+
 if [ -f "$alreadyConfiguredSemaphore" ]; then
     echo "Octopus Tentacle is already configured. Skipping reconfiguration."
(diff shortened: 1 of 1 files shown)


uglybugger (approved):
Looks good to me 👍

PR #338, opened 2021-11-04
Replace harden script with executable

tothegills:
# Background

This PR replaces the Tentacle install folder hardening script with an executable that does the same job. Customers are reporting that Tentacle will not install because their security scanners are picking up the PowerShell script. This is an attempt at mitigating that problem while keeping the hardening script.

Fixes https://github.com/OctopusDeploy/OctopusTentacle/issues/342


source/Octopus.Tentacle.Hardener/Program.cs
```diff
@@ -0,0 +1,50 @@
...
+            var currentDirectory = Directory.GetCurrentDirectory();
+            RemoveInheritanceWhilePreservingExistingEntries(currentDirectory);
+            RemoveWriteAccessRulesForBuiltInUsers(currentDirectory);
+        }
+
+        static void RemoveInheritanceWhilePreservingExistingEntries(string directory)
+        {
+            var accessControl = Directory.GetAccessControl(directory, AccessControlSections.Access);
+            accessControl.SetAccessRuleProtection(true, true);
+            Directory.SetAccessControl(directory, accessControl);
```
MissedTheMark, line 21: 👍

MissedTheMark (approved):
Looks good.  Replicates the PowerShell hardening script nicely 👍 🎉

PR #341, opened 2021-11-08
Add PackageRetentionJournal environment variable

N-lson:
# Background

Package retention requires a new journal file to be stored on the Tentacle to keep track of package usages. This PR adds a new environment variable similar to the existing `JournalPath` used for deployments which we can use in Calamari to retrieve the correct file.

### DIFF
--- a/source/Octopus.Tentacle.Tests/Commands/StubTentacleConfiguration.cs
+++ b/source/Octopus.Tentacle.Tests/Commands/StubTentacleConfiguration.cs
@@ -31,7 +31,7 @@ namespace Octopus.Tentacle.Tests.Commands
         public string PackagesDirectory { get; private set; }
         public string LogsDirectory { get; private set; }
         public string JournalFilePath { get; private set; }
-        public string PackageRetentionFilePath { get; private set; }
+        public string PackageRetentionJournalFilePath { get; private set; }
         public X509Certificate2 TentacleCertificate { get; set; }
         public string ListenIpAddress { get; set; }
         public bool NoListen { get; set; }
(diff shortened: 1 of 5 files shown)


source/Octopus.Tentacle.Tests/Commands/StubTentacleConfiguration.cs
```diff
@@ -31,6 +31,7 @@ public IEnumerable<string> TrustedOctopusThumbprints
         public string PackagesDirectory { get; private set; }
         public string LogsDirectory { get; private set; }
         public string JournalFilePath { get; private set; }
+        public string PackageRetentionFilePath { get; private set; }
```
tothegills, line 34: `PackageRetentionJournalFilePath`?

N-lson: Woops, yeah should be that

source/Octopus.Tentacle/Configuration/TentacleConfiguration.cs
```diff
@@ -97,10 +97,8 @@ public string PackagesDirectory
             get { return Path.Combine(ApplicationDirectory, "Packages"); }
         }
 
-        public string JournalFilePath
-        {
-            get { return Path.Combine(home.HomeDirectory ?? ".", "DeploymentJournal.xml"); }
-        }
+        public string JournalFilePath => Path.Combine(home.HomeDirectory ?? ".", "DeploymentJournal.xml");
+        public string PackageRetentionJournalPath => Path.Combine(home.HomeDirectory ?? ".", "PackageRetentionJournal.json");
```
N-lson, line 101: I went with `PackageRetentionJournalPath` instead of `PackageRetentionJournalFilePath` to match with the environment variable name even though the existing tentacle journal is called `JournalFilePath`. I'm open to using either name but `File` in the name doesn't seem necessary to me.

tothegills (approved):

PR #343, opened 2021-11-17
Adjustments to more Nullable Reference Type things for Configuration

slewis74:
# Background

This is to deal with changes in the later .NET SDKs that cause Visual Studio and Rider, and sometimes the builds, to hit errors.

# Review

👀 the code, there are only a few lines.

# Checks

### DIFF
--- a/source/Octopus.Tentacle/Configuration/TentacleConfiguration.cs
+++ b/source/Octopus.Tentacle/Configuration/TentacleConfiguration.cs
@@ -54,7 +54,7 @@ namespace Octopus.Tentacle.Configuration
         [Obsolete("This configuration entry is obsolete as of 3.0. It is only used as a Subscription ID where one does not exist.")]
         public string? TentacleSquid => settings.Get("Octopus.Communications.Squid", (string?)null);
 
-        public IEnumerable<OctopusServerConfiguration> TrustedOctopusServers => settings.Get(TrustedServersSettingName, new OctopusServerConfiguration[0]);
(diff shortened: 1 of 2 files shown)


source/Octopus.Tentacle/Configuration/TentacleConfiguration.cs
```diff
@@ -72,7 +72,7 @@ public string ApplicationDirectory
             get
             {
                 var path = settings.Get<string>(DeploymentApplicationDirectorySettingName);
-                if (string.IsNullOrWhiteSpace(path))
+                if (path is null || string.IsNullOrWhiteSpace(path))
```
johnsimons, line 75: is this required?

johnsimons: @slewis74 maybe we should consider only enabling nullability for netcore?

slewis74: Do you have an example of how to do that? I tried and it went back to possible null reference errors in the compile.

johnsimons: I just had a play with the code, and I am confused. It doesn't seem the projects have nullability explicitly enabled, which I thought it means it is NOT?
Also, I noticed `TentacleConfiguration.cs` file has it explicitly ON, why?

slewis74: because that's the pattern that has been used to progressively switch things over. If you switch it on at the project level a ton of things light up. My focus here was to unblock the current issue.

johnsimons (approved):
LGTM

PR #345, opened 2021-11-23
Pass isolation mutex name to workspace

tothegills:
# Background

This works in conjunctions with the changes to Octopus Shared that allows passing a script isolation mutex name with a command: https://github.com/OctopusDeploy/OctopusShared/pull/165
This PR sets the isolation mutex name on the workspace to match the name sent with the command.


N-lson:
Would adding another test to `ScriptServiceFixture` which uses `new StartScriptCommand()` instead of the builder be valuable? I noticed `BuiltInWorkerExecutionTarget` news up a command directly in OctopusServer, so having a test that checks for this may be nice.

PR #346, opened 2021-11-30
Passed the package details via the command line

droyad:
`Octo pack` does not [support nuspec](https://github.com/OctopusDeploy/OctopusClients/issues/330) files (not to be confused with OctoPack that does).

Fixes https://github.com/OctopusDeploy/ResearchAndDevelopment/issues/158

From the built package:
image


tothegills (approved):
👍

PR #350, opened 2022-01-05
Add Tenant/TenantTag/TenantedDeploymentParticipation options to Tentacle docker image, add documentation for Space/MachinePolicy options

N-lson:
# Background

Tentacle allows you to specify Tenants/TenantTags via the CLI. Currently these options don't exist when running a new Tentacle Docker container.

Fixes https://github.com/OctopusDeploy/Issues/issues/7251

This PR also adds some documentation for the `Space` and `MachinePolicy` environment variables - currently both of these can be configured for the Linux image, but only `Space` existed for the Windows image. Neither of these were included in the readme, so I've added some documentation for these as well.

## Before

Tentacles could not be created with Tenant options

### Linux
Default Tentacle with no extra options:
image

Tentacle with Tenant (`--env TargetTenant="T1"`):
image

Tentacle with Tenant Tag (`--env TargetTenantTag="TS1/TT1"`):
image

Tentacle with Tenanted Deployment Mode (`--env TargetTenantedDeploymentParticipation="TenantedOrUntenanted"`):
image

Tentacle with Space (`--env Space="Space2"`):
image

Tentacle with Machine Policy (`--env MachinePolicy="TestMachinePolicy"`):
image

Tentacle with a bunch of options (`--env TargetRole="web" --env TargetEnvironment="Dev,Test" --env TargetTenant="T2,T3" --env TargetTenantTag="TS1/TT1" --env MachinePolicy="MP2" --env Space="Space2" --env TargetTenantedDeploymentParticipation="TenantedOrUntenanted"`):
image
image

### Windows
Tentacle with a bunch of options (`--env TargetRole="web" --env TargetEnvironment="Development,Test" --env TargetTenant="T2,T3" --env TargetTenantTag="TTS1/TT1" --env MachinePolicy="MP2" --env Space="Space2" --env TargetTenantedDeploymentParticipation="TenantedOrUntenanted"`):
image
image

# Review

# Checks

### DIFF
--- a/docker/linux/scripts/configure-tentacle.sh
+++ b/docker/linux/scripts/configure-tentacle.sh
@@ -72,6 +72,13 @@ function validateVariables() {
 		fi
     fi
 
+	if [[ ! -z "$TargetTenant" || -z "$TargetTenantTag" ]]; then
+		if [[ "$TargetTenantedDeploymentParticipation" != "Tenanted" || "$TargetTenantedDeploymentParticipation" != "TenantedOrUntenanted" ]]; then
+			echo "The 'TargetTenant' and 'TargetTenantTag' environment variables are not valid when the 'TargetTenantedDeploymentParticipation' variable is not set to 'Tenanted' or 'TenantedOrUntenanted'"
+			exit 1
+		fi
+	fi
+
     echo " - server endpoint '$ServerUrl'"
     echo " - api key '##########'"
   if [[ ! -z "$ServerPort" ]]; then
@@ -142,15 +149,15 @@ function registerTentacle() {
 			done
 		fi
 
-		if [[ ! -z "$Tenant" ]]; then
-			IFS=',' read -ra TENANT <<< "$Tenant"
+		if [[ ! -z "$TargetTenant" ]]; then
+			IFS=',' read -ra TENANT <<< "$TargetTenant"
 			for i in "${TENANTS[@]}"; do
 				ARGS+=('--tenant' "$i")
 			done
 		fi
 
-		if [[ ! -z "$TenantTag" ]]; then
-			IFS=',' read -ra TENANTTAGS <<< "$TenantTag"
+		if [[ ! -z "$TargetTenantTag" ]]; then
+			IFS=',' read -ra TENANTTAGS <<< "$TargetTenantTag"
 			for i in "${TENANTTAGS[@]}"; do
 				ARGS+=('--tenanttag' "$i")
 			done
@@ -190,6 +197,10 @@ function registerTentacle() {
 
 	if [[ ! -z "$TargetName" ]]; then
 		ARGS+=('--name' "$TargetName")
+	fi
+	
+	if [[ ! -z "$TargetTenantedDeploymentParticipation" ]]; then
+		ARGS+=('--tenanted-deployment-participation' "$TargetTenantedDeploymentParticipation")
 	fi
 
 	tentacle "${ARGS[@]}"
--- a/docker/linux/scripts/configure-tentacle.sh
+++ b/docker/linux/scripts/configure-tentacle.sh
@@ -141,6 +141,20 @@ function registerTentacle() {
 				ARGS+=('--role' "$i")
 			done
 		fi
+
+		if [[ ! -z "$Tenant" ]]; then
+			IFS=',' read -ra TENANT <<< "$Tenant"
+			for i in "${TENANTS[@]}"; do
+				ARGS+=('--tenant' "$i")
+			done
+		fi
+
+		if [[ ! -z "$TenantTag" ]]; then
+			IFS=',' read -ra TENANTTAGS <<< "$TenantTag"
+			for i in "${TENANTTAGS[@]}"; do
+				ARGS+=('--tenanttag' "$i")
+			done
+		fi
 	fi
 
 	ARGS+=(
--- a/docker/linux/scripts/configure-tentacle.sh
+++ b/docker/linux/scripts/configure-tentacle.sh
@@ -92,6 +92,15 @@ function validateVariables() {
   if [[ ! -z "$TargetName" ]]; then
     echo " - name '$TargetName'"
   fi
+  if [[ ! -z "$TargetTenant" ]]; then
+    echo " - tenant '$TargetTenant'"
+  fi
+  if [[ ! -z "$TargetTenantTag" ]]; then
+    echo " - tenant tag '$TargetTenantTag'"
+  fi
+  if [[ ! -z "$TargetTenantedDeploymentParticipation" ]]; then
+    echo " - tenanted deployment participation '$TargetTenantedDeploymentParticipation'"
+  fi
   if [[ ! -z "$Space" ]]; then
     echo " - space '$Space'"
   fi
@@ -147,7 +156,7 @@ function registerTentacle() {
 		fi
 
 		if [[ ! -z "$TargetTenant" ]]; then
-			IFS=',' read -ra TENANT <<< "$TargetTenant"
+			IFS=',' read -ra TENANTS <<< "$TargetTenant"
 			for i in "${TENANTS[@]}"; do
 				ARGS+=('--tenant' "$i")
 			done
--- a/docker/linux/scripts/configure-tentacle.sh
+++ b/docker/linux/scripts/configure-tentacle.sh
@@ -11,6 +11,7 @@ instanceName=Tentacle
 configurationDirectory=/etc/octopus
 applicationsDirectory=/home/Octopus/Applications
 alreadyConfiguredSemaphore="$configurationDirectory/.configuredSemaphore"
+defaultListeningPort=10933
 
 mkdir -p $configurationDirectory
 mkdir -p $applicationsDirectory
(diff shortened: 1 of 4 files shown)


docker/windows/Scripts/configure-tentacle.ps1
```diff
@@ -146,6 +157,13 @@ function Validate-Variables() {
     }
   }
 
+  if($TargetTenant -ne $null -or $TargetTenantTag -ne $null) {
+    if($TargetTenantedDeploymentParticipation -ne "Tenanted" -or $TargetTenantedDeploymentParticipation -ne "TenantedOrUntenanted"){
+      Write-Error "The 'TargetTenant' and 'TargetTenantTag' environment variables are not valid when the 'TargetTenantedDeploymentParticipation' variable is not set to 'Tenanted' or 'TenantedOrUntenanted'"
```
tothegills, line 162: Same question as above re requiring participation to be set explicitly.

tothegills: Thinking about it a bit more, I wonder if the validation is inside out. If tenanted deployment participation is specified, ensure a tenant or tag has also been specified.

tothegills (commented):
Some questions/comments inline.

docker/windows/Scripts/configure-tentacle.ps1
tothegills: .. and some more. I wonder why we are performing this type of validation at all, instead of letting the register command do it?

N-lson: I didn't check the behavior of the register-with command - I saw that the script already had some validation and thought of including some for the tenant variables. But if the register command handles all this anyway, then I agree it doesn't seem like any validation would be needed in this script.

docker/windows/Scripts/configure-tentacle.ps1
```diff
@@ -135,6 +138,14 @@ function Validate-Variables() {
       Write-Error "The 'TargetRole' environment variable is not valid in combination with the 'TargetWorkerPool' variable"
       exit 1;
     }
+    if($TargetTenant -ne $null) {
```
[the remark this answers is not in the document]
N-lson: Yeah they should've been included in both linux and windows (although this may not be needed at all with what you mentioned in the other comment).

docker/windows/Scripts/configure-tentacle.ps1
N-lson: I think there's some validation in this script since the messages that are returned from the commands themselves don't match with what environment variables the user needs to provide via the docker command. Eg. if you don't provide `ServerUrl` (the docker environment variable name) with no validation the command returns this message:
```
Please specify an Octopus Server, e.g., --server=http://your-octopus-server
```
Whereas it currently says:
```
Missing 'ServerUrl' environment variable
```
The second one seems seems a bit clearer, otherwise the user would need to check what `server` is on the Docker image. Though I like the first error message more just in terms of how it reads.

There's another example which definitely should have some sort of validation, where if you don't specify a listening/server port you get:
```
Unrecognized command line arguments: False
```
which is a bit too unclear.

So I think certain validation should stay to help the user understand what they're missing, but checking the logic like whether tenant/tag exist with deployment participation can probably be left to the command itself.

N-lson: Actually it looks like the Windows docker image has a default listening port but the Linux one doesn't. So that case shouldn't require any validation, it's just that the Linux image is missing a default value.

docker/linux/scripts/configure-tentacle.sh
```diff
@@ -91,6 +92,9 @@ function validateVariables() {
   if [[ ! -z "$TargetName" ]]; then
     echo " - name '$TargetName'"
   fi
+  if [[ ! -z "$Space" ]]; then
+    echo " - space '$Space'"
+  fi
```
N-lson, line 97: Added this to match the Windows output

docker/linux/scripts/configure-tentacle.sh
```diff
@@ -103,7 +107,7 @@ function configureTentacle() {
 	if [[ ! -z "$ServerPort" ]]; then
 		tentacle configure --instance "$instanceName" --noListen "True"
 	else
-		tentacle configure --instance "$instanceName" --port $ListeningPort --noListen "False"
+		tentacle configure --instance "$instanceName" --port $defaultListeningPort --noListen "False"
```
N-lson, line 110: If a `ListeningPort` wasn't provided before for a Linux container it would fail here with an unhelpful error message. It should default to `10933`.

docker/windows/Scripts/configure-tentacle.ps1
```diff
@@ -254,6 +277,11 @@ function Register-Tentacle(){
     $arg += "`"$Space`"";
   }
 
+  if($null -ne $MachinePolicy) {
+    $arg += "--policy";
+    $arg += "`"$MachinePolicy`"";
+  }
+
```
N-lson, line 284: I added this to the Windows image since the Linux image does this already.

docker/readme.md
```diff
@@ -43,13 +43,18 @@ docker run --publish 10931:10933 --tty --interactive --env "ListeningPort=10931"
...
+- **Space**: The name of the space which the Tentacle will be added to. Defaults to the default space.
 - **TargetEnvironment**: Comma delimited list of environments to add this target to.
 - **TargetRole**: Comma delimited list of roles to add to this target.
-- **TargetWorkerPool**: Comma delimited list of worker pools to add to this target to (not to be used with environments or role variable).
+- **TargetWorkerPool**: Comma delimited list of worker pools to add this target to (not to be used with environment, tenant, tenant tag or role variable).
 - **TargetName**: Optional Target name, defaults to host.
+- **TargetTenant**: Comma delimited list of tenants to add to this target.
+- **TargetTenantTag**: Comma delimited list of tenant tags to add to this target.
+- **TargetTenantedDeploymentParticipation**: The tenanted deployment mode of the target. Allowed values are Untenanted, TenantedOrUntenanted and Tenanted. Defaults to Untenanted.
+- **MachinePolicy**: The name of the machine policy that will apply to this Tentacle. Defaults to the default machine policy.
```
N-lson, line 57: From the existing configuration, `MachinePolicy` and `Space` already exist for both images, and `MachinePolicy` exists for Linux. I've added policy to the Windows image, and I've added documentation for these two parameters. Unless there's a reason why we don't want to expose this configuration for some reason?

docker/windows/Scripts/configure-tentacle.ps1
```diff
@@ -169,11 +126,68 @@ function Validate-Variables() {
...
+  }
+  if($TargetTenantedDeploymentParticipation -ne $null) {
+    Write-Log " - tenanted deployment participation '$env:TargetTenantedDeploymentParticipation'"
+  }
   if($null -ne $Space) {
     Write-Log " - space '$Space'"
   }
 }
 
+function Configure-Tentacle
```
N-lson, line 143: Moved methods in here around to match the Linux script which has the methods in the order they're called

docker/linux/scripts/configure-tentacle.sh
tothegills: It looks like previously the port could be specified by setting `$ListeningPort`, is there a way to customise the port now?

N-lson: This line sets the port in the Tentacle's configuration, but when it does the `register-with` it [overrides it](https://github.com/OctopusDeploy/OctopusTentacle/pull/350/files/569e20ed37693d933ec5c5a61d2ee626e63e934e#diff-3d9f587a7deebe13ba8499e4940aaa4d4b4cb9a1c7181e73ec0eaf86a9f2bf4bR181) with `$ListeningPort` if it exists and is different:
```
		if [[ ! -z "$ListeningPort" && "$ListeningPort" != $defaultListeningPort ]]; then
			ARGS+=('--tentacle-comms-port' $ListeningPort)
		fi
```
It does this a similar thing in the Windows version already:
https://github.com/OctopusDeploy/OctopusTentacle/blob/main/docker/windows/Scripts/configure-tentacle.ps1#L47
https://github.com/OctopusDeploy/OctopusTentacle/blob/main/docker/windows/Scripts/configure-tentacle.ps1#L208

tothegills (commented):
Confused by how default port is used. I see that it's consistent in both Windows and Linux, but don't understand how the Tentacle port will line up with the port used when registering. I expect them to match but it looks like they are not going to.

docker/linux/scripts/configure-tentacle.sh
tothegills: Will setting `--tentacle-comms-port` alter the Tentacle configuration?

tothegills: Is it a container thing? The Tentacle should always be installed on port 10933, but may be mapped to another external port that the Octopus Server should use for comms?

N-lson: > The Tentacle should always be installed on port 10933, but may be mapped to another external port that the Octopus Server should use for comms?

I'm _pretty sure_ this is the case. Eg. in this image:
https://user-images.githubusercontent.com/11970672/149034313-3fd39afe-3616-4897-b65c-4e91a6632319.png

I set up the container with `-p 10934:10933` - meaning that port 10934 on the host maps to port 10933 on the Docker instance. I added `--env ListeningPort="10934"` which tells Server to use 10934 for comms.

N-lson: > Will setting --tentacle-comms-port alter the Tentacle configuration?

I don't think so. It is a bit confusing that it gets configured like this, maybe a better way would be to set `$ListeningPort` to `$defaultListeningPort` at the start of the script if it doesn't have a value and then use `$ListeningPort` for everything.

tothegills: Using just `$ListeningPort` as the linux image was, we suspect the docker ports must line up for registration to work:

```
-p 10933:10933
$ListeningPort = 10933
configure --port $ListeningPort
register-with --tentacle-comms-port $ListeningPort
```

If we start using the default port, the user will be able to remap the docker ports (seems like a good idea):
``
-p 10934:10933
$ListeningPort = 10934
configure --port $DefaultPort
register-with --tentacle-comms-port $ListeningPort
```

Suspected problem with changing it is that it will break anyone currently using a `$ListeningPort` that is not 10933. They would need to have their docker ports matching to make the registration work. eg: `-p 10934:10934`

N-lson: We're going to revert this change for now. Slack thread for discussion: https://octopusdeploy.slack.com/archives/C01J6U3MHJ4/p1641958569075100

tothegills (approved):

PR #351, opened 2022-01-10
Revert Harden executable to script, sign script during build

N-lson:
# Background

The Harden script was converted to an executable in this PR: https://github.com/OctopusDeploy/OctopusTentacle/pull/338. This was done since some customers require .ps1 scripts to be signed in order to run. However, this executable fails to run as part of Chocolatey verification, causing Tentacle to fail midway. New versions of Tentacle on Chocolatey are not being tagged as the latest official version as they are failing this verification step.

An alternative to the executable would be to sign the script which was also proposed in this PR: https://github.com/OctopusDeploy/OctopusTentacle/pull/321. Originally the executable was done since we weren't sure whether signing the script would introduce any security-related concerns. However, since the rest of the files (eg. Tentacle.exe, dlls) are already being signed using the same certificate, signing the script shouldn't have any side effects.

Fixes https://github.com/OctopusDeploy/ResearchAndDevelopment/issues/147

## Before

Using an instance of [chocolatey-test-environment](https://github.com/chocolatey-community/chocolatey-test-environment) with the latest Tentacle changes `6.1.1230`:
image

Running locally with an unsigned script, `-ExecutionPolicy Bypass` removed from the line running the Harden script and `Return="ignore"` removed:
image

## After

Using an instance of [chocolatey-test-environment](https://github.com/chocolatey-community/chocolatey-test-environment) with execution policy set to `AllSigned`:
image

Running locally with execution policy set to `AllSigned`:
image

# Testing

- Create an instance of [chocolatey-test-environment](https://github.com/chocolatey-community/chocolatey-test-environment)
- Build this branch of Tentacle using the `PackWindows` nuke build option
- Locate the Chocolatey package `OctopusTentacle\_artifacts\chocolatey`. I edited the `chocolateyInstall.ps1` to point to `C:\Packages\Octopus.Tentacle.6.1.1232-nelson-sign-script.msi` instead of the Url. Add this to the packages folder in the test env
- Add the .msi files to the packages folder in the test env
- Follow the test env instructions to try install the package

# Review

# Checks


installer/Octopus.Tentacle.Installer/Product.wxs
```diff
@@ -140,16 +140,11 @@
...
     -->
-    <CustomAction Id="HardenEXE"
-                  Directory="INSTALLLOCATION"
-                  ExeCommand="&quot;[INSTALLLOCATION]Octopus.Tentacle.Hardener.exe&quot;"
-                  Execute="deferred"
-                  Return="ignore"
-                  Impersonate="no"
-                 />
-
+    <CustomAction Id="SetExecProp" Property="WixQuietExecCmdLine" Value='"[POWERSHELLEXE]" -NoProfile -NonInteractive -InputFormat None -File "[INSTALLLOCATION]Harden-InstallationDirectory.ps1"'></CustomAction>
```
N-lson, line 143: The original command was:
```
[POWERSHELLEXE]" -NoProfile -NonInteractive -InputFormat None -ExecutionPolicy Bypass -File "[INSTALLLOCATION]Harden-InstallationDirectory.ps1"
```
But I don't think bypassing the execution policy is what we want to do so I've removed it.

installer/Octopus.Tentacle.Installer/Product.wxs
```diff
@@ -140,16 +140,11 @@
...
+    <CustomAction Id="HardenInstallationDirectory" BinaryKey="WixCA" DllEntry="WixQuietExec" Execute="immediate" Return="ignore"/>
```
N-lson, line 144: I don't know if we want this `Return="ignore"` here - if it's included, then even if the harden script fails, the installation will still say it succeeded.

tothegills: In consultation with TSO we decided that we would make best effort to run the script, but it was not essential for a successful Tentacle install.

installer/Octopus.Tentacle.Installer/Product.wxs
N-lson: I did some more testing and it looks like `-ExecutionPolicy Bypass` is needed. Maybe it's something to do with Wix - even though the policy is set to RemoteSigned on my laptop, when I write the output to a log it says running scripts is disabled.

tothegills (approved):

Issue #360, opened 2022-04-13
Tentacle is susceptible to a bug in halibut where exceptions in the DataStream can cause tentacle to use 100% of the CPU 


LukeButters:
Tentacle is susceptible to https://github.com/OctopusDeploy/Halibut/issues/141

We have observed this behaviour in experiments.

This issue may have been seen for real in https://github.com/OctopusDeploy/Halibut/issues/117

octoreleasebot:
Release Note: Tentacle will no longer get stuck in an infinite loop when the stream is closed, which can occur when an Exception is thrown in a DataStream.

PR #361, opened 2022-04-19
Fix the stream closed halibut bug.

LukeButters:
The is, if octopus server closes the stream, tentacle will consume 100% CPU trying to read the rest of the stream.

The fix is actually in: https://github.com/OctopusDeploy/Halibut/pull/142
Fixes: https://github.com/OctopusDeploy/OctopusTentacle/issues/360

# Review

Does this make sense?

# Checks

### DIFF
--- a/source/Octopus.Tentacle.Tests/Octopus.Tentacle.Tests.csproj
+++ b/source/Octopus.Tentacle.Tests/Octopus.Tentacle.Tests.csproj
@@ -25,7 +25,6 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Octopus.Tentacle\Octopus.Tentacle.csproj" />
-    <PackageReference Include="Halibut" Version="4.4.9" />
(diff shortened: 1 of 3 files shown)


source/Octopus.Tentacle.Tests/Octopus.Tentacle.Tests.csproj
```diff
@@ -25,7 +25,6 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Octopus.Tentacle\Octopus.Tentacle.csproj" />
-    <PackageReference Include="Halibut" Version="4.4.9" />
```
LukeButters, line 28: I think this reference doesn't need to exist since it would be included via `Octopus.Tentacle.csproj`

LukeButters:
Is the build really working? that feels way to fast

tothegills (approved):

LukeButters:
Seems to work with octopus [here](https://github.com/OctopusDeploy/OctopusDeploy/pull/11931)

PR #369, opened 2022-05-28
Use apt to install the latest Docker daemon and upgrade docker-ce

### LINKED ISSUE #381: Current Tentacle Docker image does not work on Linode, GitLab, some other hosts
### Team

- [X] I've assigned a team label to this issue

### What happened?

When running on a container host which does hot have the `systemd` cgroup moint point, Tentacle cannot run/use Docker-in-Docker.

### Reproduction

1. Procure a VM running without the systemd cgroup mount point. A Linode VM will work but there will be others I'm not aware of.
2. Run a Tentacle container on that VM.
3. Attempt to run a script on that Tentacle using Docker-in-Docker.

If it helps, I can give a quick demo of this failure mode using existing infrastructure rather than needing someone to provision anything. Please reach out directly if that would be useful.

### Error and Stacktrace

```shell
docker: Error response from daemon: cgroups: cgroup mountpoint does not exist: unknown.
```

### More Information

It looks like this is caused by shipping with an older version of Docker on the Tentacle image. A simple upgrade to the latest `docker-ce` package appears to fix the issue.

There is a [relevant issue](https://github.com/docker/for-linux/issues/1334) on the Docker repository itself that suggests that this has been fixed in a later version, which appears to be the case based on my brief experimentation.

### Workaround

# Workaround using custom Tentacle Dockerfile
```
FROM octopusdeploy/tentacle
RUN apt-get update && apt-get upgrade -y docker-ce
```

# Workaround for a running Tentacle container:
```
docker exec -it <tentacle_container> bash
apt-get update && apt-get upgrade -y docker-ce
killall dockerd
/scripts/dockerd-entrypoint.sh &
```

uglybugger:
# Background

Fixes #381 .

# Testing

1. Push the changes; wait until a prerelease build is ready.
2. `docker run -it --rm ...` that generated image.
3. Confirm that Tentacle can run Docker-in-Docker commands on relevant virtual machines.

To do the last:
```bash
docker run -it --rm --privileged docker.packages.octopushq.com/octopusdeploy/tentacle:<version_from_build> bash
```

In the resulting Docker container:
```
# Launch the Docker daemon. This is normally done by the Tentacle's startup script itself.
/scripts/dockerd-entrypoint.sh &

# Confirm that we can talk to the nested Docker daemon
docker ps

# Confirm that the nested Docker daemon can run a container.
docker run -it --rm alpine pwd
```

# Review

* We previously used a `curl`ed script to install Docker; however, the installation procedure using apt is now much more usable.
* Ultimately, all this really does is version-bumps Docker using a nicer installation mechanism.
* It would be good to better understand whether _everything_ the previous installation script did can be removed. It appears to but I don't have a good feel for what the downstream dependencies might be, so I've kept them to be safe.

# Checks

### DIFF
--- a/docker/linux/install-scripts/install-docker.sh
+++ b/docker/linux/install-scripts/install-docker.sh
@@ -3,19 +3,42 @@ set -eux
 
 # This script is adapted from https://github.com/docker-library/docker/blob/master/19.03/dind/Dockerfile
 
+
+# Add the apt sources for Docker (they're not part of the stock Debian distro).
 apt-get update
 
+apt-get install -y --no-install-recommends \
+    apt-transport-https \
+    ca-certificates \
+    curl \
+    gnupg \
+    lsb-release
+
+curl -fsSL https://download.docker.com/linux/debian/gpg | gpg --dearmor -o /usr/share/keyrings/docker-archive-keyring.gpg
+
+echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/docker-archive-keyring.gpg] https://download.docker.com/linux/debian \
+  $(lsb_release -cs) stable" | tee /etc/apt/sources.list.d/docker.list > /dev/null
+
+
+# Install Docker and its runtime dependencies.
 # https://github.com/docker/docker/blob/master/project/PACKAGERS.md#runtime-dependencies
+
+apt-get update
 apt-get install -y \
     btrfs-progs \
+    containerd.io \
+    docker-ce \
+    docker-ce-cli \
+    dos2unix \
     e2fsprogs \
     iptables \
+    jq \
     openssl \
+    pigz \
+    sudo \
     uidmap \
     xfsprogs \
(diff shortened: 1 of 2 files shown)


docker/linux/install-scripts/install-docker.sh
```diff
@@ -33,9 +56,6 @@ dos2unix /usr/local/bin/dind
 chmod +x /usr/local/bin/dockerd-entrypoint.sh
 dos2unix /usr/local/bin/dockerd-entrypoint.sh
 
-export VERSION=19
-curl -sSL https://get.docker.com/ | sh
-
```
uglybugger, line 38: This runs the Docker installation script. We don't need that any longer because we're installing it via `apt-get install` above.

ankithkonda (approved):
Thanks for putting this together @uglybugger 🙇

ankithkonda:
Hey @uglybugger, are you happy to merge this? :)

PR #387, opened 2022-07-25
Allow only one msi when upgrading to slim down Tentacle upgrade

N-lson:
# Background

As part of the Tentacle upgrade work, this reduces the need to upload both msi files when we already know the architecture of a machine.

# Results

Relates to https://github.com/OctopusDeploy/Issues/issues/7082

Fixes https://github.com/OctopusDeploy/OctopusTentacle/issues/388 (this is mainly so we can get a "meaningful change" for this Tentacle release so that the release is publicly available)

# How to review this PR

Quality :heavy_check_mark:

### DIFF
--- a/source/Octopus.Tentacle.Upgrader/Program.cs
+++ b/source/Octopus.Tentacle.Upgrader/Program.cs
@@ -89,10 +89,21 @@ namespace Octopus.Tentacle.Upgrader
 
         static string SelectAppropriateMsi(IList<string> args)
         {
-            var msi = args.Count == 2 ? args[1] : Environment.Is64BitOperatingSystem ? args[2] : args[1];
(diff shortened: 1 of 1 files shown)


source/Octopus.Tentacle.Upgrader/Program.cs
```diff
@@ -89,7 +89,7 @@ static void LogStartupParameters(IList<string> args)
 
         static string SelectAppropriateMsi(IList<string> args)
         {
-            var msi = Environment.Is64BitOperatingSystem ? args[2] : args[1];
```
tothegills, line 92: This is a bit confusing to me. When there are only 2 arguments, do things a new way. Otherwise do things the old way. I wonder how this could be expressed in a way that makes it clear to the reader why there are two slightly different code paths based on the argument count?

tothegills (approved):

PR #390, opened 2022-07-28
Remove all windows docker related files

N-lson:
Relates to https://github.com/OctopusDeploy/ResearchAndDevelopment/issues/534

Since we're removing these Pester tests (they haven't been running for a while anyway) it may be nice to have E2E tests for the docker image similar to `Octopus.Server.Containers.Testing` in Server.


veochen-octopus (approved):

BenH-Puregym:
@N-lson , why have you removed all of the windows docker files? Isn't it supposed to be the code which sets up the windows tentacle in dockerhub? https://hub.docker.com/r/octopusdeploy/tentacle

N-lson:
Hey @BenH-Puregym, the Windows Tentacle Docker image is no longer being published or maintained - see [our docs](https://octopus.com/docs/installation/octopus-tentacle-container). This image will still be available for Tentacle 6.1 and lower but it's recommended to switch to the Linux image. This is following the same approach we took for the [Windows Octopus Server Docker image](https://octopus.com/docs/installation/octopus-server-windows-container) which was deprecated a while back.

BenH-Puregym:
What about apps that must be running on a windows build agent as they require dotnet framework?

N-lson:
Thanks for raising that point @BenH-Puregym! It turns out we did originally decide to maintain a version of the Windows Tentacle Docker image last year. This was something we missed when removing support for any images using an EOL Windows Server version. We'll re-add the required files and will maintain a single version of the Windows Tentacle Docker image. This will use **Windows Server 2019 (1809)** and will eventually move over to Windows Server 2022.

Apologies for any concerns and we appreciate the feedback!

PR #395, opened 2022-08-10
Move Octopus.Shared into Octopus.Tentacle

tothegills:
# Background
This PR moves code from the `Octopus.Shared` project into the `Octopus.Tentacle` project. Also consolidates `Octopus.Shared.Tests` into `Octopus.Tentacle.Tests`. There are some tests that must run on a VM, these have been moved into a new project `Octopus.TentacleTests.Integration`.

The `Octopus.Shared` project remains, containing the script contracts for backwards compatibility.

This PR should not change any functionality.

[Server build](https://github.com/OctopusDeploy/OctopusDeploy/pull/13556) using this Tentacle.


tothegills:
Fixes https://github.com/OctopusDeploy/ResearchAndDevelopment/issues/486

source/Octopus.Tentacle.Tests.Integration/Class1.cs
```diff
@@ -0,0 +1,8 @@
+﻿using System;
+
+namespace Octopus.Tentacle.Tests.Integration
+{
+    public class Class1
```
N-lson, line 5: Do we need this?

source/Octopus.Tentacle/Configuration/Proxy/ProxyPasswordMaskValuesProvider.cs
```diff
@@ -9,15 +9,15 @@ namespace Octopus.Tentacle.Configuration.Proxy
...
     }
     
     public class ProxyPasswordMaskValuesProvider : IProxyPasswordMaskValuesProvider
     {
         static readonly Regex UrlEncodedCharactersRegex = new Regex(@"%[A-F0-9]{2}", RegexOptions.Compiled);
-        public IEnumerable<string> GetProxyPasswordMaskValues(string proxyPassword)
+        public IEnumerable<string> GetProxyPasswordMaskValues(string? proxyPassword)
         {
-            if (string.IsNullOrEmpty(proxyPassword)) 
+            if (proxyPassword == null || string.IsNullOrEmpty(proxyPassword))
```
N-lson, line 20: Looks like the existing condition should already check for nullability?

source/Octopus.Tentacle/Commands/CreateInstanceCommand.cs
```diff
@@ -1,15 +1,15 @@
...
 
 namespace Octopus.Tentacle.Commands
 {
     public class CreateInstanceCommand : AbstractCommand
     {
         readonly IApplicationInstanceManager instanceManager;
-        string instanceName;
-        string config;
-        string home;
+        string instanceName = null!;
```
N-lson, line 10: Do we want to revisit these `null!` at some point to clean them up? Think they're fine to keep in as part of this PR.

source/Octopus.Tentacle/Octopus.Tentacle.csproj
```diff
@@ -13,6 +13,7 @@
 		<TargetFrameworks>net452;netcoreapp3.1</TargetFrameworks>
 		<TreatWarningsAsErrors>true</TreatWarningsAsErrors>
 		<LangVersion>9</LangVersion>
+		<Nullable>enable</Nullable>
```
N-lson, line 16: 👍

N-lson (approved):
Looks like a good move 👍 things that aren't shared make more sense belonging in `Octopus.Tentacle` itself.

source/Octopus.Tentacle/Commands/CreateInstanceCommand.cs
tothegills: Yes, I might change these back for the test project so they are easier to spot if we flip nullable on.

source/Octopus.Tentacle/Configuration/Proxy/ProxyPasswordMaskValuesProvider.cs
tothegills: It doesn't and results in a compiler error.

N-lson: damn

source/Octopus.Tentacle/Commands/CreateInstanceCommand.cs
tothegills: There is some "prior art" in the tests that uses nullable and `null!` initializers. Will leave them as-is for now, tech-debt to be paid off in the future.

PR #405, opened 2022-08-16
Prevent tests from rebuilding binaries

tothegills:
Currently we build the Tentacle solution and publish binaries as artifacts. Our tests depend on the artifacts, but then rebuild each time a test runs. We are not testing the binaries that we ship. This PR removes the build phase from our tests.

### DIFF
--- a/build/Build.Tests.cs
+++ b/build/Build.Tests.cs
@@ -18,17 +18,14 @@ partial class Build
 {
     [PublicAPI]
     Target TestWindows => _ => _
-        .DependsOn(BuildWindows)
         .Executes(() => RunTests(TestFramework, TestRuntime));
 
     [PublicAPI]
     Target TestLinux => _ => _
-        .DependsOn(BuildLinux)
         .Executes(() => RunTests(TestFramework, TestRuntime));
 
     [PublicAPI]
     Target TestOsx => _ => _
-        .DependsOn(BuildOsx)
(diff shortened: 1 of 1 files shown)


ankithkonda:
Could we put a conditional on these so that they don't depend on the build if it's on teamcity? But running locally they should build?

tothegills:
I don't see a way to add conditional dependencies.

ankithkonda:
looks like this should work - https://octopusdeploy.slack.com/archives/C024XB4RMUP/p1662047706698699?thread_ts=1661732747.777679&cid=C024XB4RMUP
`OnlyWhenStatic(() => Host is TeamCity)`

tothegills:
From what I understand `OnlyWhenStatic` is a pre-condition for the current target. So to use that, we could write:

```
    Target BuildWindowsLocally => _ => _
        .DependsOn(BuildWindows)
        .OnlyWhenStatic(() => !IsLocalBuild);
```

And the same for OSX and Linux. I'm not sure the tradeoff of forks in the build script is worth the convenience of typing less on the cli when someone runs a build script locally?

ankithkonda:
How about `--skip`? https://nuke.build/docs/getting-started/execution/#skipping-targets

Nuke build script won't need to change. But in TeamCity, test config we can say `./build.sh --Target Test --skip Compile`

Screen Shot 2022-09-07 at 10 01 20 am

N-lson (approved):

PR #408, opened 2022-09-05
Pack Tentacle contracts

tothegills:
# Background

This PR creates a NuGet package containing `Octopus.Tentacle.Contracts.dll`. This package will be consumed by Octopus Server and Dynamic Workers so each of those solutions do not maintain their own copies of the Tentacle contracts.

Relates to https://github.com/OctopusDeploy/ResearchAndDevelopment/issues/487


build/Build.Pack.cs
```diff
@@ -257,11 +259,26 @@ void BuildMsiInstallerForPlatform(MSBuildTargetPlatform platform, AbsolutePath w
...
+                .SetProcessArgumentConfigurator(_ => _ .Add("/p:NoWarn=NU5104")));
+        });
+
     [PublicAPI]
     Target PackWindows => _ => _
         .Description("Packs all the Windows targets.")
         .DependsOn(PackWindowsZips)
-        .DependsOn(PackChocolateyPackage);
+        .DependsOn(PackChocolateyPackage)
+        .DependsOn(PackContracts);
```
N-lson, line 281: Only on Windows?

build/Build.Pack.cs
```diff
@@ -257,11 +259,26 @@ void BuildMsiInstallerForPlatform(MSBuildTargetPlatform platform, AbsolutePath w
...
+        .Description("Packs the NuGet package for Tentacle contracts.")
+        .Executes(() =>
+        {
+            DotNetPack(p => p
+                .SetProject(RootDirectory / "source/Octopus.Tentacle.Contracts/Octopus.Tentacle.Contracts.csproj")
+                .SetVersion(OctoVersionInfo.FullSemVer)
+                .SetOutputDirectory(ArtifactsDirectory)
+                .DisableIncludeSymbols()
+                .SetVerbosity(DotNetVerbosity.Normal)
+                .SetProcessArgumentConfigurator(_ => _ .Add("/p:NoWarn=NU5104")));
```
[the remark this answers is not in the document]
tothegills: No we don't need it. 👍 The offending references are gone.

build/Build.Pack.cs
tothegills: It's a cross-platform package, so only needs to be packed once. I'm piggybacking on the Windows build, but could be part of any of the builds or have it's own. What do you think would be the cleanest/clearest approach?

N-lson: I think having it as a dependency on `Pack` would be the clearest. On TeamCity we never call `Pack` though. We directly call `PackCrossPlatformBundle`. So we could either:

- Call `Pack` on TeamCity instead of `PackCrossPlatformBundle` and update the name of the build step
- Add another build step for calling `PackContracts`

Not sure which would be better, both sound ok to me.

tothegills: I like the idea of decoupling it from the Window pack. I think a build step that runs `PackContracts` is a nice solution.

N-lson (approved):

PR #422, opened 2022-10-04
net48 and net6

veochen-octopus:
# Background

- `net452` is end of support and is preventing us from getting security patches both in the framework itself and 3rd party libraries.
- We'll need to support Ubuntu 22.04 pretty soon so `net6` is also required.

# Results

Fixes https://app.shortcut.com/octopusdeploy/story/12030/target-net-framework-4-8-in-tentacle

## Before

```xml
<TargetFrameworks>net452;netcoreapp3.1</TargetFrameworks>
```

## After

```xml
<TargetFrameworks>net48;net6.0</TargetFrameworks>
```

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Octopus.Tentacle/Security/Certificates/CertificateEncoder.cs
```diff
@@ -186,7 +186,7 @@ static bool HasPrivateKey(X509Certificate2 certificate2)
         {
             try
             {
-                return certificate2.HasPrivateKey && certificate2.PrivateKey != null;
+                return certificate2.GetRSAPrivateKey() != null;
```
veochen-octopus, line 189: This is the only code change I had to make. The `EphemeralKeySet` flag is inside a `#if NET472_OR_GREATER || NETCOREAPP || NETSTANDARD` condition, so it was always ignored on `net452` and `PersistKeySet` would be used. From `net472` we have `EphemeralKeySet` but it was erroring due to private key not being found. As per the MS docs on `EphemeralKeySet` support:

> Keys loaded in this manner are almost always loaded via Windows CNG. Therefore, callers must access the private key by calling extension methods, such as [cert.GetRSAPrivateKey()](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.rsacertificateextensions.getrsaprivatekey). The [X509Certificate2.PrivateKey](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509certificate2.privatekey#system-security-cryptography-x509certificates-x509certificate2-privatekey) property does not function.

I'm assuming `GetECDsaPrivateKey` doesn't come into play here. There's only reference to RSA.

source/Octopus.Tentacle/Security/Certificates/CertificateEncoder.cs
```diff
@@ -103,7 +103,7 @@ static X509Certificate2 FromBase64String(string? thumbprint, string certificateS
             }
         }
         
-#if NET472_OR_GREATER || NETCOREAPP || NETSTANDARD
+#if NETCOREAPP || NETSTANDARD
```
veochen-octopus, line 106: This used to be bypassed as we were targeting net452, but breaks net48. With `EphemeralKeySet` the cert loses the private key once added to the key store. As per the doc
> Since the keys are not persisted to disk, certificates loaded with this flag are not good candidates to add to an X509Store.

veochen-octopus:
Cloning the tentacle project in teamcity for the net48 build is a bit wonky for some reason. The branch filter doesn't work all the time. Thinking once the PR is ready, we update the main tentacle project in teamcity first to target net48, then get a green build and merge.

source/Octopus.Tentacle/Security/Certificates/CertificateEncoder.cs
veochen-octopus: This essentially allows us to keep the same behaviour as `net452`

veochen-octopus:
I'll post net48 green builds here til we update the main build chain. Here's the current one - https://build.octopushq.com/buildConfiguration/TeamFireAndMotion_OctopusTentacleNet48_SensibleDefaultsChainFullChain/5905429?hideProblemsFromDependencies=false&hideTestsFromDependencies=false

source/Octopus.Tentacle.Tests.Integration/Octopus.Tentacle.Tests.Integration.csproj
```diff
@@ -14,25 +14,25 @@
...
-        <PackageReference Include="NSubstitute" Version="4.2.2" />
-        <PackageReference Include="FluentAssertions" Version="5.10.3" />
-        <PackageReference Include="Assent" Version="1.6.0" />
-        <PackageReference Include="NUnit" Version="3.12.0" />
-        <PackageReference Include="NUnit3TestAdapter" Version="3.13.0" />
+        <PackageReference Include="NSubstitute" Version="4.4.0" />
+        <PackageReference Include="FluentAssertions" Version="6.7.0" />
+        <PackageReference Include="Assent" Version="1.8.2" />
+        <PackageReference Include="NUnit" Version="3.13.3" />
+        <PackageReference Include="NUnit3TestAdapter" Version="4.2.1" />
```
veochen-octopus, line 29: Taking the opportunity to update _some_ test packages. No non-test dependencies were changed.

source/Octopus.Tentacle/Security/Certificates/CertificateEncoder.cs
veochen-octopus: Also TODO: Add a comment that this _may_ be reverted to `.PrivateKey` once .NET framework is dropped entirely. Will use this change to trigger the final build before merge.

tothegills:
What kind of customer comms are we doing? Does that need to happen before this is merged?

veochen-octopus:
> What kind of customer comms are we doing? Does that need to happen before this is merged?

@tothegills I think @octokhor is handling the comms side of things. Given the change of plan, I'll piggyback the net6 changes onto this PR as well so we end up with only 1 major version bump.

source/Octopus.Tentacle.Tests/Configuration/Crypto/LinuxMachineKeyEncryptorFixture.cs
```diff
@@ -70,7 +70,7 @@ class InMemoryCryptoKeyNixSource : ICryptoKeyNixSource
             readonly byte[] iv;
             public InMemoryCryptoKeyNixSource()
             {
-                var d = new RijndaelManaged();
+                var d = Aes.Create();
```
veochen-octopus, line 73: `RijndaelManaged` obsolete. All references changed to `Aes.Create`

source/Octopus.Tentacle/Certificates/CertificateGenerator.cs
```diff
@@ -97,7 +97,7 @@ X509Certificate2 Generate(string fullName, bool exportable)
             using (var ms = new MemoryStream())
             {
                 store.Save(ms, exportpw.ToCharArray(), random);
-                var platformSpecificX509KeyStorageFlags = PlatformDetection.IsRunningOnMac ? X509KeyStorageFlags.DefaultKeySet : X509KeyStorageFlags.EphemeralKeySet;
+                var platformSpecificX509KeyStorageFlags = PlatformDetection.IsRunningOnNix ? X509KeyStorageFlags.EphemeralKeySet : X509KeyStorageFlags.DefaultKeySet;
```
veochen-octopus, line 100: Don't think `EphemeralKeySet` works with Windows at all. See other change.

source/Octopus.Tentacle/Security/AesEncryption.cs
```diff
@@ -20,69 +20,62 @@ public AesEncryption(string password)
...
         {
-            var provider = new AesCryptoServiceProvider
-            {
-                Mode = CipherMode.CBC,
-                Padding = PaddingMode.PKCS7,
-                KeySize = 128,
-                BlockSize = 128,
-                Key = key
-            };
+            var provider = Aes.Create();
```
veochen-octopus, line 60: Another `obsolete`. Updated to current standard.

veochen-octopus:
Not sure about the best way to update the teamcity build. There are quite a few changes to be made and it's hard to keep track of them. Thinking probably the easiest is to replace the old tentacle project with the new one and make sure everything has the correct names and ids, get a green build and merge.

docker/windows/Dockerfile
```diff
@@ -20,13 +21,10 @@ LABEL \
 EXPOSE 10933
 
 # Either from a build artifact dependency or local build output.
-COPY _artifacts/msi/Octopus.Tentacle.${BUILD_NUMBER}-x64.msi /tmp/
-
-WORKDIR /tmp
-RUN msiexec /i Octopus.Tentacle.%BUILD_NUMBER%-x64.msi /qn /l*v Tentacle-Installation.log
+COPY _artifacts/tentacle "${INSTALLATION_FOLDER}"
```
veochen-octopus, line 24: This is the only way I've found that works with `COPY` and Windows paths that contain spaces, i.e. with an `ARG`. Open to suggestions.

veochen-octopus:
Latest green build - https://build.octopushq.com/buildConfiguration/TeamFireAndMotion_OctopusTentacleNet48_SensibleDefaultsChainFullChain/5962281?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&expandPull+Request+Details=true

tothegills:
It'd be worth creating an Octopus Server branch, updating the Tentacle version and running the Octopus Server test suite.

installer/Octopus.Tentacle.Installer/Product.wxs
```diff
@@ -1,80 +1,59 @@
...
     -->
-    <PropertyRef Id="NETFRAMEWORK45"/>
-    <PropertyRef Id="NETFRAMEWORK40FULLINSTALLROOTDIR"/>
-    <Condition Message="This application requires Microsoft .NET Framework 4.5 Runtime in order to run. Please install the .NET Framework and then run this installer again.">
-      <![CDATA[Installed OR NETFRAMEWORK45]]>
-    </Condition>
-
+    <PropertyRef Id="NETFRAMEWORK45" />
+    <PropertyRef Id="NETFRAMEWORK40FULLINSTALLROOTDIR" />
+    <Condition Message="This application requires Microsoft .NET Framework 4.5 Runtime in order to run. Please install the .NET Framework and then run this installer again."><![CDATA[Installed OR NETFRAMEWORK45]]></Condition>
```
tothegills, line 52: I don't think this message about .NET framework 4.5 is still true

source/Octopus.Tentacle.Tests/Configuration/InMemoryKeyValueStoreFixture.cs
```diff
@@ -91,8 +91,9 @@ public void ComplexTypeGetsHandledCorrectly()
             var store = new InMemoryKeyValueStore(mapper);
 
             var settings = store.TryGet<TestConfig[]>("Test");
-            settings.value.Single().SettingA.Should().Be("some value", "strings should get parsed");
-            settings.value.Single().SomethingElse.Should().Be(12, "ints should get parsed");
+            settings.value.Should().NotBeNull();
```
tothegills, line 94: The null check is an important assertion

source/Octopus.Tentacle/Communications/OctopusServerChecker.cs
```diff
@@ -38,34 +38,40 @@ public string CheckServerCommunicationsIsOpen(Uri serverAddress, IWebProxy? prox
 
             Retry("Checking that server communications are open", () =>
             {
-#pragma warning disable DE0003
+                // TODO see if can deal with SYSLIB0014
```
tothegills, line 41: Is this TODO still to do? What are the plans for doing it?

source/Octopus.Tentacle/Security/Certificates/CertificateEncoder.cs
tothegills: Does this TODO need action?

source/Octopus.Tentacle/Versioning/AssemblyExtensions.cs
```diff
@@ -9,13 +9,13 @@ public static class AssemblyExtensions
         public static string GetFileVersion(this Assembly assembly)
         {
             var fileVersionInfo = FileVersionInfo.GetVersionInfo(assembly.FullLocalPath());
-            return fileVersionInfo.FileVersion;
+            return fileVersionInfo.FileVersion ?? "";
```
tothegills, line 12: Is "" a good backup value?

tothegills (commented):
Great work, way to move this forward.
I think there are some installer pre-reqs that need updating.
I think Fixes should be a public issue in this repository.
It would be great to have a green Octopus Server build with these changes integrated.

installer/Octopus.Tentacle.Installer/Product.wxs
veochen-octopus: 🤔 I think this was auto generated. Let me see where it got this information from.

source/Octopus.Tentacle/Versioning/AssemblyExtensions.cs
veochen-octopus: Good question. We could return "0.0.0". It would be more friendly to the semantic version parser. But it would also hide the fact that there is no version. Thinking about the responsibility of this method, _should_ it know what a good default is? Maybe we should simply return a `string?` and let the consumer decide what they want to use as the default. For example

```csharp
// SemanticVersionInfo.cs
SemanticVersion = SemanticVersion.Parse(assembly.GetInformationalVersion() ?? "0.0.0");

// OctopusTentacle.cs
public static readonly string InformationalVersion = AssemblyExtensions.GetInformationalVersion(Assembly) ?? "Unknown";
```

installer/Octopus.Tentacle.Installer/Product.wxs
veochen-octopus: Can I please get some help on this? Don't want to dump too much time into learning Wix (can do that later). But from what I can see from [the manual](https://wixtoolset.org/documentation/manual/v3/customactions/wixnetfxextension.html), there's no good replacement for `NETFRAMEWORK45`  for `net48`. This is wix v3. Not sure about v4.

Also found a [snippet in their source code](https://github.com/wixtoolset/wix3/blob/c02e48ec301a60eba88a3b519d47e88eeaa4c978/src/ext/NetFxExtension/wixlib/NetFx48.wxs) that has both similar and very different settings. A little confused...

source/Octopus.Tentacle/Communications/OctopusServerChecker.cs
veochen-octopus: Removing this TODO. Fix would be to properly inject an `HttpClient` (or the factory) and go async from there. Treating it as out of scope for now, as it might not be a local change. But important to address later on. Ideally make tentacle more async overall.

source/Octopus.Tentacle/Security/Certificates/CertificateEncoder.cs
veochen-octopus: Yeah that was confusing. Looking at it again, no action needed.

veochen-octopus:
Main changes

- Updated `AssemblyExtensions.GetInformationalVersion` to return a `string?` and force the consumer to provide a default, i.e. `assembly.GetInformationalVersion() ?? "0.0.0"`
- Updated `Product.wxs` to require net48. (Couldn't merge your PR Shane so I've copied your change)
- Enabled Ubuntu 22.04 test and added `libssl3` to the deb dependencies so that the test is passing
- Renamed new Teamcity build to `Octopus Tentacle 2022` and updated build ids.
  - Need to make sure the build number is valid before switching to the new build

build/Build.Tests.cs
```diff
@@ -85,8 +85,7 @@ void RunLinuxPackageTestsFor(TestConfigurationOnLinuxDistribution testConfigurat
                 new TestConfigurationOnLinuxDistribution(NetCore, "linux-x64", "debian:oldstable-slim", "deb"),
                 new TestConfigurationOnLinuxDistribution(NetCore, "linux-x64", "debian:stable-slim", "deb"),
                 new TestConfigurationOnLinuxDistribution(NetCore, "linux-x64", "linuxmintd/mint19.3-amd64", "deb"),
-                // new TestConfigurationOnLinuxDistribution(NetCore, "linux-x64", "ubuntu:latest", "deb"), // 22.04 doesn't support netcore, https://github.com/dotnet/core/issues/7038
-                // new TestConfigurationOnLinuxDistribution(NetCore, "linux-x64", "ubuntu:rolling", "deb"), // 22.04 doesn't support netcore, https://github.com/dotnet/core/issues/7038
+                new TestConfigurationOnLinuxDistribution(NetCore, "linux-x64", "ubuntu:jammy", "deb"), // 22.04
```
tothegills, line 88: It's helpful to test against `latest` and `rolling` because they alert us to problems that are not caught by testing the explicit versions.

veochen-octopus: I've added these back and turned them on. I assume when this starts to fail, we'll disable them again until we're ready for the next upgrade?

tothegills (commented):
Looks good to me. Intending to merge this PR on November 21st.

veochen-octopus:
Ah yes, the "something else" I was trying to remember in the discussion points from earlier is this PR https://github.com/OctopusDeploy/OctopusTentacle/pull/405. I did the same thing without knowing about it. Seeing there's an ongoing discussion in the other PR this could change, which we can do afterwards.

N-lson (approved):

PR #414, opened 2022-09-14
Added ability to specify a different server address for the tentacle comms

pawelpabich:
# Background

[We are adding polling tentacles over standard ports (443) to Octopus Cloud](https://docs.google.com/document/d/1iOigyxajs1rLbGGUZSwl9AR8HMJFkww5MSvZoiRU6eM/edit#heading=h.g165efrlab8g). For that, we need the polling tentacle to use a different server address for the web traffic and the tentacle traffic.

This PR is just a start. We should add the same option to [the Tentacle Manager](https://github.com/OctopusDeploy/OctopusTentacle/tree/main/source/Octopus.Manager.Tentacle) and [Tentacle docker images](https://github.com/OctopusDeploy/OctopusTentacle/blob/main/docker). Anything else that I've missed here?

# Results

## Before

The same address was used for both the web and tentacle traffic.

## After

```C#
.\tentacle.exe register-worker  .... --server "https://pawel-stable3.testoctopus.app"  --server-comms-port "443" --server-comms-address "https://polling.pawel-stable3.testoctopus.app"
```

image

PR #440, opened 2023-01-11
Added ability to specify a different server address for the tentacle comms

bradleyjford:
# Background

> [We are adding polling tentacles over standard ports (443) to Octopus Cloud](https://docs.google.com/document/d/1iOigyxajs1rLbGGUZSwl9AR8HMJFkww5MSvZoiRU6eM/edit#heading=h.g165efrlab8g). For that, we need the polling tentacle to use a different server address for the web traffic and the tentacle traffic.

This PR supersedes the PoC [#414](https://github.com/OctopusDeploy/OctopusTentacle/pull/414) with updates to the Docker files and the addition of unit tests. 

We have introduced a new `--server-comms-address` CLI argument to the `register-worker` and `register-machine` commands.

The `--server-comms-port` takes precedence over any port specified by the `--server-comms-address`. If only `--server-comms-address` is specified, the port implied by the URL is used (i.e. the default for `https://` is 443). This aims to reduce confusion but also maintain backwards compatibility.

# Results

See [How we use GitHub Issues](https://github.com/OctopusDeploy/Issues/blob/master/docs/CONTRIBUTING.internal.md) (including [this flowchart](https://whimsical.com/r-d-incoming-work-workflow-aug-21-NsDnGQXcwBLwU66a88Zhue) 

## Before

The same address was used for both the web and tentacle traffic.

## After

```
.\tentacle.exe register-worker  .... --server "https://pawel-stable3.testoctopus.app"  --server-comms-address "https://polling.pawel-stable3.testoctopus.app"
```

image

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


zentron (changes requested):
The individual changes look like they make sense.

I can see how providing the `server-comms-address` separate to the existing `server` argument allows you to use the `server` parameter for purely the registration invocation, leaving comms to go via another address. all good 👍 

Given though we use that `--server`  argument for other commands i'd like to make sure we are covering this behavioural addition in those places as well.
Should this argument also be added to the `PollCommand` as well?

sahanocto:
@zentron As requested, we have updated the `PollCommand` to accept the new `--server-comms-address` argument. We are happy for you to have another review. Thanks.

zentron:
Thanks for following up. Happy with the final merge

zentron (approved):
Approved

PR #441, opened 2023-01-12
Consolidate CoreUtilities

N-lson:
# Background

[CoreUtilities](https://github.com/OctopusDeploy/CoreUtilities) is being archived and should be consolidated into Tentacle.

[sc-33213]

# How to review this PR
Quality :heavy_check_mark:


tothegills (approved):
Think you can squash this one, the history will just add noise

PR #460, opened 2023-04-17
Add Tentacle ScriptServiceV2(Alpha)

nathanwoctopusdeploy:
# Background

Adds a new ScriptServiceV2 that is idempotent and truthful.

Full context for this change can be found in this [RFC](https://docs.google.com/document/d/1Mf5c5U_vZFTMCNjVLr6EvzWvnIRz_C5DJygXFyDAmIQ/edit#heading=h.dltuw5i0lklv) [internal]

- Add ScriptServiceV2
- Add new request/command and response contracts for ScriptServiceV2
- Added a capability to Tentacle `ScriptServiceV2Alpha`. Alpha so this can be tested on branch instances prior to going GA

## Full Server Chain with updated Tentacle
- 🍏   vNext https://github.com/OctopusDeploy/OctopusDeploy/pull/18465
- 🍏  2023.2 https://github.com/OctopusDeploy/OctopusDeploy/pull/18464
- 🍏 2023.1 https://github.com/OctopusDeploy/OctopusDeploy/pull/18467
- 🍏 2022.4 https://github.com/OctopusDeploy/OctopusDeploy/pull/18468
- 🍏 Dynamic Worker Image Ubuntu https://github.com/OctopusDeploy/DynamicWorkerVmImage-Ubuntu2204/pull/38
- 🍏  Dynamic Worker Image Windows https://github.com/OctopusDeploy/DynamicWorkerVmImages/pull/126

🍏  WIP - Draft PR to use the ScriptServiceV2 https://github.com/OctopusDeploy/OctopusDeploy/pull/18366 - E2E Tests passing and manually tested on a pre-prod cloud instance

## ScriptServiceV2

- Idempotent (StartScript can be called many times with the same script ticket and will only start the script once)
- Truthful. failure exit code returned if an unknown script ticket is provided or the final result of running a script is unknown
- GetStatus guarantees all log messages are returned 
- CompleteScript is a pure cleanup operation

ScriptServiceV2 also supports waiting for a short period of time before returning (e.g. to wait for the script to run) by using the `DurationToWaitForScriptToFinish` property on the start script request

## Tentacle Client

This PR adds the concept of a "Tentacle Client" which is something that knows how to:
- Run a script. Taking care of orchestrating the script execution.
- Upload and download files.

The tentacle client hides away:
- Making retries.
- Cancelling a script if the CancellationToken is triggered or walking away if the script is known to not be started.
- Determining what services the remote tentacle supports e.g. does it support ScriptService v2 and v1 or just v1?

This means clients can just call `ExecuteScript` and not worry about how the script is executed.

## Test improvements

- Adds and expands on Port Forwarder from Halibut allowing for testing of network failures.
- Integration tests are by default run in parallel.
- SeriLog is available in Integration tests and used extensively to give consistent logs.
-  An `IntegrationTest` base class is added which provides that logger and a CancellationToken.
- Test have the nunit timeout attribute added.
- Old tentacles downloaded at test time are now cached in team city. The same version of a tentacle will only be downloaded once in a test run.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


nathanwoctopusdeploy:
[sc-46274]

source/Octopus.Tentacle/Scripts/ScriptStateWorkspace.cs
```diff
@@ -0,0 +1,130 @@
...
+
+            try
+            {
+                if (!Exists(aquireLock: false))
+                {
+                    throw new InvalidOperationException($"ScriptState does not exists at {StateFilePath}");
+                }
+
+                var serialized = SerializeState(state);
+                fileSystem.WriteAllText(StateFilePath, serialized);
```
nathanwoctopusdeploy, line 91: Look at an atomic way to store the state file to guarantee we perform updates

source/Octopus.Tentacle/Scripts/NoopScriptStateWorkspace.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using Octopus.Tentacle.Contracts;
+
+namespace Octopus.Tentacle.Scripts
+{
+    public class NoopScriptStateWorkspace : IScriptStateWorkspace
+    {
+        public bool StateExists => false;
+        public ScriptState Create()
```
LukeButters, line 9: Maybe Create doesn't need to return a ScriptState, that would make the no-op implementation just do nothing.

source/Octopus.Tentacle/Scripts/IScriptStateWorkspace.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace Octopus.Tentacle.Scripts
+{
+    public interface IScriptStateWorkspace
```
LukeButters, line 5: I wonder if this should be named `IStoreScriptState`.

The Workspace I think is just a place we happen to store the state.

source/Octopus.Tentacle/Scripts/NoopScriptStateWorkspace.cs
```diff
@@ -0,0 +1,23 @@
...
+            return new ScriptState(new ScriptTicket(string.Empty));
+        }
+
+        public void Save(ScriptState state)
+        {
+        }
+
+        public ScriptState Load()
+        {
+            return new ScriptState(new ScriptTicket(string.Empty));
```
LukeButters, line 20: It feels dodgy to have an implementation that returns misleading values. It could be a trap someone falls into.

Could this be an in-memory implemented instead?

source/Octopus.Tentacle/Scripts/ScriptStateWorkspace.cs
```diff
@@ -0,0 +1,130 @@
...
+        public ScriptStateWorkspace(ScriptTicket scriptTicket, IScriptWorkspace workspace, IOctopusFileSystem fileSystem)
+        {
+            this.scriptTicket = scriptTicket;
+            this.workspace = workspace;
+            this.fileSystem = fileSystem;
+        }
+
+        string StateFilePath => workspace.ResolvePath("scriptstate.json");
+
+        public bool StateExists => Exists(aquireLock: true);
```
LukeButters, line 26: The Exists method could be renamed `ExistsNoLock()` (Dropping the param) instead since it seems every other calls to exists takes the lock itself. This place this the only place that doesn't take a lock.

source/Octopus.Tentacle/Scripts/ScriptStateWorkspace.cs
```diff
@@ -0,0 +1,130 @@
...
+            try
+            {
+                if (Exists(aquireLock: false))
+                {
+                    throw new InvalidOperationException($"ScriptState already exists at {StateFilePath}");
+                }
+
+                var state = new ScriptState(scriptTicket);
+                var serialized = SerializeState(state);
+                fileSystem.WriteAllText(StateFilePath, serialized);
```
LukeButters, line 41: TODO writes should be atomic

source/Octopus.Tentacle/Services/Scripts/ScriptServiceV2.cs
```diff
@@ -0,0 +1,134 @@
...
+        {
+            this.shell = shell;
+            this.workspaceFactory = workspaceFactory;
+            this.stateWorkspaceFactory = stateWorkspaceFactory;
+            this.log = log;
+        }
+
+        public ScriptTicket StartScript(StartScriptCommand command, ScriptTicket ticket)
+        {
+            var workspace = workspaceFactory.PrepareWorkspace(command, ticket);
```
LukeButters, line 32: Should this method ensure that it is only called once? Prepare workspace looks like something that on a second call would override the files it already wrote down.

Perhaps it would make sense for this class to have some sort of lock[s] controlling concurrency to the same ScriptTicket.

If this is called again with the same ScriptTicket while this method is being run with that same script ticket the second should just take the result of the first. :D

source/Octopus.Tentacle/Services/Scripts/ScriptServiceV2.cs
```diff
@@ -0,0 +1,134 @@
...
+            var stateWorkspace = GetScriptStateWorkspace(ticket, workspace);
+
+            // If the state already exists then this script is already running/has already run and we should not run it again
+            if (stateWorkspace.StateExists)
+            {
+                var state = stateWorkspace.Load();
+                if (state.HasStarted())
```
LukeButters, line 39: So if the script has not started this will try to start again?

LukeButters (approved):

LukeButters (approved):

LukeButters (dismissed):

source/Octopus.Tentacle/Services/Scripts/ScriptService.cs
```diff
@@ -125,6 +72,19 @@ public ScriptStatusResponse CompleteScript(CompleteScriptCommand command)
             return response;
         }
 
+        RunningScript LaunchShell(ScriptTicket ticket, string serverTaskId, IScriptWorkspace workspace, CancellationTokenSource cancel)
```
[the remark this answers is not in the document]
nathanwoctopusdeploy: Any ideas on where / what to call it @LukeButters

source/Octopus.Tentacle.Contracts/StartScriptCommandV2.cs
```diff
@@ -0,0 +1,94 @@
...
+            string? isolationMutexName,
+            string[] arguments,
+            string taskId,
+            ScriptTicket scriptTicket,
+            TimeSpan durationToWaitForScriptToFinish)
+        {
+            Arguments = arguments;
+            TaskId = taskId;
+            ScriptTicket = scriptTicket;
+            DurationToWaitForScriptToFinish = durationToWaitForScriptToFinish;
```
nathanwoctopusdeploy, line 23: This is a new property to optionally allow us to wait for a duration for the script to finish in the StartScript call.

This will allow us to optimise for short scripts e.g. StartScript(wait max 5 seconds) -> CompleteScript dropping hte need to poll with GetStatus

source/Octopus.Tentacle.Tests/Capabilities/CapabilitiesServiceFixture.cs
```diff
@@ -12,8 +12,7 @@ public void CapabilitiesAreReturned()
             new CapabilitiesService()
                 .GetCapabilities()
                 .SupportedCapabilities
-                .Count
-                .Should().Be(0);
+                .Should().BeEquivalentTo(new[]{ "IScriptServiceV2Alpha" });
```
nathanwoctopusdeploy, line 15: `IScriptServiceV2` is being added as `Alpha` to allow us to add the service and test it with branch instances of Server while not making it GA until we are happy with the functionality.

source/Octopus.Tentacle/Scripts/RunningScript.cs
```diff
@@ -53,37 +67,51 @@ public void Execute()
...
+            }
+
+            if (stateRepository != null)
+            {
+                var scriptState = stateRepository.Load();
+                scriptState.Complete(exitCode);
+                stateRepository.Save(scriptState);
             }
+
+            ExitCode = exitCode;
```
nathanwoctopusdeploy, line 107: Refactor to ensure we dispose of the script writer before we set the exit code and state

source/Octopus.Tentacle/Scripts/ScriptWorkspaceFactory.cs
```diff
@@ -86,12 +86,16 @@ void SaveFileToDisk(IScriptWorkspace workspace, ScriptFile scriptFile)
...
-
         string FindWorkingDirectory(ScriptTicket ticket)
         {
-            var work = fileSystem.GetFullPath(Path.Combine(home.HomeDirectory ?? "", "Work", ticket.TaskId));
+            var work = GetWorkingDirectoryPath(ticket);
             fileSystem.EnsureDirectoryExists(work);
             return work;
         }
+
+        public string GetWorkingDirectoryPath(ScriptTicket ticket)
```
nathanwoctopusdeploy, line 96: Refactored for a test

source/Octopus.Tentacle/Services/Capabilities/CapabilitiesService.cs
```diff
@@ -9,7 +9,7 @@ public class CapabilitiesService : ICapabilitiesService
     {
         public CapabilitiesResponse GetCapabilities()
         {
-            return new CapabilitiesResponse(new List<string>());
+            return new CapabilitiesResponse(new List<string>{ nameof(IScriptServiceV2)+"Alpha" });
```
nathanwoctopusdeploy, line 12: `IScriptServiceV2` is being added as `Alpha` to allow us to add the service and test it with branch instances of Server while not making it GA until we are happy with the functionality.

source/Octopus.Tentacle/Services/Capabilities/CapabilitiesService.cs
```diff
@@ -9,7 +9,7 @@ public class CapabilitiesService : ICapabilitiesService
     {
         public CapabilitiesResponse GetCapabilities()
         {
-            return new CapabilitiesResponse(new List<string>());
+            return new CapabilitiesResponse(new []{ nameof(IScriptServiceV2)+"Alpha" });
```
nathanwoctopusdeploy, line 12: This needs fixing in another PR but there is an issue in server using the capability service due to deserialization

source/Octopus.Tentacle.Tests/Integration/StartScriptCommandV2Builder.cs
```diff
@@ -0,0 +1,118 @@
...
+            return this;
+        }
+
+        public StartScriptCommandV2Builder WithScriptBody(string windowsScript, string bashScript)
+        {
+            this.scriptBody = new StringBuilder(PlatformDetection.IsRunningOnWindows ? windowsScript : bashScript);
+            return this;
+        }
+
+        public StartScriptCommandV2Builder WithReplacementInScriptBody(string oldValue, string newValue)
```
LukeButters, line 35: Does this method need to exist here?

If it does, I wonder if it is worth a comment saying this needs to be called after the script is set. It feels like a trap since you could do:

```
WithReplacementInScriptBody("foo", "bar")
.WithScriptBody("foo")
```

And be left with foo.

nathanwoctopusdeploy: This is a copy of the original builder

source/Octopus.Tentacle.Tests/Integration/StartScriptCommandV2Builder.cs
```diff
@@ -0,0 +1,118 @@
...
+
+        public StartScriptCommandV2Builder WithScriptBody(string scriptBody)
+        {
+            this.scriptBody = new StringBuilder(scriptBody);
+            return this;
+        }
+
+        public StartScriptCommandV2Builder WithScriptBody(string windowsScript, string bashScript)
+        {
+            this.scriptBody = new StringBuilder(PlatformDetection.IsRunningOnWindows ? windowsScript : bashScript);
```
LukeButters, line 31: Would this method be useful? It is possible for the thing building the `StartScriptCommandV2` to be on a different OS to the remote target.

nathanwoctopusdeploy: This is a copy of the original builder

nathanwoctopusdeploy: It is being used?

source/Octopus.Tentacle/Scripts/RunningScript.cs
```diff
@@ -53,37 +67,51 @@ public void Execute()
                             token,
                             log))
                         {
-                            RunScript(shellPath, writer);
+                            State = ProcessState.Running;
+
+                            if (stateRepository != null)
```
LukeButters, line 72: This entire if block could perhaps be re-factored out to a method `RecordScriptHasStarted()`

source/Octopus.Tentacle/Scripts/RunningScript.cs
```diff
@@ -53,37 +67,51 @@ public void Execute()
...
             }
             catch (Exception)
             {
                 // Something went really really wrong, probably creating or writing to the log file (Disk space)
-                ExitCode = ScriptExitCodes.FatalExitCode;
-                State = ProcessState.Complete;
+                exitCode = ScriptExitCodes.FatalExitCode;
+            }
+
+            if (stateRepository != null)
```
LukeButters, line 100: This could also be factored out into a little method.

source/Octopus.Tentacle/Scripts/RunningScript.cs
LukeButters: To be super duper safe we mark the script as complete the ExitCode and state should be put into a try finally

source/Octopus.Tentacle/Scripts/RunningScript.cs
```diff
@@ -53,37 +67,51 @@ public void Execute()
...
+            {
+                var scriptState = stateRepository.Load();
+                scriptState.Complete(exitCode);
+                stateRepository.Save(scriptState);
             }
+
+            ExitCode = exitCode;
+            State = ProcessState.Complete;
```
LukeButters, line 108:
```suggestion
                exitCode = ScriptExitCodes.FatalExitCode;
            } finally {
               // No matter what the script is not complete, mark it as such.
               try {
                    if (stateRepository != null)
                    {
                        var scriptState = stateRepository.Load();
                        scriptState.Complete(exitCode);
                        stateRepository.Save(scriptState);
                    }
               } finally {
                   // Don't trust that the stateRepository and so place this code in yet another finally.
                  ExitCode = exitCode;
                  State = ProcessState.Complete;
               }
            } 
```

source/Octopus.Tentacle/Scripts/ScriptState.cs
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        public bool HasCompleted()
+        {
+            return State == ProcessState.Complete;
+        }
+
+        public void Start()
+        {
+            Started = DateTimeOffset.UtcNow;
```
LukeButters, line 32: utc very nice choice :D

source/Octopus.Tentacle/Scripts/ScriptStateStore.cs
```diff
@@ -0,0 +1,140 @@
...
+
+            try
+            {
+                if (!ExistsNoLock())
+                {
+                    throw new InvalidOperationException($"ScriptState does not exists at {StateFilePath}");
+                }
+
+                var serialized = SerializeState(state);
+                using var writer = GetStreamWriter(FileMode.Create);
```
LukeButters, line 91: This write is not atomic, we could reduce the chance of a corrupt state file by writing to a tmp file next to the state file then moving it into place.

source/Octopus.Tentacle/Services/Scripts/ScriptServiceV2.cs
```diff
@@ -0,0 +1,163 @@
...
+
+                runningScript.Process = process;
+                runningScript.CancellationTokenSource = cancellationTokenSource;
+
+                if (command.DurationToWaitForScriptToFinish > TimeSpan.Zero)
+                {
+                    var waited = Stopwatch.StartNew();
+                    while (process.State != ProcessState.Complete && waited.Elapsed < command.DurationToWaitForScriptToFinish)
+                    {
+                        Thread.Sleep(TimeSpan.FromMilliseconds(100));
```
LukeButters, line 84: This could be dropped even lower, I would go 10ms or less. The amount of CPU cycles it takes to wake up is going to be way less than the halibut request responses sent if this new feature was not used.

source/Octopus.Tentacle/Services/Scripts/ScriptServiceV2.cs
```diff
@@ -0,0 +1,163 @@
...
+            var runningScript = runningScripts.GetOrAdd(
+                command.ScriptTicket,
+                s =>
+                {
+                    var workspace = workspaceFactory.GetWorkspace(command.ScriptTicket);
+                    var scriptState = scriptStateStoreFactory.Get(s, workspace);
+                    return new RunningScriptWrapper(scriptState);
+                });
+
+            using (runningScript.ScriptStateStore.GetExclusiveLock())
```
LukeButters, line 43: Should this lock be be in the `ScriptStateStore`? Is the lock tied to the store?

I wonder if instead the lock should live in `runningScripts` inside of `RunningScriptWrapper`.  It could be called `StartScriptMutex`.

source/Octopus.Tentacle/Util/SemaphoreSlimReleaser.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using System.Threading;
+
+namespace Octopus.Tentacle.Util
+{
+    internal class SemaphoreSlimReleaser : IDisposable
```
LukeButters, line 6: Consider making an extension method like:
```
public static IDisposable Lock(this SemaphoreSlim semaphore) {
  semaphire.acquire();
   return new SemaphoreSlimReleaser(semaphore);
}
```

You can use it like:

```
using var _ = sempahore.Lock();
code here 
is under 
the lock
```

source/Octopus.Tentacle.Tests/Integration/StartScriptCommandV2Builder.cs
LukeButters: oh sorry I see its in the integration tests project! ok makes sense.

LukeButters: Perhaps a name like `WithScriptBodyForCurrentOS` would make it super clear.

source/Octopus.Tentacle/Scripts/ScriptStateStore.cs
nathanwoctopusdeploy: Good spot, I've overlooked this change we discussed somehow!!

source/Octopus.Tentacle/Services/Scripts/ScriptServiceV2.cs
nathanwoctopusdeploy: Yeh makes sense 👍

source/Octopus.Tentacle.Tests/Integration/ScriptServiceV2Fixture.cs
```diff
@@ -0,0 +1,465 @@
...
+
+        [Test]
+        public async Task StartScriptShouldNotWaitForALongScriptToFinish()
+        {
+            var bashScript = "sleep 10";
+            var windowsScript = "Start-Sleep -Seconds 10";
+
+            var startScriptCommand = new StartScriptCommandV2Builder()
+                .WithScriptBodyForCurrentOs(windowsScript, bashScript)
+                .WithDurationStartScriptCanWaitForScriptToFinish(TimeSpan.FromSeconds(5))
```
LukeButters, line 191: Consider dropping the wait time to 1s, tests that work based on timings can be flaky anything we can do to mitigate that is worthwhile.

LukeButters (approved):

source/Octopus.Tentacle.Tests/Integration/ScriptServiceV2Fixture.cs
```diff
@@ -0,0 +1,460 @@
...
+            finalResponse.State.Should().Be(ProcessState.Complete);
+            finalResponse.ExitCode.Should().Be(0);
+            logs.Count.Should().BeGreaterThan(1);
+        }
+
+        [Test]
+        public async Task ShouldReturnANonZeroExitCodeForAFailingScript()
+        {
+            var windowsScript = "& ping.exe nope -n 1";
+            var bashScript = "ping nope -c 1";
```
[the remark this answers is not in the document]
nathanwoctopusdeploy: This was copied from the original tests!!

source/Octopus.Tentacle/Scripts/ScriptStateStore.cs
```diff
@@ -0,0 +1,111 @@
...
+            this.scriptTicket = scriptTicket;
+            this.workspace = workspace;
+            this.fileSystem = fileSystem;
+        }
+
+        string StateFilePath => workspace.ResolvePath("scriptstate.json");
+
+        public ScriptState Create()
+        {
+            using var _ = scriptStateLock.Lock();
```
pawelpabich, line 28: Just curious. We are using synchronous primitives here. Would `lock(object)`, which doesn't require a customer releaser, work here? Are you preparing to make Tentacle Async?

source/Octopus.Tentacle/Scripts/ScriptState.cs
```diff
@@ -0,0 +1,44 @@
...
+namespace Octopus.Tentacle.Scripts
+{
+    public class ScriptState
+    {
+        public ScriptState(ScriptTicket scriptTicket)
+        {
+            ScriptTicket = scriptTicket;
+        }
+
+        public ScriptTicket ScriptTicket { get; set; }
```
pawelpabich, line 13: Consider making the setters private so there is no way to workaround the logic encapsulated in methods like `Complete`.

source/Octopus.Tentacle/Scripts/ScriptState.cs
```diff
@@ -0,0 +1,44 @@
...
+            return State == ProcessState.Complete;
+        }
+
+        public void Start()
+        {
+            Started = DateTimeOffset.UtcNow;
+            State = ProcessState.Running;
+        }
+
+        public void Complete(int exitCode, bool ranToCompletion = true)
```
pawelpabich, line 36: 👍Good encapsulation.

source/Octopus.Tentacle/Scripts/ScriptStateStoreFactory.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using Octopus.Tentacle.Contracts;
+using Octopus.Tentacle.Util;
+
+namespace Octopus.Tentacle.Scripts
+{
+    internal class ScriptStateStoreFactory : IScriptStateStoreFactory
```
pawelpabich, line 7: It looks like this custom code does what Autofac delegate factory does: https://autofac.readthedocs.io/en/latest/advanced/delegate-factories.html

source/Octopus.Tentacle/Services/Scripts/ScriptServiceV2.cs
```diff
@@ -0,0 +1,165 @@
...
+                    runningScript.ScriptStateStore.Create();
+                }
+
+                var cancellationTokenSource = new CancellationTokenSource();
+                var process = LaunchShell(command.ScriptTicket, command.TaskId, workspace, runningScript.ScriptStateStore, cancellationTokenSource);
+
+                runningScript.Process = process;
+                runningScript.CancellationTokenSource = cancellationTokenSource;
+
+                if (command.DurationToWaitForScriptToFinish > TimeSpan.Zero)
```
pawelpabich, line 80: Consider making `DurationToWaitForScriptToFinish ` nullable to avoid relying on magic values like 0 which are open to interpretation.

source/Octopus.Tentacle/Services/Scripts/ScriptServiceV2.cs
```diff
@@ -0,0 +1,165 @@
...
+            var scriptLog = runningScript?.ScriptLog ?? workspace.CreateLog();
+            var logs = scriptLog.GetOutput(lastLogSequence, out var next);
+
+            if (runningScript != null)
+            {
+                return new ScriptStatusResponseV2(ticket, runningScript.State, runningScript.ExitCode, logs, next);
+            }
+
+            // If we don't have a RunningProcess we check the ScriptStateStore to see if we have persisted a script result
+            var scriptStateStore = scriptStateStoreFactory.Get(ticket, workspace);
```
pawelpabich, line 135: `ScriptStateStoreFactory.Get` never returns null. Is there a reason to use `?` here?

source/Octopus.Tentacle.Contracts/CancelScriptCommandV2.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+
+namespace Octopus.Tentacle.Contracts
+{
+    public class CancelScriptCommandV2
```
pawelpabich, line 5: We version the whole interfaces, not individual operations, so the number of classes under `Octopus.Tentacle.Contracts` can grow quickly.  Did you consider putting all `IScriptServiceV2` classes in a subfolder?

pawelpabich (approved):
The overall structure makes sense to me. I added a few ideas for you to consider.

The biggest problem with multiple versions of  a service is that at the end of the day, all versions use the same building blocks, which requires us to sprinkle them with a decent dose of version-specific `IFs`.  Let's keep thinking about this problem, without blocking this PR,  so we can avoid modifying the existing code as much as possible.

source/Octopus.Tentacle/Services/Scripts/ScriptServiceV2.cs
nathanwoctopusdeploy: There is not. I suspect it's due to some significant refactoring during development. Good spot, will clean up 👍

source/Octopus.Tentacle/Scripts/ScriptStateStore.cs
nathanwoctopusdeploy: It wasn't intentionally done to support any future changes.

source/Octopus.Tentacle/Scripts/ScriptStateStoreFactory.cs
nathanwoctopusdeploy: Would this require the consumer to take a dependency on `IOctopusFileSystem` so it could pass it to the delegate factory?

source/Octopus.Tentacle.Contracts/CancelScriptCommandV2.cs
nathanwoctopusdeploy: Yeh good idea 👍

source/Octopus.Tentacle/Scripts/ScriptState.cs
nathanwoctopusdeploy: Good spot 👍

source/Octopus.Tentacle/Scripts/ScriptStateStoreFactory.cs
pawelpabich: No, that's the beauty of this solution. Autofac will inject what it can. E.g. HubClient takes 11 dependencies and but the client needs to provide only 3 of them: https://github.com/OctopusDeploy/CloudPortal/blob/7d905aac2bb4826a89d879a5d19f0cd29ec3c950/source/CloudPortal.Core/Hub/HubClient.cs#L38

nathanwoctopusdeploy: Ok I'll revisit this. 👍

source/Octopus.Tentacle.Tests.Integration/ScriptServiceV2IntegrationTest.cs
```diff
@@ -0,0 +1,175 @@
...
+                allLogs.Should().Contain("hello");
+            }
+        }
+
+        private static string JoinLogs(List<ProcessOutput> logs)
+        {
+            return String.Join(" ", logs.Select(l => l.Text).ToArray());
+        }
+
+        // [Test]
```
LukeButters, line 61: This requires https://github.com/OctopusDeploy/OctopusTentacle/pull/480/files

source/Octopus.Tentacle.Tests/Client/RpcCallRetryHandlerFixture.cs
```diff
@@ -0,0 +1,576 @@
...
+            catch (HalibutClientException) { }
+
+            stopWatch.Elapsed.Should().BeGreaterOrEqualTo(TimeSpan.FromSeconds(9)).And.BeLessThan(TimeSpan.FromSeconds(20));
+            callCount.Should().BeGreaterThan(1);
+        }
+
+
+        [TestCase(TimeoutStrategy.Optimistic)]
+        [TestCase(TimeoutStrategy.Pessimistic)]
+        public async Task CancelsTheExecutingActionIfItIsARetryAfterTheTimeout(TimeoutStrategy timeoutStrategy)
```
LukeButters, line 228: This failed here https://build.octopushq.com/buildConfiguration/TeamFireAndMotion_OctopusTentacleNet48_SensibleDefaultsUnitTestLinux/7821407?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandPull+Request+Details=true&expandBuildChangesSection=true&expandBuildProblemsSection=true&expandBuildTestsSection=true

source/Octopus.Tentacle.Tests.Integration/ScriptServiceV2IntegrationTest.cs
```diff
@@ -0,0 +1,175 @@
...
+            }
+        }
+
+        private static string JoinLogs(List<ProcessOutput> logs)
+        {
+            return String.Join(" ", logs.Select(l => l.Text).ToArray());
+        }
+
+        [Test]
+        public async Task WhenTentacleRestartsWhileRunningAScript_TheExitCodeShouldBe_UnknownResultExitCode()
```
LukeButters, line 62: This has failures, likely because of existing bugs in halibut.

https://build.octopushq.com/buildConfiguration/TeamFireAndMotion_OctopusTentacleNet48_SensibleDefaultsChainFullChain/7828186?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&expandBuildProblemsSection=true&expandPull+Request+Details=true&expandBuildTestsSection=true

To fix just retry the call after the tentacle is restarted

nathanwoctopusdeploy: When I rebased I lost the `nameof(IScriptServiceV2) + "Alpha"` in the capability service v2 so no retries are happening

LukeButters: I fixed that issue up since that was failing a test which verified retries.

This is unrelated to that as this one doesn't use the tentacle client and instead calls the services directly.

source/Octopus.Tentacle.Contracts/Octopus.Tentacle.Contracts.csproj
```diff
@@ -28,7 +28,7 @@
     </Otherwise>
   </Choose>
   <ItemGroup>
-    <PackageReference Include="Halibut" Version="5.0.464" />
+    <PackageReference Include="Halibut" Version="6.0.57" />
```
nathanwoctopusdeploy, line 31: Bump to latest?

source/Octopus.Tentacle.Tests.Integration/Support/PollingTentacleBuilder.cs
```diff
@@ -164,9 +165,10 @@ private void RunTentacleCommand(string tentacleExe, string[] args, TemporaryDire
             Action<string> CommandOutput,
             CancellationToken cancellationToken)
         {
+            var log = new SerilogLoggerBuilder().Build().ForContext<RunningTestTentacle>();
             Action<string> allOutput = s =>
             {
-                TestContext.WriteLine(s);
+                log.Fatal("[Tentacle] " +s );
```
nathanwoctopusdeploy, line 171: `Fatal`??

source/Octopus.Tentacle.Tests.Integration/Util/SerilogLoggerBuilder.cs
```diff
@@ -12,10 +16,36 @@ public class SerilogLoggerBuilder
...
+            var sink = PlatformDetection.IsRunningOnNix ? (ILogEventSink)new NUnitLinuxSink(new MessageTemplateTextFormatter(OutputTemplate)) : new NUnitSink(new MessageTemplateTextFormatter(OutputTemplate));
             return new LoggerConfiguration()
                 .MinimumLevel.Debug()
-                .WriteTo.Sink(new NUnitSink(new MessageTemplateTextFormatter(OutputTemplate)))
+                .WriteTo.Sink(sink)
                 .CreateLogger();
         }
+        
+        
+        public class NUnitLinuxSink : ILogEventSink
```
nathanwoctopusdeploy, line 28: What is this needed for?

source/Octopus.Tentacle.Client/Retries/RpcCallRetryHandler.cs
```diff
@@ -0,0 +1,102 @@
...
+            this.timeoutStrategy = timeoutStrategy;
+            RetryTimeout = retryTimeout;
+        }
+
+        public TimeSpan RetryTimeout { get; }
+
+        public async Task<T> ExecuteWithRetries<T>(
+            Func<CancellationToken, Task<T>> action,
+            Func<Exception, TimeSpan, int, TimeSpan, CancellationToken, Task>? onRetryAction,
+            Func<TimeSpan, CancellationToken, Task>? onTimeoutAction,
```
LukeButters, line 25: @nathanwoctopusdeploy it isn't super obvious what each of the params to the Funcs are. E.g. what is the timespan?

Could these be delegates so that the parameters can be named and or documented.

source/Octopus.Tentacle.Client/Scripts/ScriptExecutionOrchestrator.cs
```diff
@@ -0,0 +1,300 @@
...
+                        ct =>
+                        {
+                            using (ct.Register(connectCancellationTokenSource.TryCancel))
+                            {
+                                return scriptServiceV2.StartScript(startScriptCommand);
+                            }
+                        },
+                        logger,
+                        cancellationToken,
+                        abandonActionOnCancellation: false); // TODO - Should be true
```
LukeButters, line 74: @nathanwoctopusdeploy  Is it worth noting why its true or false?

source/Octopus.Tentacle.Tests.Integration/ClientFileTransferRpcRetryTimeoutsWork.cs
```diff
@@ -0,0 +1,157 @@
...
+                .WithPortForwarderDataLogging()
+                .WithResponseMessageTcpKiller(out var responseMessageTcpKiller)
+                .WithTentacleServiceDecorator(new TentacleServiceDecoratorBuilder()
+                    .LogCallsToFileTransferService()
+                    .CountCallsToFileTransferService(out var fileTransferServiceCallCounts)
+                    .RecordExceptionThrownInFileTransferService(out var fileTransferServiceException)
+                    .DecorateFileTransferServiceWith(d => d
+                        .BeforeUploadFile((service, _, _) =>
+                        {
+                            service.EnsureTentacleIsConnectedToServer(Logger);
```
LukeButters, line 44: @nathanwoctopusdeploy Should this be within the `if == null` just below? Otherwise I think this will be trying to ensure the tentacle is connected possibly when the port-forwareder is down.

source/Octopus.Tentacle.Tests.Integration/Support/RunningTentacle.cs
```diff
@@ -54,11 +59,41 @@ public async Task Stop(CancellationToken cancellationToken)
...
+                cancellationTokenSource.Cancel();
+                cancellationTokenSource.Dispose();
+                cancellationTokenSource = null;
+            }
+
+            var t = runningTentacleTask;
+            runningTentacleTask = null;
+
+            var stopDuration = Stopwatch.StartNew();
+            while (t?.IsCompleted == false && stopDuration.Elapsed < TimeSpan.FromSeconds(10))
```
LukeButters, line 75: @nathanwoctopusdeploy could this be:
```
await When.Any(t, Task.Delay(10);
```
with a try catch to ignore any errors?

source/Octopus.Tentacle/Scripts/ScriptStateStore.cs
```diff
@@ -0,0 +1,111 @@
...
+
+            var serialized = SerializeState(state);
+            var tempFilePath = workspace.ResolvePath(Guid.NewGuid().ToString());
+            var backupPath = workspace.ResolvePath(Guid.NewGuid().ToString());
+            using (var writer = GetStreamWriter(tempFilePath, FileMode.Create))
+            {
+                writer.Write(serialized);
+            }
+
+            File.Replace(tempFilePath, StateFilePath, backupPath, true);
```
LukeButters, line 81: very nice @nathanwoctopusdeploy

source/Octopus.Tentacle.Tests.Integration/Support/RunningTentacle.cs
nathanwoctopusdeploy: no as it's not async and I wanted to avoid sync to async in the dispose

nathanwoctopusdeploy (commented):
LGTM

source/Octopus.Tentacle.Client/ITentacleClient.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Threading.Tasks;
+using Halibut;
+using Octopus.Diagnostics;
+using Octopus.Tentacle.Client.Scripts;
+using Octopus.Tentacle.Contracts;
+using Octopus.Tentacle.Contracts.ScriptServiceV2;
+
+namespace Octopus.Tentacle.Client
+{
+    public interface ITentacleClient : IDisposable
```
pawelpabich, line 12: As mentioned [here](https://octopusdeploy.slack.com/archives/C04JFJXECS1/p1685666412563579), the Tentacle Client interface might need to be tweaked to be useful for reliable deployments. What is our versioning strategy here? Can we easily evolve it?

source/Octopus.Tentacle.Contracts/ClientServices/IClientCapabilitiesServiceV2.cs
```diff
@@ -0,0 +1,10 @@
+using Halibut.ServiceModel;
+using Octopus.Tentacle.Contracts.Capabilities;
+
+namespace Octopus.Tentacle.Client.ClientServices
+{
+    public interface IClientCapabilitiesServiceV2
+    {
+        public CapabilitiesResponseV2 GetCapabilities(HalibutProxyRequestOptions halibutProxyRequestOptions);
```
pawelpabich, line 8: I'm struggling to wrap my head around Client services. It's unclear to me what problem they are trying to solve and how they differ from the services they map to. 

Also, it is not clear to me why we have introduced `halibutProxyRequestOptions`  on the client services instead of adding a cancellation token to the V2 interface and how would they work with cancellation token that we will introduce as part of Async Halibut work.

source/Octopus.Tentacle.Tests.Integration/Support/TentacleVersions.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+
+namespace Octopus.Tentacle.Tests.Integration.Support
+{
+    public static class TentacleVersions
+    {
+        public const string v5_0_4_FirstLinuxRelease = "5.0.4"; // First linux Release 9/9/2019
```
pawelpabich, line 7: 👍 nice naming convention.

source/Octopus.Tentacle.Client/TentacleClient.cs
```diff
@@ -0,0 +1,117 @@
...
+        readonly RpcCallRetryHandler rpcCallRetryHandler;
+        readonly IClientScriptService scriptServiceV1;
+        readonly IClientScriptServiceV2 scriptServiceV2;
+        readonly IClientFileTransferService fileTransferServiceV1;
+        readonly IClientCapabilitiesServiceV2 capabilitiesServiceV2;
+
+        public TentacleClient(ServiceEndPoint serviceEndPoint,
+            IHalibutRuntime halibutRuntime,
+            IScriptObserverBackoffStrategy scriptObserverBackOffStrategy,
+            ITentacleServiceDecorator? tentacleServicesDecorator,
```
pawelpabich, line 31: If the main reason for having this decorator is observability, then could we expose something like `OnRemovtCall` event and decouple it from the services that Tentacle Client uses behind the scenes? 

E.g. https://github.com/OctopusDeploy/OctopusClients/blob/master/source/Octopus.Server.Client/IOctopusAsyncClient.cs#L43-L48

pawelpabich (commented):
I mostly focused on the new Tentacle Client. I like its simple to use interface but a few things are not clear to me and I would have to clarify them before we merge this PR.

source/Octopus.Tentacle.Client/ITentacleClient.cs
nathanwoctopusdeploy: Do we need to version the TentacleClient? It will be "versioned" through the nuget package version. If it evolves / changes and clients want to make use of the new changes then they will need to bump the package version they reference and make any changes to align with the new client. 

Old clients won't stop working as we will maintain backward compatibility in Tentacle (as we currently do).

source/Octopus.Tentacle.Contracts/ClientServices/IClientCapabilitiesServiceV2.cs
LukeButters: This Halibut [PR](https://github.com/OctopusDeploy/Halibut/pull/238) may help answer some of those questions.

nathanwoctopusdeploy: A change has recently been made to Halibut to try and improve connecting cancellation tokens. This is a cancellation token passed to a Halibut Client that is honoured when a client is trying to connect to a listening tentacle or a polling request has been queued but not yet dequeued by tentacle. After those events, and when the request is "in-flight" the connecting cancellation token is no longer used.

The issue we previously had was that this connecting cancellation token was provided to the creator of the Client so a single cancellation token would control every RPC call made using that client. Once the connecting cancellation token was cancelled, the Clients was unusable for new requests (as they would immediately cancel).

We had an implementation using this Client wide Cancellation Token but it was messy, we had to have a script service v2 with a Cancellation token and then a second script service v2 without a cancellation token so we could continue to make RPC calls even after the Execution of the Script had been cancelled and the connecting cancellation token cancelled.

The current TentacleRemoteEndpointFacade and ScriptObserver does not suffer from this issue because it is not very good at cancelling requests when the execution is cancelled. We have improved on this and hit this issue. 

Client Services were introduced so that the interface for the client and the interface for the Service could diverge and allow additional parameters to be passed to the client that doesn't go over the wire to the service.  Example the connecting cancellation token. We now allow it to be passed for each RPC call but this can't/shouldn't go over the wire to Tentacle, it's client side only.

HalibutProxyRequestOptions is a container for the client specific parameters e.g. connect cancellation token.

nathanwoctopusdeploy:
+semver: minor

source/Octopus.Tentacle.Client/TentacleClient.cs
nathanwoctopusdeploy: As discussed we will revisit this after this PR merges to hide the Service Decorator from the client and keep it internal to the Tentacle solution

nathanwoctopusdeploy:

PR #467, opened 2023-04-28
Fixes failing service installation tests on Linux

geofflamrock:
# Background

Two integration tests have been failing for Linux since the agent on which the tests run got updated from Ubuntu 16 to 18.
- CanInstallServiceAsRoot
- CanInstallServiceAsUser

These have been failing with an error message "System.IndexOutOfRangeException : Index was outside the bounds of the array.". The tests attempt to install the tentacle service and then check a bunch of properties using `systemctl`. Changing to Ubuntu 18 looks like its added in a couple of properties that have an empty string as their value:

image

The test hasn't handled these well and caused the failure.

# Results

This PR fixes the test so that it allows values with empty strings to return in the property set that is gathered from the installed service. We only check a few specific properties, so allowing others that we don't care about to have empty strings so that we don't fail the test shouldn't impact the behaviour of the test.

Shortcut: [sc-37201]

## Before

Tests fail with index out of bounds error

image

## After

Tests continue as expected and pass

image

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


michaelongso (approved):
Tested the change locally and it was 🟢 . LGTM 👍

PR #470, opened 2023-05-03
Adds just enough to run tentacle out of proc as a polling tentacle.

LukeButters:
# Background

Adds the ability to test end to end a polling tentacle with a tentacle client.

It is now possible to write a test which:
* configures a polling tentacle.
* Runs the tentacle (calling the real tentacle executable)
* Makes RPC calls to that tentacle via halibut.

[SC-46171]

More work in this area is to be done including adding support for listening tentacles, writing a test that actually tests the services, adding the ability to test against old tentacles. All arriving soon in a PR near you!

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Octopus.Tentacle.Tests.Integration/PollingTentacleTest.cs
```diff
@@ -0,0 +1,55 @@
...
+                        .WithServiceUri(tentacleId)
+                        .Build(CancellationToken.None);
+
+                    var res = tentacleClient.ScriptService.GetStatus(new ScriptStatusRequest(new ScriptTicket("1212"), 111));
+                    res.ExitCode.Should().Be(0);
+                }, cts.Token);
+                
+                await Task.WhenAny(runningTentacle, testTask);
+                cts.Cancel();
+                await Task.WhenAll(runningTentacle, testTask);
```
LukeButters, line 51: The octopus server tests get around this `await whrn any; await when all` by waiting for tentacle to print a particular message depending on if it is a polling or listening. I guess we could do that here. Thoughts?

acodrington (approved):
Everything looks good, as we worked on and discussed! :shipit:

PR #472, opened 2023-05-05
Adds support for testing against existing versions of tentacles

LukeButters:
# Background

We are going to make changes to tentacle and need clients to still be able to talk to older tentacle.

This adds the ability to wrote those tests, which will be used in a future PR which adds a working capability service.

[SC-46171]

# Results

An existing version of tentacle can now be downloaded unpacked and run in a test.

The tentacle is cached locally for speedy development locally.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Tests.Integration/Support/LinuxTentacleFetcher.cs
+++ b/source/Octopus.Tentacle.Tests.Integration/Support/LinuxTentacleFetcher.cs
@@ -1,0 +1,67 @@
+using System;
+using System.IO;
+using System.IO.Compression;
+using System.Threading;
+using System.Threading.Tasks;
+using NUnit.Framework;
+using Octopus.Tentacle.Util;
+
+namespace Octopus.Tentacle.Tests.Integration.Support
+{
+    public class LinuxTentacleFetcher : ITentacleFetcher
+    {
+        static string LinuxDownloadUrlForVersion(string versionString) => $"https://download.octopusdeploy.com/linux-tentacle/tentacle-{versionString}-linux_x64.tar.gz";
+
+        public async Task<string> GetTentacleVersion(string downloadPath, string version)
+        {
+            var directoryPath = Path.Combine(downloadPath, version);
+            if (!Directory.Exists(directoryPath))
+            {
+                Directory.CreateDirectory(directoryPath);
+            }
+
+            return await DownloadAndExtractFromUrl(directoryPath, LinuxDownloadUrlForVersion(version));
+        }
+
+        static async Task<string> DownloadAndExtractFromUrl(string directoryPath, string url)
+        {
+            var downloadFilePath = Path.Combine(directoryPath, Guid.NewGuid().ToString("N"));
+
+            TestContext.WriteLine($"Downloading {url} to {downloadFilePath}");
+            await OctopusPackageDownloader.DownloadPackage(url, downloadFilePath);
+
+            var extractionDirectory = new DirectoryInfo(Path.Combine(directoryPath, "extracted"));
+            ZipFile.ExtractToDirectory(downloadFilePath, extractionDirectory.FullName);
+
+            ExtractTarGzip(downloadFilePath, extractionDirectory.FullName);
+            return Path.Combine(extractionDirectory.FullName, "tentacle", "Tentacle");
+        }
+
+        public static void ExtractTarGzip(string gzArchiveName, string destFolder)
+        {
+            if (!Directory.Exists(destFolder))
+            {
+                Directory.CreateDirectory(destFolder);
+            }
+
+            // We need to use tar directly rather than a C# implementation
+            // All C# implementations that we have tried here have not been able to preserve the file permissions stored in the archive
+            // In practice, this means that after extraction the executables don't have the executable bit.
+            // Falling back to good old fashioned `tar` does the job nicely :)
+            using var tmp = new TemporaryDirectory();
+            var exitCode = SilentProcessRunner.ExecuteCommand(
+                "tar",
+                $"xzvf {gzArchiveName} -C {destFolder}",
+                tmp.DirectoryPath,
+                TestContext.WriteLine,
+                TestContext.WriteLine,
+                TestContext.WriteLine,
+                CancellationToken.None);
+
+            if (exitCode != 0)
+            {
+                throw new Exception("Error extracting archive");
+            }
+        }
+    }
+}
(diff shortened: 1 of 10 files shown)


source/Octopus.Tentacle.Tests.Integration/Support/LinuxTentacleFetcher.cs
```diff
@@ -0,0 +1,67 @@
...
+            await OctopusPackageDownloader.DownloadPackage(url, downloadFilePath);
+
+            var extractionDirectory = new DirectoryInfo(Path.Combine(directoryPath, "extracted"));
+            ZipFile.ExtractToDirectory(downloadFilePath, extractionDirectory.FullName);
+
+            ExtractTarGzip(downloadFilePath, extractionDirectory.FullName);
+            return Path.Combine(extractionDirectory.FullName, "tentacle", "Tentacle");
+        }
+
+        public static void ExtractTarGzip(string gzArchiveName, string destFolder)
```
LukeButters, line 40: Copied as is from Octopus server repo.

source/Octopus.Tentacle.Tests.Integration/Support/LinuxTentacleFetcher.cs
```diff
@@ -0,0 +1,67 @@
+using System;
+using System.IO;
+using System.IO.Compression;
+using System.Threading;
+using System.Threading.Tasks;
+using NUnit.Framework;
+using Octopus.Tentacle.Util;
+
+namespace Octopus.Tentacle.Tests.Integration.Support
```
LukeButters, line 9: Most of this is copied from octopus server repo.

source/Octopus.Tentacle.Tests.Integration/Support/NugetTentacleFetcher.cs
```diff
@@ -0,0 +1,86 @@
...
+using System.Runtime.InteropServices;
+using System.Threading.Tasks;
+using NUnit.Framework;
+using Octopus.Tentacle.Util;
+
+namespace Octopus.Tentacle.Tests.Integration.Support
+{
+    public class NugetTentacleFetcher : ITentacleFetcher
+    {
+        static string DownloadUrlForVersion(string versionString) => $"https://f.feedz.io/octopus-deploy/dependencies/packages/Octopus.Tentacle.CrossPlatformBundle/{versionString}/download";
```
LukeButters, line 14: For reference the nuget package is 460MB.

source/Octopus.Tentacle.Tests.Integration/Support/NugetTentacleFetcher.cs
```diff
@@ -0,0 +1,86 @@
...
+            }
+
+            if (PlatformDetection.IsRunningOnMac)
+            {
+                return $"tentacle-{GetFramework()}-osx-{Architecture()}.tar.gz";
+            }
+
+            if (PlatformDetection.IsRunningOnNix)
+            {
+                return $"tentacle-{GetFramework()}-linux-{Architecture()}.tar.gz";
```
LukeButters, line 65: This worked for me, although in practice this path wont be hit since the smaller tentacle binaries are already available for use.

source/Octopus.Tentacle.Tests.Integration/Support/NugetTentacleFetcher.cs
```diff
@@ -0,0 +1,86 @@
...
+        public string TentacleArtifactName()
+        {
+            if (PlatformDetection.IsRunningOnWindows)
+            {
+                return $"tentacle-{GetFramework()}-win-{Architecture()}.zip";
+            }
+
+            if (PlatformDetection.IsRunningOnMac)
+            {
+                return $"tentacle-{GetFramework()}-osx-{Architecture()}.tar.gz";
```
LukeButters, line 60: No idea if this will work.

source/Octopus.Tentacle.Tests.Integration/Support/TentacleFetcherFactory.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        private ITentacleFetcher GetBase()
+        {
+            if (PlatformDetection.IsRunningOnNix) return new LinuxTentacleFetcher();
+            return new NugetTentacleFetcher();
+        }
+
+        public ITentacleFetcher Create()
+        {
+            if (TentacleExeFinder.IsRunningInTeamCity())
```
LukeButters, line 16: I wonder if we should have caching in teamcity? My worry was we might find the tests only work because of the cache but we don't find out until way later.

Then again I think the agents are ephemeral. Thoughts?

source/Octopus.Tentacle.Tests.Integration/Support/NugetTentacleFetcher.cs
```diff
@@ -0,0 +1,93 @@
...
+            {
+                return $"tentacle-{TentacleBinaryFrameworkForCurrentOs()}-linux-{Architecture()}.tar.gz";
+            }
+
+            throw new Exception("Wow this is one fancy OS that tentacle probably can't run on");
+        }
+
+        public static string TentacleBinaryFrameworkForCurrentOs()
+        {
+            // This wont work for future versions of dotnet
```
acodrington, line 75: Yeah, it'd be good if we could invert this logic i.e. if its Framework then download `net48`, otherwise whatever the 'last dotnet version is

source/Octopus.Tentacle.Tests.Integration/Support/TentacleFetcherFactory.cs
acodrington: yeah not sure about this one. As you said, it's not worth caching if the agents themselves get spun up and torn down constantly.

Unless perhaps there's some shared storage the agents have access to? Even then, it's only worthwhile if it's signifcantly faster than re-downloading each time.

acodrington (approved):
Yeah looks good. A couple of comments, but nothing to stop the merge. :shipit:

source/Octopus.Tentacle.Tests.Integration/Support/NugetTentacleFetcher.cs
LukeButters: In this case we have to download the entire 460MB nuget package which contains the tentacle bins for all OSs and architectures supported.

Only the linux and mac Tentacle binaries appear to be distributed on their own. I implemented a different ITentacleFetcher  for linux to take advantage of that which means on linux only ~35MB would ever need to be downloaded.

source/Octopus.Tentacle.Tests.Integration/Support/TentacleFetcherFactory.cs
LukeButters: Ah good point.

I was more concerned about us suddenly depending on the cache being already filled. One of those classic someone updated the build server and now it wont build because the thing it used to get some dependency from doesn't exist any more.

I guess nuget should always be safe. I will make a future PR that adds caching for teamcity in.

PR #494, opened 2023-05-24
Solution and Dependency changes

nathanwoctopusdeploy:
Change to the Solution which visual studio makes every time the solution is opened.

Changes to project dependencies to try and resolve an intermittent build issue on Team City


source/Octopus.Tentacle.Tests/Octopus.Tentacle.Tests.csproj
```diff
@@ -29,7 +29,6 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.3.2" />
     <ProjectReference Include="..\Octopus.Tentacle.CommonTestUtils\Octopus.Tentacle.CommonTestUtils.csproj" />
-    <ProjectReference Include="..\Octopus.Tentacle\Octopus.Tentacle.csproj" />
```
acodrington, line 32: The Test project doesn't depend on the app project?! What am I missing here?

nathanwoctopusdeploy: Inherited dependencies

LukeButters (approved):
I guess visual studio knows best.

PR #501, opened 2023-05-26
Sast/script service v2 fix v1scripttests

LukeButters:
# Background

- Fixes tests `ClientScriptExecutionScriptServiceV1IsNotRetried` the issue appears to be that killing tentacle before the script finishes results in the build hanging.
- `RunTestsInParallelLocallyIfEnabledButNeverOnTeamCity` makes misleading as it **always runs tests in parallel** 
    - (TODO will fix in a smaller PR)
    - Test time is now ~5mins rather than ~17mins
- Adds that to more tests.
- `ClientAndTentacle` can now be used to get a `LegacyTentacleClientBuilder`.
- Removed the extra check if a test has over run.
- `TentacleBinaryCache` now uses the same tmp dir as `TempDirectory` which means `move` now works.
- `TentacleBinaryCache` attempts to move the downloaded tentacles, failing back to copy if needed. (Confirmed logs do not contain `Could not move` meaning tentacle was moved into place).
- `ScriptServiceDecoratorBuilder` now allows chaining of decorators.
- 

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Client/Retries/CancellationTokenExtensionMethods.cs
+++ b/source/Octopus.Tentacle.Client/Retries/CancellationTokenExtensionMethods.cs
@@ -1,0 +1,39 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Octopus.Tentacle.Client.Retries
+{
+    internal static class CancellationTokenExtensionMethods
+    {
+        public static Task<TResult> AsTask<TResult>(this CancellationToken cancellationToken)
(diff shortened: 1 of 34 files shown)


LukeButters:
What is going on [here](https://build.octopushq.com/buildConfiguration/TeamFireAndMotion_OctopusTentacleNet48_IntegrationTestNetFramework/7975151?buildTab=log&focusLine=5577&logView=linear)

PR #518, opened 2023-06-09
Updated ENTRYPOINT to exec syntax for Windows Dockerfile to workaround containerd issue

michaelongso:
# Background

[SC-48697](https://app.shortcut.com/octopusdeploy/story/48697/issue-with-windows-tentacle-container-running-in-fargate)

We have a reported issue where Windows Tentacle Container fails to run in ECS Fargate. The issue is linked back to the open [containerd issue](https://github.com/containerd/containerd/issues/5067) when parsing the _shell_ form for `ENTRYPOINT` syntax. The known workaround is to use the [_exec_ form](https://docs.docker.com/engine/reference/builder/#understand-how-cmd-and-entrypoint-interact:~:text=has%20two%20forms%3A-,The%20exec%20form,-%2C%20which%20is%20the).

# Results

Fixes [issue-8143](https://github.com/OctopusDeploy/Issues/issues/8143)

# Behavioural Update from _shell_ to _exec_

The interaction between `ENTRYPOINT` and `CMD` is now changed due to the usage of _exec_ syntax, where additional `CMD` will now be appended as part of the `ENTRYPOINT` arguments. Previously with _shell_ syntax, additional `CMD` is ignored.

See the [docker doc](https://docs.docker.com/engine/reference/builder/#understand-how-cmd-and-entrypoint-interact:~:text=Understand%20how%20CMD%20and%20ENTRYPOINT%20interact%F0%9F%94%97) for a details explanation.

## Before

Running Windows Tentacle Container on ECS Fargate will result in:
```
CannotStartContainerError: CannotStartContainerError: hcs::System::CreateProcess: "cmd /S /C powershell -File ./configure-tentacle.ps1 -Verbose && powershell -File ./run-tentacle.ps1 -Verbose" 0353f39e1acb446bbf43d99137706834-1111712579: The system cannot find the file specified.: unknown
```

## After

Screenshot 2023-06-09 at 12 30 03 pm

Screenshot 2023-06-09 at 12 30 51 pm

## Validation with Azure Container with Windows OS

Screenshot 2023-06-09 at 12 31 03 pm

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/docker/windows/Scripts/configure-and-run-tentacle-wrapper.ps1
+++ b/docker/windows/Scripts/configure-and-run-tentacle-wrapper.ps1
@@ -1,0 +1,2 @@
+﻿& ./configure-tentacle.ps1 -Verbose
+& ./run-tentacle.ps1 -Verbose
--- a/docker/windows/Dockerfile
+++ b/docker/windows/Dockerfile
@@ -30,4 +30,4 @@ WORKDIR /Scripts
 COPY docker/windows/Scripts/*.ps1 /Scripts/
 
 HEALTHCHECK --interval=20s --timeout=20s --retries=6 CMD powershell -file ./healthcheck-tentacle.ps1
-CMD ["powershell.exe", "-File", "configure-and-run-tentacle-wrapper.ps1"]
+ENTRYPOINT ["powershell.exe", "-File", "configure-and-run-tentacle-wrapper.ps1"]
--- a/docker/windows/Dockerfile
+++ b/docker/windows/Dockerfile
@@ -28,5 +28,6 @@ RUN Tentacle create-instance --console --instance Tentacle --config C:\Octopus\T
 
 WORKDIR /Scripts
 COPY docker/windows/Scripts/*.ps1 /Scripts/
+
 HEALTHCHECK --interval=20s --timeout=20s --retries=6 CMD powershell -file ./healthcheck-tentacle.ps1
-ENTRYPOINT powershell -File ./configure-tentacle.ps1 -Verbose  && powershell -File ./run-tentacle.ps1 -Verbose
+CMD ["powershell.exe", "-File", "configure-and-run-tentacle-wrapper.ps1"]


docker/windows/Scripts/configure-and-run-tentacle-wrapper.ps1
```diff
@@ -0,0 +1,2 @@
+﻿& ./configure-tentacle.ps1 -Verbose
```
geofflamrock, line 1: Am I right in remembering that the `&` here makes the script run in the background?

LukeButters: in powershell it seems `&` at the start means "call a process" and `&` at the end means run in the background. How confusing!
https://ss64.com/ps/call.html
https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_operators?view=powershell-7.3&viewFallbackFrom=powershell-7.1#call-operator-
https://stackoverflow.com/questions/22074507/what-does-the-symbol-in-powershell-mean#:~:text=%26%20is%20the%20call%20operator%20which,a%20script%2C%20or%20a%20function.

geofflamrock: Ah right, yeah that could take a bit to get used to. My potential concern was the two scripts not running in sequence anymore, but seems fine if the `&` at the start just means "run this process and wait please".

geofflamrock (approved):
LGTM, a couple of things you might consider doing based on when we chatted:
- The PR title has `CMD` in it which was the original approach we were looking to take, it might need updating now
- We identified that having `ENTRYPOINT` in exec form changes the behaviour if someone were to run this container with a specific command. Previously this would be ignored and now it will be appended to the command. Because we are calling a script now I'm not sure whether this will fail or not. It feels like it could be worth noting this in the PR description in case anyone comes across it.

Another thought: What does the `ENTRYPOINT` for the linux container look like? Is it worth ensuring these are consistent?

docker/windows/Scripts/configure-and-run-tentacle-wrapper.ps1
LukeButters: The tentacle tests do run the windows tentacle docker container, so if it is completely broken the test chain should catch it.

michaelongso:
> The PR title has CMD
Updated.

> We identified that having ENTRYPOINT in exec form changes the behaviour

Added _ Behavioural Update from shell to exec_ section in the PR description.

> Another thought: What does the ENTRYPOINT for the linux container look like
The `Dockerfile` for the Linux container is using `CMD`. There's little commonality due to the requirement for `.sh` and `.ps1`, so it is not worthwhile to attempt consistency.

PR #526, opened 2023-06-20
Use Process.Kill recursive API when cancelling scripts in .NET 6

geofflamrock:
# Background

There's been a recent increase of intermittently failing integration tests for cancellation of running scripts on Linux. 

This has been tracked down to the way that we currently do process cleanup in Linux:
- Run a bash script `kill -TERM {pid}` and check this succeeded
- Check if the process has exited (`Process.HasExited`) and close the process to ensure that it releases any underlying resources (`Process.Close`). 
- Wait for the process to exit: `Process.WaitForExit()`.

It appears that sometimes `Process.HasExited` returns false and we don't close the underlying process resource, and this causes `WaitForExit` to block until the process finishes. Whether the process never really receives the call to exit properly, or there is some strange behaviour with child processes and output streams as outlined in this [runtime issue](https://github.com/dotnet/runtime/issues/29232) is a bit unclear. The behaviour we see is that the process runs until completion and result in our tests failing, but it's probable that cancellation just doesn't work sometimes in the wild.

# Results

This PR changes the way that processes are stopped by using the new overload `Process.Kill(recursive: true)` that exists in .NET 3.0 onwards. 

The .NET Framework doesn't have this new API so we continue to use the existing Windows specific method where we iterate the child processes through management object searching and clean them all up.

Shortcut [sc-48624]

Fixes https://github.com/OctopusDeploy/OctopusTentacle/issues/528

# How to review this PR

Quality :heavy_check_mark:

I confirmed the improved behaviour here by modifying the intermittently failing test to run in a loop 100 times, which surfaced the issue quite regularly, see [this test run](https://build.octopushq.com/buildConfiguration/TeamFireAndMotion_OctopusTentacleNet48_SensibleDefaultsChainFullChain/8445673?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&expandPull+Request+Details=true&expandBuildProblemsSection=true&expandBuildTestsSection=true) that also had some additional logging showing what was happening.

image

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Tests/Integration/ScriptServiceV2Fixture.cs
+++ b/source/Octopus.Tentacle.Tests/Integration/ScriptServiceV2Fixture.cs
@@ -507,7 +507,7 @@ namespace Octopus.Tentacle.Tests.Integration
         {
             foreach (var log in logs)
             {
-                TestContext.Out.WriteLine($"{0:yyyy-MM-dd HH:mm:ss}: {1}", log.Occurred, log.Text);
+                TestContext.Out.WriteLine("{0:yyyy-MM-dd HH:mm:ss K}: {1}", log.Occurred.ToLocalTime(), log.Text);
             }
         }
 
--- a/source/Octopus.Tentacle.Tests/Integration/ScriptServiceV2Fixture.cs
+++ b/source/Octopus.Tentacle.Tests/Integration/ScriptServiceV2Fixture.cs
@@ -507,7 +507,7 @@ namespace Octopus.Tentacle.Tests.Integration
         {
             foreach (var log in logs)
             {
-                TestContext.Out.WriteLine(log.Text);
+                TestContext.Out.WriteLine($"{0:yyyy-MM-dd HH:mm:ss}: {1}", log.Occurred, log.Text);
             }
         }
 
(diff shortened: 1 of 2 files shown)


source/Octopus.Tentacle.Tests/Integration/ScriptServiceV2Fixture.cs
```diff
@@ -507,7 +507,7 @@ void WriteLogsToConsole(List<ProcessOutput> logs)
         {
             foreach (var log in logs)
             {
-                TestContext.Out.WriteLine(log.Text);
+                TestContext.Out.WriteLine("{0:yyyy-MM-dd HH:mm:ss K}: {1}", log.Occurred.ToLocalTime(), log.Text);
```
LukeButters, line 510: I am surprised that this (the old code) is not using the Logger from the base class. The logger is created by `SerilogLoggerBuilder` which can just be `new`ed up at any time if required.

geofflamrock: 🤷 yeah I'm not sure, I'm not overly familiar with what is available in the tests here, I'll take a look though, thanks.

LukeButters: If you end up making it go to the seri log Logger, that change can be merged in as it would be helpful.

LukeButters: To close the loop this is not integration tests and so can not use the serilog builder I mentioned above.

LukeButters (approved):
LGTM

PR #529, opened 2023-06-22
Adding RPC Call observability to tentacle client

sburmanoctopus:
[sc-46303]

# Background

We need to know information about tentacle operations and RPC calls in Octopus Server for monitoring and metrics regarding retries.

This is needed to address some of the requirements in [this document](https://whimsical.com/retry-rpc-call-monitoring-YS9YTjocXz79jJqVBS18g2)

# Results

## Before
We had no way of knowing how often an RPC call took, or how many times it was retried etc.

## After
Now we have a way to hook into the RPC calls get gather metrics on them.

We now use an executor that wraps the `RpcCallRetryHandler`, and gathers metrics about the calls (via a builder).

### Other Fixes
`Octopus.Tentacle.Client.csproj` was explicitly including `ClientServices` with no comments or git history explaining why it was there. It was removed.

`ScriptBuilder` did not wrap paths in single quotes (causing failures on machines with spaces in their paths). Single quotes added.

`CommonTestUtils.csproj` was using C# 8 instead of what everything else was using. Updated to C# 9. 

# How to review this PR
I added the logic for determining the `AdditionalTimeFromRetries` here. This is a question Octopus Server will ask. But it felt nicer to answer the question closer to the data. Otherwise, we can lift and shift this into server itself later.

Quality :heavy_check_mark:

# Pre-requisites


source/Octopus.Tentacle.Client/Scripts/ScriptExecutionOrchestrator.cs
```diff
@@ -68,19 +73,21 @@ public async Task<ScriptStatusResponseV2> ExecuteScript(CancellationToken script
             {
                 try
                 {
-                    scriptStatusResponse = await rpcCallRetryHandler.ExecuteWithRetries(
+                    scriptStatusResponse = await rpcCallExecutor.ExecuteWithRetries(
+                        nameof(scriptServiceV2.StartScript),
```
nathanwoctopusdeploy, line 77: What does the rpcCallName end up using nameof here? Does it include the service and operation name?

sburmanoctopus: No, it will just be `StartScript`.

source/Octopus.Tentacle.Client/TentacleClient.cs
```diff
@@ -50,41 +57,76 @@ public class TentacleClient : ITentacleClient
                 capabilitiesServiceV2 = tentacleServicesDecorator.Decorate(capabilitiesServiceV2);
             }
 
-            rpcCallRetryHandler = new RpcCallRetryHandler(retryDuration, TimeoutStrategy.Pessimistic);
+            var rpcCallRetryHandler = new RpcCallRetryHandler(retryDuration, TimeoutStrategy.Pessimistic);
```
nathanwoctopusdeploy, line 60: Was the intention to use the RpcCallExecutorFactory here?

source/Octopus.Tentacle.Client/Execution/RpcCallExecutorFactory.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using Octopus.Tentacle.Client.Retries;
+using Octopus.Tentacle.Contracts.Observability;
+using Polly.Timeout;
+
+namespace Octopus.Tentacle.Client.Execution
+{
+    internal static class RpcCallExecutorFactory
```
nathanwoctopusdeploy, line 8: It doesn't look like this is used in Tentacle code, just test code. Was that intentional?

source/Octopus.Tentacle.Client/TentacleClient.cs
sburmanoctopus: Lol, yes, thank you 😁

source/Octopus.Tentacle.Client/Execution/RpcCallExecutorFactory.cs
sburmanoctopus: No, I just forgot to use it (highlighted in your previous comment 😁)

source/Octopus.Tentacle.Tests/Client/RpcCallExecutorFixture.cs
```diff
@@ -0,0 +1,405 @@
...
+
+            attempt.End.Should().BeAfter(attempt.Start);
+            attempt.Duration.Should().Be(attempt.End - attempt.Start);
+        }
+
+        private static void ThenAttemptShouldHaveFailed(TimedOperation attempt, Exception expectedException)
+        {
+            attempt.Succeeded.Should().BeFalse();
+            attempt.Exception.Should().BeEquivalentTo(expectedException);
+            attempt.WasCancelled.Should().BeFalse();
```
nathanwoctopusdeploy, line 391: Doesn't look like cancellation and WasCancelled is ever tested

sburmanoctopus: Is it not tested in `ExecuteWithRetries_MetricsShouldBeFailed_WhenCancellationTokenIsCancelled`? 

Or am I misunderstanding something?

nathanwoctopusdeploy: No, I missed it ;) Nothing to see here!!

source/Octopus.Tentacle.Client/Scripts/ScriptExecutionOrchestrator.cs
nathanwoctopusdeploy: Probably not needed now but if we log RPC Metrics in Server it may be necessary to know which service the operation is on e.g. SSv1 and SSv2 both have a StartScript operation

nathanwoctopusdeploy (approved):
LGTM. Lovely job!

source/Octopus.Tentacle.Client/Scripts/ScriptExecutionOrchestrator.cs
sburmanoctopus: Agreed.

I might keep it simple and leave it as is for now until we need it.

nathanwoctopusdeploy:

PR #531, opened 2023-06-27
Added caching for ICapabilitiesServiceV2.GetCapabilities RPC call

nathanwoctopusdeploy:
# Background

This PR adds caching using the Halibut Response Caching feature https://github.com/OctopusDeploy/Halibut/pull/251 for the `ICapabilitiesService .GetCapabilities` RPC call

It also configures error response caching so that a Service Not Found error response is cached. This caters for Tentacle before the ICapabilitiesService was added, which returns this response and ensures that caching is effective for older versions of Tentacle.

# Results

## Before

Every call to ICapabilitiesServiceV2.GetCapabilities would make an RPC call to Tentacle.

## After

The first call to ICapabilitiesServiceV2.GetCapabilities will make an RPC call to Tentacle. Subsequent calls within 10 minutes will use a cached response. After 10 minutes, the cached response is expired and an RPC call will be made to Tentacle

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Octopus.Tentacle.Client/TentacleClient.cs
```diff
@@ -87,6 +98,19 @@ public async Task<UploadResult> UploadFile(string fileName, string path, DataStr
             return (DataStream?)dataStream;
         }
 
+        internal async Task<CapabilitiesResponseV2> GetCapabilities(ILog logger, CancellationToken cancellationToken)
```
nathanwoctopusdeploy, line 101: This is purely needed to be able to test the caching in an integration test effectively.

Internal so calls clients can't call it directly

source/Octopus.Tentacle.Contracts/Octopus.Tentacle.Contracts.csproj
```diff
@@ -28,7 +28,7 @@
     </Otherwise>
   </Choose>
   <ItemGroup>
-    <PackageReference Include="Halibut" Version="6.0.128" />
+    <PackageReference Include="Halibut" Version="6.0.179-pull-251" />
```
nathanwoctopusdeploy, line 31: TODO - BUMP TO A RELEASE VERSION

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionCanBeCancelled.cs
```diff
@@ -65,14 +67,14 @@ public async Task DuringGetCapabilities_ScriptExecutionCanBeCancelled(TentacleTy
                             {
                                 if (rpcCall == RpcCall.RetryingCall && capabilityServiceV2Exceptions.GetCapabilitiesLatestException == null)
                                 {
-                                    service.EnsureTentacleIsConnectedToServer(Logger);
```
nathanwoctopusdeploy, line 68: As responses are cached, we can't make a call to the ICapabilitesService to ensure tentacle is connected but need to use another service.

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionCanRecoverFromNetworkIssues.cs
```diff
@@ -239,14 +243,16 @@ public async Task WhenANetworkFailureOccurs_DuringGetCapabilities_TheClientIsAbl
                         {
                             if (capabilitiesServiceV2Exceptions.GetCapabilitiesLatestException == null)
                             {
-                                inner.EnsureTentacleIsConnectedToServer(Logger);
```
nathanwoctopusdeploy, line 242: As responses are cached, we can't make a call to the ICapabilitesService to ensure tentacle is connected but need to use another service.

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionRetriesTimeout.cs
```diff
@@ -43,9 +47,9 @@ public async Task WhenRpcRetriesTimeOut_DuringGetCapabilities_TheRpcCallIsCancel
                     .DecorateCapabilitiesServiceV2With(d => d
                         .BeforeGetCapabilities((service) =>
                         {
-                            service.EnsureTentacleIsConnectedToServer(Logger);
```
nathanwoctopusdeploy, line 46: As responses are cached, we can't make a call to the ICapabilitesService to ensure tentacle is connected but need to use another service.

source/Octopus.Tentacle.Tests.Integration/Util/TcpTentacleHelpers/ResponseMessageTcpKillerWorkarounds.cs
```diff
@@ -17,13 +17,6 @@ public static void EnsureTentacleIsConnectedToServer(this IClientScriptServiceV2
             logger.ForContext(typeof(ResponseMessageTcpKillerWorkarounds)).Information("Finished GetStatus work around call");
         }
 
-        public static void EnsureTentacleIsConnectedToServer(this IClientCapabilitiesServiceV2 service, ILogger logger)
```
nathanwoctopusdeploy, line 20: This can no longer be used to ensure tentacle is connected as responses are cached

source/Octopus.Tentacle.Tests/Capabilities/BackwardsCompatibleCapabilitiesV2DecoratorFixture.cs
```diff
@@ -0,0 +1,33 @@
+using FluentAssertions;
+using NUnit.Framework;
+using Octopus.Tentacle.Contracts.Capabilities;
+
+namespace Octopus.Tentacle.Tests.Capabilities
+{
+    public class BackwardsCompatibleCapabilitiesV2DecoratorFixture
```
nathanwoctopusdeploy, line 7: This was a rename

source/Octopus.Tentacle.Tests/Capabilities/BackwardsCompatibleCapabilitiesV2TestServices.cs
```diff
@@ -0,0 +1,34 @@
+﻿using Halibut;
+using Halibut.Exceptions;
+using Octopus.Tentacle.Contracts.Capabilities;
+
+namespace Octopus.Tentacle.Tests.Capabilities
+{
+    internal class BackwardsCompatibleCapabilitiesV2TestServices
+    {
+        internal const string OldTentacleMissingCapabilitiesServiceMessage = @"Error: ICapabilitiesServiceV2 has not been registered through an IAutofacServiceSource";
```
nathanwoctopusdeploy, line 9: Pulled out of an existing test for resuse

source/Octopus.Tentacle/Octopus.Tentacle.csproj
```diff
@@ -95,7 +95,7 @@
 		<PackageReference Include="System.Diagnostics.EventLog" Version="4.5.0" />
 		<PackageReference Include="System.Text.Encoding.CodePages" Version="6.0.0" />
 		<PackageReference Include="System.DirectoryServices.AccountManagement" Version="4.5.0" />
-		<PackageReference Include="System.Security.Cryptography.ProtectedData" Version="4.5.0" />
+		<PackageReference Include="System.Security.Cryptography.ProtectedData" Version="5.0.0" />
```
nathanwoctopusdeploy, line 98: This comes from a bump of some nuget packages in Halibut to add the Memory Cache

source/Octopus.Tentacle/Configuration/WindowsMachineKeyEncryptor.cs
```diff
@@ -7,10 +7,12 @@ namespace Octopus.Tentacle.Configuration
 {
     public class WindowsMachineKeyEncryptor : IMachineKeyEncryptor
     {
+#pragma warning disable CA1416
```
nathanwoctopusdeploy, line 10: This comes from a bump of some nuget packages in Halibut to add the Memory Cache

source/Octopus.Tentacle.Contracts/Capabilities/BackwardsCompatibleCapabilitiesV2Helper.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        public static bool ExceptionTypeLooksLikeTheServiceWasNotFound(string typeFullName)
+        {
+            return typeFullName == NoMatchingServiceOrMethodHalibutClientExceptionTypeFullName ||
+                typeFullName == MethodNotFoundHalibutClientExceptionTypeFullName ||
+                typeFullName == ServiceNotFoundHalibutClientExceptionTypeFullName ||
+                typeFullName == AmbiguousMethodMatchHalibutClientExceptionTypeFullName;
+        }
+
+        public static bool ExceptionMessageLooksLikeTheServiceWasNotFound(string exceptionMessage)
```
nathanwoctopusdeploy, line 28: This logic was refactored from the decorator for reuse

source/Octopus.Tentacle.Client/Octopus.Tentacle.Client.csproj
```diff
@@ -35,7 +35,4 @@
     <ItemGroup>
       <ProjectReference Include="..\Octopus.Tentacle.Contracts\Octopus.Tentacle.Contracts.csproj" />
     </ItemGroup>
-    <ItemGroup>
```
nathanwoctopusdeploy, line 38: Unsure what this was doing here!!

nathanwoctopusdeploy:
[sc-46261]

source/Octopus.Tentacle.Client/TentacleClient.cs
```diff
@@ -33,6 +33,17 @@ public class TentacleClient : ITentacleClient
         {
             this.scriptObserverBackOffStrategy = scriptObserverBackOffStrategy;
 
+            halibutRuntime.OverrideErrorResponseMessageCaching += response =>
```
LukeButters, line 36: What does `+=` do here?

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionCanBeCancelled.cs
sburmanoctopus: Not a big deal, but is there a way to make this more obvious so the next person doesn't have the bright idea of "using `service` so we don't need `scriptServiceV2`"?

source/Octopus.Tentacle/Configuration/WindowsMachineKeyEncryptor.cs
sburmanoctopus, line 10: Just wondering, why did this become something we need to disable?

source/Octopus.Tentacle.Client/TentacleClient.cs
sburmanoctopus: It would be nice to test this without adding a method that isn't used outside tests. 

Are there any ways we can achieve this?

If not, then that's fine. I thought I'd ask 🙂

sburmanoctopus (approved):
LGTM. A couple of small comments/questions, but nothing to block a merge when ready 🙂

source/Octopus.Tentacle/Configuration/WindowsMachineKeyEncryptor.cs
nathanwoctopusdeploy: It's due to a nuget bump, they have decorated methods that are OS-specific more widely and result in this warning. We only call this code when it's Windows so it's not of any concern (caller checks the OS first)

It could be cleaned up/made safer so it doesn't have this code path when we compile for anything other than Windows but didn't want to do that in this PR.

source/Octopus.Tentacle.Client/TentacleClient.cs
nathanwoctopusdeploy: Chains delegate handlers but the result from the last handler wins

image

source/Octopus.Tentacle.Client/TentacleClient.cs
sburmanoctopus: Could we call the service without going via the client? Or exercise the path when executing the script? I'm not sure how feasible these ideas are though 😞

source/Octopus.Tentacle.Client/TentacleClient.cs
LukeButters: Why did you opt to chain it over just replacing what it has?

nathanwoctopusdeploy: It doesn't make that much sense in this scenario. Instead, it makes more sense to actually chain to allow the caller to hook up their own error handling if needed e.g. Server wanted to do something additional to error caching.

```
var innerHandler = halibutRuntime.OverrideErrorResponseMessageCaching;
            halibutRuntime.OverrideErrorResponseMessageCaching = response =>
            {
                if (BackwardsCompatibleCapabilitiesV2Helper.ExceptionTypeLooksLikeTheServiceWasNotFound(response.Error.HalibutErrorType) ||
                    BackwardsCompatibleCapabilitiesV2Helper.ExceptionMessageLooksLikeTheServiceWasNotFound(response.Error.Message))
                {
                    return true;
                }

                // TentacleClient doesn't own the HalibutRuntime so allow other handlers to be configured that override the error caching behaviour
                return innerHandler?.Invoke(response) ?? false;
            };
```

source/Octopus.Tentacle.Client/TentacleClient.cs
nathanwoctopusdeploy: We probably can with decorators!! I'll try and remove the need for this!!

nathanwoctopusdeploy: I've removed the need for this by refactoring the test to call ExecuteScript and using decorators and a bit of port forwarder orchestration to ensure cached responses are returned.

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionCanBeCancelled.cs
nathanwoctopusdeploy: I have no good ideas for this so leaving as is for now.

source/Octopus.Tentacle.Client/TentacleClient.cs
nathanwoctopusdeploy: I've changed this to make more sense

source/Octopus.Tentacle.Tests.Integration/CapabilitiesServiceV2Test.cs
```diff
@@ -51,5 +59,57 @@ public class CapabilitiesServiceV2Test : IntegrationTest
...
+            using var clientAndTentacle = await new ClientAndTentacleBuilder(tentacleType)
+                .WithPortForwarder(out var portForwarder)
+                .WithTentacleVersion(version)
+                .WithTentacleServiceDecorator(new TentacleServiceDecoratorBuilder()
+                    .DecorateCapabilitiesServiceV2With(d => d
+                        .AfterGetCapabilities((response) =>
+                        {
+                            capabilitiesResponses.Add(response);
+
+                            if (resumePortForwarder)
```
LukeButters, line 79: Oh this port forwarder is needed since the decorators wrap halibut and it is halibut doing the caching so unless we do this we wont know if it is coming from the cache or by a call to tentacle.

I guess this is where some sort of tentacle proxy would be super useful.

nathanwoctopusdeploy:

PR #533, opened 2023-06-30
Adding which service was called to RPC metrics

sburmanoctopus:
[sc-52746]

# Background

We need to know if our new feature is being used. To populate the respective metrics in Octopus Server, we will achieve this by knowing which service was called.

# Results

Related to https://github.com/OctopusDeploy/Issues/issues/8229

## Before

Only the method name was recorded in the metrics.

## After
The service name is added as well.

We went with the name of the interface instead of the object, as the object tends to be a decorator and other unhelpful classes.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


nathanwoctopusdeploy (approved):
LGTM

PR #534, opened 2023-06-30
Allow RPC retries to be disabled

acodrington:
RPC retries are currently always enabled when using ScriptService v2. If the user does not have RPC retries enabled on the machine policy, then SSv1 must be used.

It would be preferable for SSv2 to be used instead, as even though SSv1 will be maintained in the future, SSv2 will be the main focus for receiving updates and new features.

This PR adds the ability to disable RPC retries when instantiating TentacleClient, allowing SSv2 to be used without needing to fallback to SSv1.

Implemented as part of [sc-51287].

Notes:
* `Execute` has been renamed to `ExecuteWithNoRetries` in `RpcCallExecutor`, for more consistent naming to match `ExecuteWithRetries`.
* `ExecuteWithNoRetries`/`Execute` methods have been made async.
* 'Abandon action on cancellation' functionality has been added when retries are disabled.

### DIFF
--- a/source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
+++ b/source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
@@ -141,26 +141,29 @@ namespace Octopus.Tentacle.Client.Execution
             await rpcCallNoRetriesHandler.ExecuteWithNoRetries(
                 async ct =>
                 {
-                    var rpcCallMetricsBuilder = RpcCallMetricsBuilder.StartWithoutRetries(rpcCall);
-                    var start = DateTimeOffset.UtcNow;
+                    await Task.Run(() =>
+                    {
+                        var rpcCallMetricsBuilder = RpcCallMetricsBuilder.StartWithoutRetries(rpcCall);
+                        var start = DateTimeOffset.UtcNow;
 
-                    try
-                    {
-                        await Task.Run(() => action(ct), ct);
-                        rpcCallMetricsBuilder.WithAttempt(TimedOperation.Success(start));
-                    }
-                    catch (Exception e)
-                    {
-                        rpcCallMetricsBuilder.WithAttempt(TimedOperation.Failure(start, e, ct));
-                        rpcCallMetricsBuilder.Failure(e, ct);
-                        throw;
-                    }
-                    finally
-                    {
-                        var rpcCallMetrics = rpcCallMetricsBuilder.Build();
-                        clientOperationMetricsBuilder.WithRpcCall(rpcCallMetrics);
-                        tentacleClientObserver.RpcCallCompleted(rpcCallMetrics);
-                    }
+                        try
+                        {
+                            action(ct);
+                            rpcCallMetricsBuilder.WithAttempt(TimedOperation.Success(start));
+                        }
+                        catch (Exception e)
+                        {
+                            rpcCallMetricsBuilder.WithAttempt(TimedOperation.Failure(start, e, ct));
+                            rpcCallMetricsBuilder.Failure(e, ct);
+                            throw;
+                        }
+                        finally
+                        {
+                            var rpcCallMetrics = rpcCallMetricsBuilder.Build();
+                            clientOperationMetricsBuilder.WithRpcCall(rpcCallMetrics);
+                            tentacleClientObserver.RpcCallCompleted(rpcCallMetrics);
+                        }
+                    }, ct);
                 },
                 abandonActionOnCancellation,
                 AbandonAfter,
--- a/source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
+++ b/source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Reflection.Emit;
 using System.Threading;
 using System.Threading.Tasks;
 using Octopus.Diagnostics;
@@ -63,7 +64,7 @@ namespace Octopus.Tentacle.Client.Execution
                         onTimeoutAction: async (timeoutDuration, _) =>
                         {
                             await Task.CompletedTask;
-                            logger.Info($"Could not communicating with Tentacle after retrying for {timeoutDuration.TotalSeconds} seconds. No more retries will be attempted.");
+                            logger.Info($"Could not communicate with Tentacle after retrying for {timeoutDuration.TotalSeconds} seconds. No more retries will be attempted.");
                         },
                         abandonActionOnCancellation,
                         abandonAfter: TimeSpan.FromSeconds(5),
@@ -90,27 +91,66 @@ namespace Octopus.Tentacle.Client.Execution
             ClientOperationMetricsBuilder clientOperationMetricsBuilder,
             CancellationToken cancellationToken)
         {
+            return Execute(rpcCall, action, abandonActionOnCancellation: false, clientOperationMetricsBuilder, cancellationToken);
+        }
+
+        public T Execute<T>(
+            RpcCall rpcCall,
+            Func<CancellationToken, T> action,
+            bool abandonActionOnCancellation,
+            ClientOperationMetricsBuilder clientOperationMetricsBuilder,
+            CancellationToken cancellationToken)
+        {
             var rpcCallMetricsBuilder = RpcCallMetricsBuilder.StartWithoutRetries(rpcCall);
             var start = DateTimeOffset.UtcNow;
+            var abandonAfter = TimeSpan.FromSeconds(5);
 
-            try
+            T ExecuteAction(Action<Exception> exceptionHandler)
             {
-                var response = action(cancellationToken);
+                try
+                {
+                    var response = action(cancellationToken);
 
-                rpcCallMetricsBuilder.WithAttempt(TimedOperation.Success(start));
-                return response;
+                    rpcCallMetricsBuilder.WithAttempt(TimedOperation.Success(start));
+                    return response;
+                }
+                catch (Exception e)
+                {
+                    rpcCallMetricsBuilder.WithAttempt(TimedOperation.Failure(start, e, cancellationToken));
+                    rpcCallMetricsBuilder.Failure(e, cancellationToken);
+
+                    exceptionHandler(e);
+
+                    throw;
+                }
+                finally
+                {
+                    var rpcCallMetrics = rpcCallMetricsBuilder.Build();
+                    clientOperationMetricsBuilder.WithRpcCall(rpcCallMetrics);
+                    tentacleClientObserver.RpcCallCompleted(rpcCallMetrics);
+                }
             }
-            catch (Exception e)
+
+            if (!abandonActionOnCancellation)
             {
-                rpcCallMetricsBuilder.WithAttempt(TimedOperation.Failure(start, e, cancellationToken));
-                rpcCallMetricsBuilder.Failure(e, cancellationToken);
-                throw;
+                return ExecuteAction(_ =>
+                {
+                });
             }
-            finally
+
+            using var abandonCancellationTokenSource = new CancellationTokenSource();
+            using (cancellationToken.Register(() =>
+                   {
+                       abandonCancellationTokenSource.TryCancelAfter(abandonAfter);
+                   }))
             {
-                var rpcCallMetrics = rpcCallMetricsBuilder.Build();
-                clientOperationMetricsBuilder.WithRpcCall(rpcCallMetrics);
-                tentacleClientObserver.RpcCallCompleted(rpcCallMetrics);
+                return ExecuteAction(e =>
+                {
+                    if (e is OperationCanceledException && abandonCancellationTokenSource.IsCancellationRequested)
+                    {
+                        throw new OperationAbandonedException(e, abandonAfter);
+                    }
+                });
             }
         }
 
@@ -120,27 +160,71 @@ namespace Octopus.Tentacle.Client.Execution
             ClientOperationMetricsBuilder clientOperationMetricsBuilder,
             CancellationToken cancellationToken)
         {
+            Execute(
+                rpcCall,
+                action,
+                abandonActionOnCancellation: false,
+                clientOperationMetricsBuilder,
+                cancellationToken);
+        }
+
+        public void Execute(
+            RpcCall rpcCall,
+            Action<CancellationToken> action,
+            bool abandonActionOnCancellation,
+            ClientOperationMetricsBuilder clientOperationMetricsBuilder,
+            CancellationToken cancellationToken)
+        {
             var rpcCallMetricsBuilder = RpcCallMetricsBuilder.StartWithoutRetries(rpcCall);
             var start = DateTimeOffset.UtcNow;
+            var abandonAfter = TimeSpan.FromSeconds(5);
 
-            try
+            void ExecuteAction(Action<Exception> exceptionHandler)
             {
-                action(cancellationToken);
+                try
+                {
+                    action(cancellationToken);
 
-                rpcCallMetricsBuilder.WithAttempt(TimedOperation.Success(start));
+                    rpcCallMetricsBuilder.WithAttempt(TimedOperation.Success(start));
+                }
+                catch (Exception e)
+                {
+                    rpcCallMetricsBuilder.WithAttempt(TimedOperation.Failure(start, e, cancellationToken));
+                    rpcCallMetricsBuilder.Failure(e, cancellationToken);
+
+                    exceptionHandler(e);
+
+                    throw;
+                }
+                finally
+                {
+                    var rpcCallMetrics = rpcCallMetricsBuilder.Build();
+                    clientOperationMetricsBuilder.WithRpcCall(rpcCallMetrics);
+                    tentacleClientObserver.RpcCallCompleted(rpcCallMetrics);
+                }
             }
-            catch (Exception e)
+
+            if (!abandonActionOnCancellation)
             {
-                rpcCallMetricsBuilder.WithAttempt(TimedOperation.Failure(start, e, cancellationToken));
-                rpcCallMetricsBuilder.Failure(e, cancellationToken);
-                throw;
+                ExecuteAction(_ => { });
+                return;
             }
-            finally
+            
+            using var abandonCancellationTokenSource = new CancellationTokenSource();
+            using (cancellationToken.Register(() =>
+                   {
+                       abandonCancellationTokenSource.TryCancelAfter(abandonAfter);
+                   }))
             {
-                var rpcCallMetrics = rpcCallMetricsBuilder.Build();
-                clientOperationMetricsBuilder.WithRpcCall(rpcCallMetrics);
-                tentacleClientObserver.RpcCallCompleted(rpcCallMetrics);
+                ExecuteAction(e =>
+                {
+                    if (e is OperationCanceledException && abandonCancellationTokenSource.IsCancellationRequested)
+                    {
+                        throw new OperationAbandonedException(e, abandonAfter);
+                    }
+                });
             }
+            
         }
     }
 }
--- a/source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
+++ b/source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
@@ -10,6 +10,8 @@ namespace Octopus.Tentacle.Client.Execution
 {
     internal class RpcCallExecutor
     {
+        private static readonly TimeSpan AbandonAfter = TimeSpan.FromSeconds(5);
+
         private readonly RpcCallRetryHandler rpcCallRetryHandler;
         private readonly ITentacleClientObserver tentacleClientObserver;
 
@@ -84,7 +86,26 @@ namespace Octopus.Tentacle.Client.Execution
             }
         }
 
-        public T Execute<T>(
+        public async Task<T> Execute<T>(
+            RpcCall rpcCall,
+            Func<CancellationToken, T> action,
+            bool abandonActionOnCancellation,
+            ClientOperationMetricsBuilder clientOperationMetricsBuilder,
+            CancellationToken cancellationToken)
+        {
+            var response = await rpcCallRetryHandler.ExecuteWithNoRetries(
+                async ct =>
+                {
+                    // Wrap the action in a task so it doesn't block on sync Halibut calls
+                    return await Task.Run(() => Execute(rpcCall, action, clientOperationMetricsBuilder, ct), ct);
+                },
+                abandonActionOnCancellation,
+                AbandonAfter,
+                cancellationToken);
+            return response;
+        }
+
+        private T Execute<T>(
             RpcCall rpcCall,
             Func<CancellationToken, T> action,
             ClientOperationMetricsBuilder clientOperationMetricsBuilder,
--- a/source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
+++ b/source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
@@ -10,8 +10,6 @@ namespace Octopus.Tentacle.Client.Execution
 {
     internal class RpcCallExecutor
     {
-        private static readonly TimeSpan AbandonAfter = TimeSpan.FromSeconds(5);
-        
         private readonly RpcCallRetryHandler rpcCallRetryHandler;
         private readonly ITentacleClientObserver tentacleClientObserver;
 
@@ -68,7 +66,7 @@ namespace Octopus.Tentacle.Client.Execution
                             logger.Info($"Could not communicate with Tentacle after retrying for {timeoutDuration.TotalSeconds} seconds. No more retries will be attempted.");
                         },
                         abandonActionOnCancellation,
-                        abandonAfter: AbandonAfter,
+                        abandonAfter: TimeSpan.FromSeconds(5),
                         cancellationToken)
                     .ConfigureAwait(false);
                 return response;
@@ -89,113 +87,59 @@ namespace Octopus.Tentacle.Client.Execution
         public T Execute<T>(
             RpcCall rpcCall,
             Func<CancellationToken, T> action,
-            bool abandonActionOnCancellation,
             ClientOperationMetricsBuilder clientOperationMetricsBuilder,
             CancellationToken cancellationToken)
         {
             var rpcCallMetricsBuilder = RpcCallMetricsBuilder.StartWithoutRetries(rpcCall);
             var start = DateTimeOffset.UtcNow;
 
-            T ExecuteAction(Action<Exception>? exceptionHandler = null)
+            try
             {
-                try
-                {
-                    var response = action(cancellationToken);
+                var response = action(cancellationToken);
 
-                    rpcCallMetricsBuilder.WithAttempt(TimedOperation.Success(start));
-                    return response;
-                }
-                catch (Exception e)
-                {
-                    rpcCallMetricsBuilder.WithAttempt(TimedOperation.Failure(start, e, cancellationToken));
-                    rpcCallMetricsBuilder.Failure(e, cancellationToken);
-
-                    exceptionHandler?.Invoke(e);
-
-                    throw;
-                }
-                finally
-                {
-                    var rpcCallMetrics = rpcCallMetricsBuilder.Build();
-                    clientOperationMetricsBuilder.WithRpcCall(rpcCallMetrics);
-                    tentacleClientObserver.RpcCallCompleted(rpcCallMetrics);
-                }
+                rpcCallMetricsBuilder.WithAttempt(TimedOperation.Success(start));
+                return response;
             }
-
-            if (!abandonActionOnCancellation)
+            catch (Exception e)
             {
-                return ExecuteAction();
+                rpcCallMetricsBuilder.WithAttempt(TimedOperation.Failure(start, e, cancellationToken));
+                rpcCallMetricsBuilder.Failure(e, cancellationToken);
+                throw;
             }
-
-            using var abandonCancellationTokenSource = new CancellationTokenSource();
-            using (cancellationToken.Register(() =>
-                   {
-                       abandonCancellationTokenSource.TryCancelAfter(AbandonAfter);
-                   }))
+            finally
             {
-                return ExecuteAction(e =>
-                {
-                    if (e is OperationCanceledException && abandonCancellationTokenSource.IsCancellationRequested)
-                    {
-                        throw new OperationAbandonedException(e, AbandonAfter);
-                    }
-                });
+                var rpcCallMetrics = rpcCallMetricsBuilder.Build();
+                clientOperationMetricsBuilder.WithRpcCall(rpcCallMetrics);
+                tentacleClientObserver.RpcCallCompleted(rpcCallMetrics);
             }
         }
+
         public void Execute(
             RpcCall rpcCall,
             Action<CancellationToken> action,
-            bool abandonActionOnCancellation,
             ClientOperationMetricsBuilder clientOperationMetricsBuilder,
             CancellationToken cancellationToken)
         {
             var rpcCallMetricsBuilder = RpcCallMetricsBuilder.StartWithoutRetries(rpcCall);
             var start = DateTimeOffset.UtcNow;
 
-            void ExecuteAction(Action<Exception>? exceptionHandler = null)
+            try
             {
-                try
-                {
-                    action(cancellationToken);
+                action(cancellationToken);
 
-                    rpcCallMetricsBuilder.WithAttempt(TimedOperation.Success(start));
-                }
-                catch (Exception e)
-                {
-                    rpcCallMetricsBuilder.WithAttempt(TimedOperation.Failure(start, e, cancellationToken));
-                    rpcCallMetricsBuilder.Failure(e, cancellationToken);
-
-                    exceptionHandler?.Invoke(e);
-
-                    throw;
-                }
-                finally
-                {
-                    var rpcCallMetrics = rpcCallMetricsBuilder.Build();
-                    clientOperationMetricsBuilder.WithRpcCall(rpcCallMetrics);
-                    tentacleClientObserver.RpcCallCompleted(rpcCallMetrics);
-                }
+                rpcCallMetricsBuilder.WithAttempt(TimedOperation.Success(start));
             }
-
-            if (!abandonActionOnCancellation)
+            catch (Exception e)
             {
-                ExecuteAction();
-                return;
+                rpcCallMetricsBuilder.WithAttempt(TimedOperation.Failure(start, e, cancellationToken));
+                rpcCallMetricsBuilder.Failure(e, cancellationToken);
+                throw;
             }
-            
-            using var abandonCancellationTokenSource = new CancellationTokenSource();
-            using (cancellationToken.Register(() =>
-                   {
-                       abandonCancellationTokenSource.TryCancelAfter(AbandonAfter);
-                   }))
+            finally
             {
-                ExecuteAction(e =>
-                {
-                    if (e is OperationCanceledException && abandonCancellationTokenSource.IsCancellationRequested)
-                    {
-                        throw new OperationAbandonedException(e, AbandonAfter);
-                    }
-                });
+                var rpcCallMetrics = rpcCallMetricsBuilder.Build();
+                clientOperationMetricsBuilder.WithRpcCall(rpcCallMetrics);
+                tentacleClientObserver.RpcCallCompleted(rpcCallMetrics);
             }
         }
     }
--- a/source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
+++ b/source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
@@ -13,11 +13,16 @@ namespace Octopus.Tentacle.Client.Execution
         private static readonly TimeSpan AbandonAfter = TimeSpan.FromSeconds(5);
 
         private readonly RpcCallRetryHandler rpcCallRetryHandler;
+        private readonly RpcCallNoRetriesHandler rpcCallNoRetriesHandler;
         private readonly ITentacleClientObserver tentacleClientObserver;
 
-        internal RpcCallExecutor(RpcCallRetryHandler rpcCallRetryHandler, ITentacleClientObserver tentacleClientObserver)
+        internal RpcCallExecutor(
+            RpcCallRetryHandler rpcCallRetryHandler,
+            RpcCallNoRetriesHandler rpcCallNoRetriesHandler,
+            ITentacleClientObserver tentacleClientObserver)
         {
             this.rpcCallRetryHandler = rpcCallRetryHandler;
+            this.rpcCallNoRetriesHandler = rpcCallNoRetriesHandler;
             this.tentacleClientObserver = tentacleClientObserver;
         }
 
@@ -93,46 +98,38 @@ namespace Octopus.Tentacle.Client.Execution
             ClientOperationMetricsBuilder clientOperationMetricsBuilder,
             CancellationToken cancellationToken)
         {
-            var response = await rpcCallRetryHandler.ExecuteWithNoRetries(
+            return await rpcCallNoRetriesHandler.ExecuteWithNoRetries(
                 async ct =>
                 {
                     // Wrap the action in a task so it doesn't block on sync Halibut calls
-                    return await Task.Run(() => Execute(rpcCall, action, clientOperationMetricsBuilder, ct), ct);
+                    return await Task.Run(() =>
+                    {
+                        var rpcCallMetricsBuilder = RpcCallMetricsBuilder.StartWithoutRetries(rpcCall);
+                        var start = DateTimeOffset.UtcNow;
+
+                        try
+                        {
+                            var response = action(ct);
+                            rpcCallMetricsBuilder.WithAttempt(TimedOperation.Success(start));
+                            return response;
+                        }
+                        catch (Exception e)
+                        {
+                            rpcCallMetricsBuilder.WithAttempt(TimedOperation.Failure(start, e, ct));
+                            rpcCallMetricsBuilder.Failure(e, ct);
+                            throw;
+                        }
+                        finally
+                        {
+                            var rpcCallMetrics = rpcCallMetricsBuilder.Build();
+                            clientOperationMetricsBuilder.WithRpcCall(rpcCallMetrics);
+                            tentacleClientObserver.RpcCallCompleted(rpcCallMetrics);
+                        }
+                    }, ct);
                 },
                 abandonActionOnCancellation,
                 AbandonAfter,
                 cancellationToken);
-            return response;
-        }
-
-        private T Execute<T>(
-            RpcCall rpcCall,
-            Func<CancellationToken, T> action,
-            ClientOperationMetricsBuilder clientOperationMetricsBuilder,
-            CancellationToken cancellationToken)
-        {
-            var rpcCallMetricsBuilder = RpcCallMetricsBuilder.StartWithoutRetries(rpcCall);
-            var start = DateTimeOffset.UtcNow;
-
-            try
-            {
-                var response = action(cancellationToken);
-
-                rpcCallMetricsBuilder.WithAttempt(TimedOperation.Success(start));
-                return response;
-            }
-            catch (Exception e)
-            {
-                rpcCallMetricsBuilder.WithAttempt(TimedOperation.Failure(start, e, cancellationToken));
-                rpcCallMetricsBuilder.Failure(e, cancellationToken);
-                throw;
-            }
-            finally
-            {
-                var rpcCallMetrics = rpcCallMetricsBuilder.Build();
-                clientOperationMetricsBuilder.WithRpcCall(rpcCallMetrics);
-                tentacleClientObserver.RpcCallCompleted(rpcCallMetrics);
-            }
         }
 
         public void Execute(
--- a/source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
+++ b/source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
@@ -90,7 +90,7 @@ namespace Octopus.Tentacle.Client.Execution
             }
         }
 
-        public async Task<T> Execute<T>(
+        public async Task<T> ExecuteWithNoRetries<T>(
             RpcCall rpcCall,
             Func<CancellationToken, T> action,
             bool abandonActionOnCancellation,
@@ -131,7 +131,7 @@ namespace Octopus.Tentacle.Client.Execution
                 cancellationToken);
         }
 
-        public async Task Execute(
+        public async Task ExecuteWithNoRetries(
             RpcCall rpcCall,
             Action<CancellationToken> action,
             bool abandonActionOnCancellation,
--- a/source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
+++ b/source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
@@ -11,6 +11,7 @@ namespace Octopus.Tentacle.Client.Execution
     internal class RpcCallExecutor
     {
         private static readonly TimeSpan AbandonAfter = TimeSpan.FromSeconds(5);
+        public TimeSpan RetryTimeout => rpcCallRetryHandler.RetryTimeout;
 
         private readonly RpcCallRetryHandler rpcCallRetryHandler;
         private readonly RpcCallNoRetriesHandler rpcCallNoRetriesHandler;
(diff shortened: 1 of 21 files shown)


source/Octopus.Tentacle.Client/TentacleClient.cs
```diff
@@ -28,17 +28,20 @@ public class TentacleClient : ITentacleClient
...
 
         public TentacleClient(
             ServiceEndPoint serviceEndPoint,
             IHalibutRuntime halibutRuntime,
             IScriptObserverBackoffStrategy scriptObserverBackOffStrategy,
             TimeSpan retryDuration,
             ITentacleClientObserver tentacleClientObserver,
-            ITentacleServiceDecorator? tentacleServicesDecorator)
+            ITentacleServiceDecorator? tentacleServicesDecorator,
+            TentacleClientSettings? settings = null)
```
nathanwoctopusdeploy, line 40: How about something more specific about RetrySettings as we alrewady have `TimeSpan retryDuration` that could move into the class as well. Also not nullable, force the caller pass in the settings

source/Octopus.Tentacle.Client/TentacleClient.cs
```diff
@@ -28,17 +28,20 @@ public class TentacleClient : ITentacleClient
...
         {
             this.scriptObserverBackOffStrategy = scriptObserverBackOffStrategy;
             this.tentacleClientObserver = tentacleClientObserver;
+            this.settings = settings ?? new TentacleClientSettings();
```
nathanwoctopusdeploy, line 44: Force the caller to pass this in

source/Octopus.Tentacle.Client/Scripts/ScriptExecutionOrchestrator.cs
```diff
@@ -24,6 +24,7 @@ internal class ScriptExecutionOrchestrator : IDisposable
         readonly Action<ScriptStatusResponseV2> onScriptStatusResponseReceived;
         readonly Func<CancellationToken, Task> onScriptCompleted;
         readonly TimeSpan onCancellationAbandonCompleteScriptAfter;
+        readonly RpcRetrySettings settings;
```
nathanwoctopusdeploy, line 27: Nit: make the property names give more context of what the property is e.g. retrySettings

acodrington: ah yeah that's fair. `settings` seemed OK before, as it was more generic to the TentacleClient, but a more specific name makes sense now.

source/Octopus.Tentacle.Client/TentacleClient.cs
```diff
@@ -28,17 +28,20 @@ public class TentacleClient : ITentacleClient
...
 
         public TentacleClient(
             ServiceEndPoint serviceEndPoint,
             IHalibutRuntime halibutRuntime,
             IScriptObserverBackoffStrategy scriptObserverBackOffStrategy,
             TimeSpan retryDuration,
             ITentacleClientObserver tentacleClientObserver,
-            ITentacleServiceDecorator? tentacleServicesDecorator)
+            ITentacleServiceDecorator? tentacleServicesDecorator,
+            RpcRetrySettings settings)
```
nathanwoctopusdeploy, line 40: Move `TimeSpan retryDuration` to `RpcRetrySettings settings`

source/Octopus.Tentacle.Client/Scripts/ScriptExecutionOrchestrator.cs
```diff
@@ -73,21 +76,34 @@ public async Task<ScriptStatusResponseV2> ExecuteScript(CancellationToken script
...
+                            // If we manage to cancel the start script call we can walk away
+                            // If we do abandon the start script call we have to assume the script is running so need
+                            // to call CancelScript and CompleteScript
+                            abandonActionOnCancellation: true,
+                            clientOperationMetricsBuilder,
+                            scriptExecutionCancellationToken).ConfigureAwait(false);
+                    }
+                    else
+                    {
+                        scriptStatusResponse = rpcCallExecutor.Execute(
```
nathanwoctopusdeploy, line 101: Is there a behavioural change here as were loosing `abandonActionOnCancellation: true` which isn't around retries but around how cancellation happens

source/Octopus.Tentacle.Client/Scripts/ScriptExecutionOrchestrator.cs
```diff
@@ -142,10 +158,7 @@ async Task<ScriptServiceVersion> DetermineScriptServiceVersionToUse(ILog logger,
 
             var tentacleCapabilities = await rpcCallExecutor.ExecuteWithRetries(
```
nathanwoctopusdeploy, line 159: Should this also be changed so no retries are performed

source/Octopus.Tentacle.Client/TentacleClient.cs
```diff
@@ -184,4 +212,9 @@ public void Dispose()
         {
         }
     }
-}
\ No newline at end of file
+
+    public record RpcRetrySettings
+    {
+        public bool AllowRetries = false;
```
nathanwoctopusdeploy, line 218: Nit: I prefer Enable rather than Allow e.g. `RetriesEnabled`

nathanwoctopusdeploy:
General approach looks good but I think there is a behavioural change also being introduced by losing the ability to specify `abandonActionOnCancellation`

source/Octopus.Tentacle.Client/Scripts/ScriptExecutionOrchestrator.cs
acodrington: Hmm good question. I assumed that `Execute` was doing the same thing as `ExecuteWithRetries`, which is probably mostly true, but I'll check on this behaviour.

source/Octopus.Tentacle.Tests.Integration/Support/ClientAndTentacleBuilder.cs
```diff
@@ -170,11 +178,11 @@ public async Task<ClientAndTentacle> Build(CancellationToken cancellationToken)
                 tentacleEndPoint,
                 server.ServerHalibutRuntime,
                 scriptObserverBackoffStrategy,
-                retryDuration,
                 tentacleClientObserver,
-                tentacleServiceDecorator);
+                tentacleServiceDecorator,
+                new RpcRetrySettings(retryDuration) { RetriesEnabled = this.retriesEnabled });
```
nathanwoctopusdeploy, line 183: Can we make the constructor require all arguments to avoid this slightly messy initialisation

source/Octopus.Tentacle.Tests.Integration/Support/TentacleTypesAndCommonVersionsToTest.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections;
+using Octopus.Tentacle.Contracts;
```
nathanwoctopusdeploy, line 3: What happened here to require the Contracts to be referenced?

source/Octopus.Tentacle.Tests.Integration/ClientFileTransfersAreNotRetried.cs
```diff
@@ -0,0 +1,97 @@
...
+using Halibut;
+using NUnit.Framework;
+using Octopus.Tentacle.Tests.Integration.Support;
+using Octopus.Tentacle.Tests.Integration.Util.Builders;
+using Octopus.Tentacle.Tests.Integration.Util.Builders.Decorators;
+using Octopus.Tentacle.Tests.Integration.Util.TcpTentacleHelpers;
+
+namespace Octopus.Tentacle.Tests.Integration
+{
+    public class ClientFileTransfersAreNotRetried : IntegrationTest
```
nathanwoctopusdeploy, line 14: It would be good to add why e.g. WhenRetriesAreDisabled

nathanwoctopusdeploy: And the negative to the existing tests `source/Octopus.Tentacle.Tests.Integration/ClientFileTransfersAreRetried.cs` to make it clear why there is/isn't retry behavour

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionScriptServiceV2IsNotRetried.cs
```diff
@@ -0,0 +1,231 @@
...
+using Octopus.Tentacle.Tests.Integration.Support;
+using Octopus.Tentacle.Tests.Integration.Util;
+using Octopus.Tentacle.Tests.Integration.Util.Builders;
+using Octopus.Tentacle.Tests.Integration.Util.Builders.Decorators;
+using Octopus.Tentacle.Tests.Integration.Util.TcpTentacleHelpers;
+
+namespace Octopus.Tentacle.Tests.Integration
+{
+    [IntegrationTestTimeout]
+    public class ClientScriptExecutionScriptServiceV2IsNotRetried : IntegrationTest
```
nathanwoctopusdeploy, line 21: Maybe move the remove the `WithRetriesDisabled` from the method names and move it to the class name as `WhenRetriesDisabled` to make it easier to understand what were testing

source/Octopus.Tentacle.Client/Retries/RpcRetrySettings.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+
+namespace Octopus.Tentacle.Client.Retries
+{
+    public record RpcRetrySettings(TimeSpan RetryDuration)
+    {
+        public bool RetriesEnabled = false;
```
nathanwoctopusdeploy, line 7: Nit: Make this private and required in the constructor

source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Reflection.Emit;
```
nathanwoctopusdeploy, line 2: ??

source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
```diff
@@ -89,28 +90,67 @@ internal RpcCallExecutor(RpcCallRetryHandler rpcCallRetryHandler, ITentacleClien
             Func<CancellationToken, T> action,
             ClientOperationMetricsBuilder clientOperationMetricsBuilder,
             CancellationToken cancellationToken)
+        {
+            return Execute(rpcCall, action, abandonActionOnCancellation: false, clientOperationMetricsBuilder, cancellationToken);
```
nathanwoctopusdeploy, line 94: Can we keep the interface the same as ExecuteWithRetries so the caller has to explicity decide if abandonActionOnCancellation is true or false and not get a default this overload provides

source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
```diff
@@ -89,28 +90,67 @@ internal RpcCallExecutor(RpcCallRetryHandler rpcCallRetryHandler, ITentacleClien
...
+        public T Execute<T>(
+            RpcCall rpcCall,
+            Func<CancellationToken, T> action,
+            bool abandonActionOnCancellation,
+            ClientOperationMetricsBuilder clientOperationMetricsBuilder,
+            CancellationToken cancellationToken)
         {
             var rpcCallMetricsBuilder = RpcCallMetricsBuilder.StartWithoutRetries(rpcCall);
             var start = DateTimeOffset.UtcNow;
+            var abandonAfter = TimeSpan.FromSeconds(5);
```
nathanwoctopusdeploy, line 106: Can we make this a field so both `Execute` and `ExecuteWithRetries` use the same value and don't drift

source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
```diff
@@ -89,28 +90,67 @@ internal RpcCallExecutor(RpcCallRetryHandler rpcCallRetryHandler, ITentacleClien
...
+                }
             }
-            catch (Exception e)
+
+            if (!abandonActionOnCancellation)
             {
-                rpcCallMetricsBuilder.WithAttempt(TimedOperation.Failure(start, e, cancellationToken));
-                rpcCallMetricsBuilder.Failure(e, cancellationToken);
-                throw;
+                return ExecuteAction(_ =>
```
nathanwoctopusdeploy, line 136: Nit: Make exceptionHandler nullable so we can omit `_ =>{}` which is a bit confusing to read. and then `exceptionHandler?.Invoke(e);`

source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
```diff
@@ -89,28 +90,67 @@ internal RpcCallExecutor(RpcCallRetryHandler rpcCallRetryHandler, ITentacleClien
...
+                {
+                });
             }
-            finally
+
+            using var abandonCancellationTokenSource = new CancellationTokenSource();
+            using (cancellationToken.Register(() =>
+                   {
+                       abandonCancellationTokenSource.TryCancelAfter(abandonAfter);
```
nathanwoctopusdeploy, line 144: Do we have test coverage for this?

source/Octopus.Tentacle.Tests.Integration/Support/TentacleTypesAndCommonVersionsToTest.cs
acodrington: I dunno tbh, I think Rider may have auto-added it when I made a typo (I made some changes here, then reverted). Rider also seems to auto-hide the `using` statements on my machine, so I didn't notice it.

I've updated my Rider settings for next time, and I'll remove this.

source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
acodrington: 🤦🏻 
See above i.e. why make a mistake just once when you can do it twice?

source/Octopus.Tentacle.Client/Scripts/ScriptExecutionOrchestrator.cs
```diff
@@ -141,24 +159,37 @@ async Task<ScriptServiceVersion> DetermineScriptServiceVersionToUse(ILog logger,
...
-                abandonActionOnCancellation: true,
-                clientOperationMetricsBuilder,
-                cancellationToken).ConfigureAwait(false);
+            CapabilitiesResponseV2 tentacleCapabilities;
+
+            if (rpcRetrySettings.RetriesEnabled)
+            {
+                tentacleCapabilities = await rpcCallExecutor.ExecuteWithRetries(
+                    RpcCall.Create<IClientCapabilitiesServiceV2>(nameof(IClientCapabilitiesServiceV2.GetCapabilities)),
+                    ct => capabilitiesServiceV2.GetCapabilities(new HalibutProxyRequestOptions(ct)),
```
nathanwoctopusdeploy, line 168: For other calls the action performed by ExecuteWithRetries and Execute is an action that is reused by both. Should we keep that consistent and do that here as well

source/Octopus.Tentacle.Client/Scripts/ScriptExecutionOrchestrator.cs
```diff
@@ -282,17 +327,31 @@ await Task.Delay(scriptObserverBackOffStrategy.GetBackoff(++iteration), scriptEx
...
+                        abandonActionOnCancellation: false,
+                        clientOperationMetricsBuilder,
+                        cancellationToken).ConfigureAwait(false);
+                }
+                else
+                {
+                    return rpcCallExecutor.Execute(
+                        RpcCall.Create<IClientScriptServiceV2>(nameof(IClientScriptServiceV2.CancelScript)),
+                        CancelScriptAction,
+                        abandonActionOnCancellation: true,
```
nathanwoctopusdeploy, line 351: `ExecuteWithRetries` uses `abandonActionOnCancellation: false,`. Did you mean to change this behavour to `abandonActionOnCancellation: true,` for `Execute`?

source/Octopus.Tentacle.Client/TentacleClient.cs
```diff
@@ -77,21 +80,34 @@ public async Task<UploadResult> UploadFile(string fileName, string path, DataStr
...
+                        RpcCall.Create<IClientFileTransferService>(nameof(IClientFileTransferService.UploadFile)),
+                        UploadFileAction,
+                        logger,
+                        abandonActionOnCancellation: false,
+                        operationMetricsBuilder,
+                        cancellationToken).ConfigureAwait(false);
+                }
+                else
+                {
+                    return rpcCallExecutor.Execute(
```
nathanwoctopusdeploy, line 105: Can we explicitly set `abandonActionOnCancellation: false,` to make it clear what the behaviour is

source/Octopus.Tentacle.Client/TentacleClient.cs
```diff
@@ -109,23 +125,34 @@ public async Task<UploadResult> UploadFile(string fileName, string path, DataStr
...
+                        RpcCall.Create<IClientFileTransferService>(nameof(IClientFileTransferService.DownloadFile)),
+                        DownloadFileAction,
+                        logger,
+                        abandonActionOnCancellation: false,
+                        operationMetricsBuilder,
+                        cancellationToken).ConfigureAwait(false);
+                }
+                else
+                {
+                    return rpcCallExecutor.Execute(
```
nathanwoctopusdeploy, line 150: Can we explicitly set abandonActionOnCancellation: false, to make it clear what the behaviour is

source/Octopus.Tentacle.Tests.Integration/ClientFileTransfersAreNotRetried.cs
```diff
@@ -0,0 +1,97 @@
...
+                .WithTentacleServiceDecorator(new TentacleServiceDecoratorBuilder()
+                    .LogCallsToFileTransferService()
+                    .CountCallsToFileTransferService(out var fileTransferServiceCallCounts)
+                    .RecordExceptionThrownInFileTransferService(out var fileTransferServiceException)
+                    .DecorateFileTransferServiceWith(b =>
+                    {
+                        b.BeforeUploadFile((service, _, ds) =>
+                        {
+                            service.EnsureTentacleIsConnectedToServer(Logger);
+                            // Only kill the connection the first time, causing the upload
```
nathanwoctopusdeploy, line 34: This comment doesn't make a lot of sense

source/Octopus.Tentacle.Tests.Integration/ClientFileTransfersAreNotRetried.cs
nathanwoctopusdeploy, line 14: Add `[IntegrationTestTimeout]`

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionScriptServiceV2IsNotRetried.cs
```diff
@@ -0,0 +1,231 @@
...
+                    .Print("hello")
+                    .WaitForFileToExist(waitForFile)
+                    .Print("AllDone"))
+                .Build();
+
+            List<ProcessOutput> logs = new List<ProcessOutput>();
+            Assert.ThrowsAsync<HalibutClientException>(async () => await clientTentacle.TentacleClient.ExecuteScriptAssumingException(startScriptCommand, logs, CancellationToken));
+
+            // Let the script finish.
+            File.WriteAllText(waitForFile, "");
```
nathanwoctopusdeploy, line 63: Because we are failing the first call and not retrying the script will never start so the File lock logic can be removed

nathanwoctopusdeploy: `.WaitForFileToExist(waitForFile)`

and

`File.WriteAllText(waitForFile, "");`

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionScriptServiceV2IsNotRetried.cs
```diff
@@ -0,0 +1,231 @@
...
+
+            var allLogs = logs.JoinLogs();
+
+            allLogs.Should().NotContain("AllDone");
+            scriptServiceExceptions.StartScriptLatestException.Should().NotBeNull();
+            scriptServiceCallCounts.StartScriptCallCountStarted.Should().Be(1);
+            scriptServiceCallCounts.GetStatusCallCountStarted.Should().Be(0);
+            scriptServiceCallCounts.CompleteScriptCallCountStarted.Should().Be(0);
+
+            // We must ensure all script are complete, otherwise if we shutdown tentacle while running a script the build can hang.
```
nathanwoctopusdeploy, line 73: I don't think we need this as we never start the original script

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionScriptServiceV2IsNotRetried.cs
```diff
@@ -0,0 +1,231 @@
...
+                .WithResponseMessageTcpKiller(out var responseMessageTcpKiller)
+                .WithRetryDuration(TimeSpan.FromMinutes(4))
+                .WithTentacleServiceDecorator(new TentacleServiceDecoratorBuilder()
+                    .LogCallsToScriptServiceV2()
+                    .RecordExceptionThrownInScriptServiceV2(out var scriptServiceExceptions)
+                    .CountCallsToScriptServiceV2(out var scriptServiceCallCounts)
+                    .DecorateScriptServiceV2With(new ScriptServiceV2DecoratorBuilder()
+                        .BeforeStartScript(() =>
+                        {
+                            if (scriptServiceExceptions.StartScriptLatestException == null)
```
nathanwoctopusdeploy, line 39: Should this ensure Tentacle is connected to make sure we don't hit another Halibut 'feature' e.g. dequeue into a broken TCP connection for polling tentacle

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionScriptServiceV2IsNotRetried.cs
```diff
@@ -0,0 +1,231 @@
...
+            var startScriptCommand = new StartScriptCommandV2Builder()
+                .WithScriptBody(new ScriptBuilder()
+                    .Print("hello")
+                    .WaitForFileToExist(waitForFile)
+                    .Print("AllDone"))
+                .Build();
+
+            List<ProcessOutput> logs = new List<ProcessOutput>();
+
+            //Assert.CatchAsync(async () => await clientTentacle.TentacleClient.ExecuteScriptAssumingException(startScriptCommand, logs, cts.Token));
```
nathanwoctopusdeploy, line 176: ?

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionScriptServiceV2IsNotRetried.cs
```diff
@@ -0,0 +1,231 @@
...
+                    .Build())
+                .Build(CancellationToken);
+
+            var startScriptCommand = new StartScriptCommandV2Builder().WithScriptBody(new ScriptBuilder().Print("hello")).Build();
+
+            List<ProcessOutput> logs = new List<ProcessOutput>();
+            await clientTentacle.TentacleClient.ExecuteScriptAssumingException(startScriptCommand, logs, CancellationToken);
+
+            // We Can not verify what will be in the logs because of race conditions in tentacle.
+            // The last complete script which we fail might come back with the logs.
```
nathanwoctopusdeploy, line 224: This isn't correct. CompleteScript on ScriptServiceV2 returns void. The last GetStatus call is guaranteed to return all logs.

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionScriptServiceV2IsNotRetried.cs
```diff
@@ -0,0 +1,231 @@
...
+                                responseMessageTcpKiller.KillConnectionOnNextResponse();
+                            }
+                        })
+                        .Build())
+                    .Build())
+                .Build(CancellationToken);
+
+            var waitForFile = Path.Combine(clientTentacle.TemporaryDirectory.DirectoryPath, "waitforme");
+
+            var startScriptCommand = new StartScriptCommandV2Builder()
```
nathanwoctopusdeploy, line 50: You can delete 

```
.WithIsolation(ScriptIsolationLevel.FullIsolation)
.WithMutexName("bob")
```

As this test does not require them and is not trying to assert any behaviour around script isolation or mutex locks

nathanwoctopusdeploy (changes requested):
Looking good. I've left some comments for cleanup/possible bug etc.

Main point would be missing tests around Cancellation (and abandon behaviour during cancellation) when rpc retries are not enabled.

nathanwoctopusdeploy:
[sc-51287]

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionScriptServiceV2IsNotRetried.cs
acodrington: I'll defer to @LukeButters on this one, as this was a copy pasta from his SSv1 tests.

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionScriptServiceV2IsNotRetried.cs
acodrington: Unused code from copy pasta, will remove.

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionScriptServiceV2IsNotRetried.cs
acodrington: This is part of ensuring all scripts are completed , which may not be needed as per your [other comment](https://github.com/OctopusDeploy/OctopusTentacle/pull/534#discussion_r1252405871).

I'm not sure about this in general, so I'll get some more info from @LukeButters

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionScriptServiceV2IsNotRetried.cs
LukeButters: It looks like the first call to start script is failed when the response comes backs, which would mean the script is running. Right?

nathanwoctopusdeploy: Yes good point, but we don't need the script to wait for anything though as the response is failed and we should never retry

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionScriptServiceV2IsNotRetried.cs
nathanwoctopusdeploy: As per Luke's other comment this needs to be left as the script has started.

source/Octopus.Tentacle.Client/Retries/RpcCallNoRetriesHandler.cs
```diff
@@ -0,0 +1,91 @@
...
+                           // Give the actionTask some time to cancel on it's own.
+                           // If it doesn't assume it does not co-operate with cancellationTokens and walk away.
+                           abandonCancellationTokenSource.TryCancelAfter(abandonAfter);
+                       }))
+                {
+                    var abandonTask = abandonCancellationTokenSource.Token.AsTask();
+
+                    try
+                    {
+                        var actionTask = action(ct);
```
sburmanoctopus, line 34: This will likely need to have [the fix](https://github.com/OctopusDeploy/OctopusTentacle/pull/540) Nathan did over the weekend applied to it.

acodrington: Yeah good catch, I've updated 👍🏻

source/Octopus.Tentacle.Client/Scripts/ScriptExecutionOrchestrator.cs
nathanwoctopusdeploy: @acodrington thoughts ⬆️

source/Octopus.Tentacle.Client/Scripts/ScriptExecutionOrchestrator.cs
```diff
@@ -358,12 +418,13 @@ void CancelAfter()
             }
             else
             {
-                var completeStatusV1 = rpcCallExecutor.Execute(
+                var completeStatusV1 = await rpcCallExecutor.ExecuteWithNoRetries(
                     RpcCall.Create<IScriptService>(nameof(IScriptService.CompleteScript)),
                     ct => scriptServiceV1.CompleteScript(new CompleteScriptCommand(lastStatusResponse.Ticket, lastStatusResponse.NextLogSequence), new HalibutProxyRequestOptions(ct)),
+                    abandonActionOnCancellation: true,
```
nathanwoctopusdeploy, line 424: Are we changing the behaviour of the original code here?

source/Octopus.Tentacle.Client/Scripts/ScriptExecutionOrchestrator.cs
```diff
@@ -266,12 +311,13 @@ await Task.Delay(scriptObserverBackOffStrategy.GetBackoff(++iteration), scriptEx
             }
             else
             {
-                var scriptStatusResponseV1 = rpcCallExecutor.Execute(
+                var scriptStatusResponseV1 = await rpcCallExecutor.ExecuteWithNoRetries(
                     RpcCall.Create<IScriptService>(nameof(IScriptService.GetStatus)),
                     ct => scriptServiceV1.GetStatus(new ScriptStatusRequest(lastStatusResponse.Ticket, lastStatusResponse.NextLogSequence), new HalibutProxyRequestOptions(ct)),
+                    abandonActionOnCancellation: true,
```
nathanwoctopusdeploy, line 317: Are we changing the behaviour of the original code here?

source/Octopus.Tentacle.Client/Scripts/ScriptExecutionOrchestrator.cs
```diff
@@ -119,11 +136,12 @@ public async Task<ScriptStatusResponseV2> ExecuteScript(CancellationToken script
             {
                 var startScriptCommandV1 = Map(startScriptCommand);
 
-                var scriptTicket = rpcCallExecutor.Execute(
+                var scriptTicket = await rpcCallExecutor.ExecuteWithNoRetries(
                     RpcCall.Create<IScriptService>(nameof(IScriptService.StartScript)),
                     ct => scriptServiceV1.StartScript(startScriptCommandV1, new HalibutProxyRequestOptions(ct)),
+                    abandonActionOnCancellation: true,
```
nathanwoctopusdeploy, line 142: Are we changing the behaviour of the original code here?

source/Octopus.Tentacle.Client/Scripts/ScriptExecutionOrchestrator.cs
```diff
@@ -320,14 +381,13 @@ await Task.Delay(scriptObserverBackOffStrategy.GetBackoff(++iteration), scriptEx
...
                 try
                 {
-                    var actionTask = Task.Run(() =>
-                    {
-                        rpcCallExecutor.Execute(
+                    var actionTask =
+                        rpcCallExecutor.ExecuteWithNoRetries(
                             RpcCall.Create<IScriptServiceV2>(nameof(IScriptServiceV2.CompleteScript)),
                             ct => scriptServiceV2.CompleteScript(new CompleteScriptCommandV2(lastStatusResponse.Ticket), new HalibutProxyRequestOptions(ct)),
+                            abandonActionOnCancellation: true,
```
nathanwoctopusdeploy, line 388: Are we changing the behaviour of the original code here?

source/Octopus.Tentacle.Client/TentacleClient.cs
```diff
@@ -128,23 +145,35 @@ public async Task<UploadResult> UploadFile(string fileName, string path, DataStr
...
+                        DownloadFileAction,
+                        logger,
+                        abandonActionOnCancellation: false,
+                        operationMetricsBuilder,
+                        cancellationToken).ConfigureAwait(false);
+                }
+                else
+                {
+                    return await rpcCallExecutor.ExecuteWithNoRetries(
+                        RpcCall.Create<IClientFileTransferService>(nameof(IClientFileTransferService.DownloadFile)),
```
nathanwoctopusdeploy, line 171: Why `IClientFileTransferService` rather than consistency by using `IFileTransferService`

source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
```diff
@@ -63,7 +69,7 @@ internal RpcCallExecutor(RpcCallRetryHandler rpcCallRetryHandler, ITentacleClien
                         onTimeoutAction: async (timeoutDuration, _) =>
                         {
                             await Task.CompletedTask;
-                            logger.Info($"Could not communicating with Tentacle after retrying for {timeoutDuration.TotalSeconds} seconds. No more retries will be attempted.");
+                            logger.Info($"Could not communicate with Tentacle after retrying for {timeoutDuration.TotalSeconds} seconds. No more retries will be attempted.");
                         },
                         abandonActionOnCancellation,
                         abandonAfter: TimeSpan.FromSeconds(5),
```
nathanwoctopusdeploy, line 75: Can we use the `AbandonAfter` field here?

source/Octopus.Tentacle.Tests.Integration/CapabilitiesServiceV2Test.cs
```diff
@@ -111,5 +114,54 @@ public async Task CapabilitiesResponseShouldBeCached(TentacleType tentacleType,
...
+
+            var logs = new List<ProcessOutput>();
+            Assert.ThrowsAsync<HalibutClientException>(async () => await clientTentacle.TentacleClient.ExecuteScriptAssumingException(startScriptCommand, logs, CancellationToken));
+
+            var allLogs = logs.JoinLogs();
+
+            allLogs.Should().NotContain("hello");
+            capabilitiesServiceExceptions.GetCapabilitiesLatestException.Should().NotBeNull();
+            capabilitiesServiceCallCounts.GetCapabilitiesCallCountStarted.Should().Be(1);
+            capabilitiesServiceCallCounts.GetCapabilitiesCallCountComplete.Should().Be(1);
```
nathanwoctopusdeploy, line 164: Should we ensure that start script was not called and that we stopped at get status when the failure occured?

source/Octopus.Tentacle.Tests.Integration/ClientFileTransfersAreNotRetriedWhenRetriesAreDisabled.cs
```diff
@@ -0,0 +1,98 @@
...
+
+            var remotePath = Path.Combine(clientTentacle.TemporaryDirectory.DirectoryPath, "UploadFile.txt");
+
+            var uploadFileTask = clientTentacle.TentacleClient.UploadFile(remotePath, DataStream.FromString("Hello"), CancellationToken);
+            
+            Func<Task> action = async () => await uploadFileTask;
+            await action.Should().ThrowAsync<HalibutClientException>();
+            
+            fileTransferServiceException.UploadLatestException.Should().NotBeNull();
+            fileTransferServiceCallCounts.UploadFileCallCountStarted.Should().Be(1);
```
nathanwoctopusdeploy, line 54: Should we check `remotePath` to make sure the file really wasn't uploaded?

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionScriptServiceV2IsNotRetriedWhenRetriesAreDisabled.cs
```diff
@@ -0,0 +1,225 @@
...
+using Octopus.Tentacle.Tests.Integration.Support;
+using Octopus.Tentacle.Tests.Integration.Util;
+using Octopus.Tentacle.Tests.Integration.Util.Builders;
+using Octopus.Tentacle.Tests.Integration.Util.Builders.Decorators;
+using Octopus.Tentacle.Tests.Integration.Util.TcpTentacleHelpers;
+
+namespace Octopus.Tentacle.Tests.Integration
+{
+    [IntegrationTestTimeout]
+    public class ClientScriptExecutionScriptServiceV2IsNotRetriedWhenRetriesAreDisabled : IntegrationTest
```
nathanwoctopusdeploy, line 21: It might make sense to move the GetCapabilities test in to here as it is part of the ExecuteScript flow

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionScriptServiceV2IsNotRetriedWhenRetriesAreDisabled.cs
```diff
@@ -0,0 +1,225 @@
...
+
+            List<ProcessOutput> logs = new List<ProcessOutput>();
+            Logger.Information("Starting and waiting for script exec");
+            Assert.ThrowsAsync<HalibutClientException>(async () => await clientTentacle.TentacleClient.ExecuteScriptAssumingException(startScriptCommand, logs, CancellationToken));
+            Logger.Information("Exception thrown.");
+
+            // Let the script finish.
+            File.WriteAllText(waitForFile, "");
+            var legacyTentacleClient = clientTentacle.LegacyTentacleClientBuilder().Build(CancellationToken);
+            await Wait.For(() => legacyTentacleClient.ScriptService.GetStatus(new ScriptStatusRequest(scriptStatusRequest.Ticket, scriptStatusRequest.LastLogSequence)).State == ProcessState.Complete, CancellationToken);
```
nathanwoctopusdeploy, line 114: `ScripService` and `ScriptServiceV2` don't chare state, so I don't think this is doing what you intend. Instead `ScriptService` won't know about the script ticket and return ProcessState.Complete

You need to use ScriptSerciceV2 here to GetStatus and ensure the script has finished

nathanwoctopusdeploy: I also don't think you need to capture `scriptStatusRequest` as `startScriptCommand` has the Ticket you need and you don't need to worry about passing the correct LastLogSequence

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionScriptServiceV2IsNotRetriedWhenRetriesAreDisabled.cs
```diff
@@ -0,0 +1,225 @@
...
+                .WithRetryDuration(TimeSpan.FromMinutes(4))
+                .WithTentacleServiceDecorator(new TentacleServiceDecoratorBuilder()
+                    .LogCallsToScriptServiceV2()
+                    .RecordExceptionThrownInScriptServiceV2(out var scriptServiceExceptions)
+                    .CountCallsToScriptServiceV2(out var scriptServiceCallCounts)
+                    .DecorateScriptServiceV2With(new ScriptServiceV2DecoratorBuilder()
+                        .BeforeGetStatus((inner, request) =>
+                        {
+                            cts.Cancel();
+                            scriptStatusRequest = request;
```
nathanwoctopusdeploy, line 144: Not needed as per other comment

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionScriptServiceV2IsNotRetriedWhenRetriesAreDisabled.cs
```diff
@@ -0,0 +1,225 @@
...
+                .Build();
+
+            List<ProcessOutput> logs = new List<ProcessOutput>();
+
+            Assert.ThrowsAsync<HalibutClientException>(async () => await clientTentacle.TentacleClient.ExecuteScriptAssumingException(startScriptCommand, logs, cts.Token));
+
+            // Let the script finish.
+            File.WriteAllText(waitForFile, "");
+            var legacyTentacleClient = clientTentacle.LegacyTentacleClientBuilder().Build(CancellationToken);
+            await Wait.For(() => legacyTentacleClient.ScriptService.GetStatus(new ScriptStatusRequest(scriptStatusRequest.Ticket, scriptStatusRequest.LastLogSequence)).State == ProcessState.Complete, CancellationToken);
```
nathanwoctopusdeploy, line 173: Not doing as you intended as per other comment

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionScriptServiceV2IsNotRetriedWhenRetriesAreDisabled.cs
```diff
@@ -0,0 +1,225 @@
...
+
+            var allLogs = logs.JoinLogs();
+
+            allLogs.Should().NotContain("AllDone");
+            scriptServiceExceptions.CancelScriptLatestException.Should().NotBeNull();
+            scriptServiceCallCounts.StartScriptCallCountStarted.Should().Be(1);
+            scriptServiceCallCounts.CancelScriptCallCountStarted.Should().Be(1);
+            scriptServiceCallCounts.CompleteScriptCallCountStarted.Should().Be(0);
```
nathanwoctopusdeploy, line 181: I think this is a bug. When we Cancel Script Execution we should also try and clean-up

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionScriptServiceV2IsNotRetriedWhenRetriesAreDisabled.cs
```diff
@@ -0,0 +1,225 @@
...
+                .Build(CancellationToken);
+            
+            var startScriptCommand = new StartScriptCommandV2Builder()
+                .WithScriptBody(new ScriptBuilder()
+                    .Print("hello")
+                    .Print("AllDone"))
+                .Build();
+
+            List<ProcessOutput> logs = new List<ProcessOutput>();
+            await clientTentacle.TentacleClient.ExecuteScriptAssumingException(startScriptCommand, logs, CancellationToken);
```
nathanwoctopusdeploy, line 216: `ExecuteScriptAssumingException` seems confusing here as it doesn't error!!

source/Octopus.Tentacle.Client/Retries/RpcCallNoRetriesHandler.cs
```diff
@@ -0,0 +1,103 @@
...
+{
+    public class RpcCallNoRetriesHandler
+    {
+        public async Task ExecuteWithNoRetries(
+            Func<CancellationToken, Task> action,
+            bool abandonActionOnCancellation,
+            TimeSpan abandonAfter,
+            CancellationToken cancellationToken)
+        {
+            await ((Func<CancellationToken, Task>)(async ct =>
```
sburmanoctopus, line 15: Just wondering, what is the purpose of wrapping the entire contents of this method in a lambda, and calling it straight away?

I suppose it starts a new thread context at that point, but we would create a new thread context as soon as we hit an `await` in this method, right?

There is likely a subtlety I'm missing. 

But if not, it would be good if we could simplify this method to this:
```
public async Task ExecuteWithNoRetries(
            Func<CancellationToken, Task> action,
            bool abandonActionOnCancellation,
            TimeSpan abandonAfter,
            CancellationToken cancellationToken)
        {
            if (!abandonActionOnCancellation)
            {
                await action(cancellationToken).ConfigureAwait(false);
            }

            using var abandonCancellationTokenSource = new CancellationTokenSource();
            using (cancellationToken.Register(() =>
                   {
                       // Give the actionTask some time to cancel on it's own.
                       // If it doesn't assume it does not co-operate with cancellationTokens and walk away.
                       abandonCancellationTokenSource.TryCancelAfter(abandonAfter);
                   }))
            {
                var abandonTask = abandonCancellationTokenSource.Token.AsTask();

                try
                {
                    var actionTask = action(cancellationToken);
                    var completedTask = await Task.WhenAny(actionTask, abandonTask).ConfigureAwait(false);
                    if (actionTask != completedTask)
                    {
                        actionTask.IgnoreUnobservedExceptions();
                    }

                    await completedTask.ConfigureAwait(false);
                }
                catch (Exception e) when (e is OperationCanceledException)
                {
                    if (abandonCancellationTokenSource.IsCancellationRequested)
                    {
                        throw new OperationAbandonedException(e, abandonAfter);
                    }

                    throw;
                }
            }
        }
```

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionCanBeCancelledWhenRetriesAreDisabled.cs
```diff
@@ -0,0 +1,506 @@
...
+
+            // ASSERT
+            // The ExecuteScript operation threw an OperationCancelledException
+            actualException.Should().BeOfType<OperationCanceledException>();
+
+            // If the rpc call could be cancelled then the correct error was recorded
+            switch (expectedFlow)
+            {
+                case ExpectedFlow.CancelRpcAndExitImmediately:
+                case ExpectedFlow.CancelRpcThenCancelScriptThenCompleteScript:
```
nathanwoctopusdeploy, line 205: This case will never be used

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionCanBeCancelledWhenRetriesAreDisabled.cs
```diff
@@ -0,0 +1,506 @@
...
+                    break;
+                case ExpectedFlow.AbandonRpcThenCancelScriptThenCompleteScript:
+                    scriptServiceV2Exceptions.StartScriptLatestException?.Should().BeOfType<HalibutClientException>();
+                    break;
+                default:
+                    throw new NotSupportedException();
+            }
+
+            // If the rpc call could be cancelled we cancelled quickly
+            if (expectedFlow == ExpectedFlow.CancelRpcThenCancelScriptThenCompleteScript)
```
nathanwoctopusdeploy, line 216: This will never be true

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionCanBeCancelledWhenRetriesAreDisabled.cs
```diff
@@ -0,0 +1,506 @@
...
+
+            scriptServiceV2CallCounts.GetStatusCallCountStarted.Should().Be(0, "Test should not have not proceeded past StartScript before being Cancelled");
+
+            switch (expectedFlow)
+            {
+                case ExpectedFlow.CancelRpcAndExitImmediately:
+                    scriptServiceV2CallCounts.CancelScriptCallCountStarted.Should().Be(0);
+                    scriptServiceV2CallCounts.CompleteScriptCallCountStarted.Should().Be(0);
+                    break;
+                case ExpectedFlow.CancelRpcThenCancelScriptThenCompleteScript:
```
nathanwoctopusdeploy, line 235: This case is not needed

source/Octopus.Tentacle.Client/Scripts/ScriptExecutionOrchestrator.cs
acodrington: Yeah, good pickup, it should be the same 👍🏻

source/Octopus.Tentacle.Tests.Integration/ClientFileTransfersAreNotRetriedWhenRetriesAreDisabled.cs
acodrington: We can't check that because of the way we're killing the connection 
(unless I'm misunderstanding something). The file will get uploaded, but the _response_ will never arrive.

acodrington: We can't check that, due to the way we kill the connection. We kill the connection when we receive a _response_ from the Tentacle, not when sending the request. Therefore, the file will have been uploaded, but we just don't get the response.

We could add a successful/unsuccessful count to the call count decorators? We currently just count 'started' and 'completed'.

nathanwoctopusdeploy (approved):
Lovely job 👍

nathanwoctopusdeploy (approved):
Lovely job. LGTM

PR #535, opened 2023-07-03
Use the updated (latest available) TeamCity.VSTest.TestAdapter.

michaelongso:
# Background

The nightly build for Tentacle has been intermittently failing due to the test runner failing to parse the report file.

There's no exact area that can pinpoint the issue, however, one of the potential causes could be the outdated `TeamCity.VSTest.TestAdapter` plugin used in Tentacle.

# Results

Updating the version to the (latest available) `1.0.36`
- There're a number of [fixes](https://github.com/JetBrains/TeamCity.VSTest.TestAdapter/releases) since the `1.0.15` release

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


LukeButters (approved):

LukeButters:
@michaelongso do we know if the report file is being created but can't be parsed or if it is not being created at all? If we don't already it might make sense to change the TC build to capture the report as a artifact that way if this issue happens again we can inspect the report file.

michaelongso:
> @michaelongso do we know if the report file is being created but can't be parsed or if it is not being created at all? If we don't already it might make sense to change the TC build to capture the report as a artifact that way if this issue happens again we can inspect the report file.

The file is there. It's locked by another process.

PR #540, opened 2023-07-08
Observe errors from task that are abandoned to stop UnobservedExceptions being handled by the Finalizer

nathanwoctopusdeploy:
# Background

As part of TentacleClient script execution, cancelling a script execution can result in the task performing the RPC call to be abandoned and not awaited.

If an error is thrown by this abandoned task, it will result in the Finalizer handling and rethrowing the exception as it has not been observed.

# Results

## Before

image

## After

If the task is being abandoned and Fails, we wrap it in a ContinueWith to ensure the exception is observer

# How to review this PR

No automated Testing has been added for this issue, but it has been manually verified that the unobserved error no longer exists.

```
From 89f26012a742adb2888b8937f829c07bf4ee2a8a Mon Sep 17 00:00:00 2001
From: Nathan Willoughby <nathan.willoughby@octopus.com>
Date: Sat, 8 Jul 2023 21:25:15 +1000
Subject: [PATCH] Manually Test Unobserved Exception

---
 .../SetUpFixtures.cs                          | 29 ++++++++++++
 .../Support/IntegrationTest.cs                | 10 +++-
 .../UnobservervedExceptionFixture.cs          | 46 +++++++++++++++++++
 3 files changed, 84 insertions(+), 1 deletion(-)
 create mode 100644 source/Octopus.Tentacle.Tests.Integration/SetUpFixtures.cs
 create mode 100644 source/Octopus.Tentacle.Tests.Integration/UnobservervedExceptionFixture.cs

diff --git a/source/Octopus.Tentacle.Tests.Integration/SetUpFixtures.cs b/source/Octopus.Tentacle.Tests.Integration/SetUpFixtures.cs
new file mode 100644
index 00000000..5580f6a4
--- /dev/null
+++ b/source/Octopus.Tentacle.Tests.Integration/SetUpFixtures.cs
@@ -0,0 +1,29 @@
+﻿using System;
+using System.Threading;
+using NUnit.Framework;
+
+{
+    internal class SetUpFixtures
+    {
+        [SetUpFixture]
+        public class TestsSetupClass
+        {
+            [OneTimeSetUp]
+            public void GlobalSetup()
+            {
+                TaskScheduler.UnobservedTaskException += (s, e) =>
+                {
+#pragma warning disable CS0219
+                    // Break here to observe e
+                    int i = 0;
+#pragma warning restore CS0219
+                };
+            }
+        }
+    }
+}
diff --git a/source/Octopus.Tentacle.Tests.Integration/Support/IntegrationTest.cs b/source/Octopus.Tentacle.Tests.Integration/Support/IntegrationTest.cs
index 1bfb4976..2a37c762 100644
--- a/source/Octopus.Tentacle.Tests.Integration/Support/IntegrationTest.cs
+++ b/source/Octopus.Tentacle.Tests.Integration/Support/IntegrationTest.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Threading;
 using NUnit.Framework;
 using Octopus.Tentacle.Tests.Integration.Util;
 using Serilog;
@@ -36,8 +37,15 @@ public void SetUp()
         }
 
         [TearDown]
-        public void TearDown()
+        public async Task TearDown()
         {
+            // invoking garbage collector explicitly
+            Console.WriteLine("GC..");
+            GC.Collect();
+            GC.WaitForPendingFinalizers();
+
+            await Task.Delay(TimeSpan.FromSeconds(10));
+
             Logger.Information("Tearing down");
             if (cancellationTokenSource != null)
             {
diff --git a/source/Octopus.Tentacle.Tests.Integration/UnobservervedExceptionFixture.cs b/source/Octopus.Tentacle.Tests.Integration/UnobservervedExceptionFixture.cs
new file mode 100644
index 00000000..f464b5dd
--- /dev/null
+++ b/source/Octopus.Tentacle.Tests.Integration/UnobservervedExceptionFixture.cs
@@ -0,0 +1,46 @@
+﻿using System;
+using NUnit.Framework;
+
+{
+    public class UnobservervedExceptionFixture : IntegrationTest
+    {
+        [Test]
+        public void ShouldCreateUnobservedException()
+        {
+            try
+            {
+#pragma warning disable CS4014
+                ThrowsAsyncHelper();
+#pragma warning restore CS4014
+            }
+            catch (Exception ex)
+            {
+                Console.WriteLine($"Caught in try/catch  : {ex.Message}");
+            }
+        }
+
+#pragma warning disable CS1998
+        private static async Task ThrowsAsyncHelper()
+#pragma warning restore CS1998
+#pragma warning restore CS1998
+        {
+#pragma warning disable CS4014
+            ThrowsAsync();
+#pragma warning restore CS4014
+        }
+
+#pragma warning disable CS1998
+        private static async Task ThrowsAsync()
+#pragma warning restore CS1998
+        {
+            Console.WriteLine("Throwing...");
+            throw new Exception("Boom!");
+        }
+    }
+}
-- 
2.35.1.windows.2
```

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Client/Retries/TaskExtensionMethods.cs
+++ b/source/Octopus.Tentacle.Client/Retries/TaskExtensionMethods.cs
@@ -1,0 +1,27 @@
+﻿using System.Threading.Tasks;
+
+namespace Octopus.Tentacle.Client.Retries
+{
+    static class TaskExtensionMethods
+    {
+        public static void IgnoreUnobservedExceptions(this Task task)
+        {
+            if (task.IsCompleted)
+            {
+                if (task.IsFaulted)
+                {
+                    var observedException = task.Exception;
+                }
+
+                return;
+            }
+
+            task.ContinueWith(
+                t =>
+                {
+                    var observedException = t.Exception;
+                },
+                TaskContinuationOptions.OnlyOnFaulted | TaskContinuationOptions.ExecuteSynchronously);
+        }
+    }
+}
--- a/source/Octopus.Tentacle.Client/Retries/RpcCallRetryHandler.cs
+++ b/source/Octopus.Tentacle.Client/Retries/RpcCallRetryHandler.cs
@@ -128,7 +128,14 @@ namespace Octopus.Tentacle.Client.Retries
                         try
                         {
                             var actionTask = action(ct);
-                            return await (await Task.WhenAny(actionTask, abandonTask).ConfigureAwait(false)).ConfigureAwait(false);
+                            var completedTask = await Task.WhenAny(actionTask, abandonTask).ConfigureAwait(false);
+
+                            if (actionTask != completedTask)
+                            {
+                                actionTask.IgnoreUnobservedExceptions();
+                            }
+
+                            return await completedTask.ConfigureAwait(false);
                         }
                         catch (Exception e) when (e is OperationCanceledException)
                         {


nathanwoctopusdeploy:
[sc-53482]

source/Octopus.Tentacle.Client/Retries/TaskExtensionMethods.cs
```diff
@@ -0,0 +1,27 @@
...
+
+                return;
+            }
+
+            task.ContinueWith(
+                t =>
+                {
+                    var observedException = t.Exception;
+                },
+                TaskContinuationOptions.OnlyOnFaulted | TaskContinuationOptions.ExecuteSynchronously);
```
LukeButters, line 24: I wonder if you can use:
```
public static async Task LogAndSwallowAnyException(this Task task, logger) {
    try {
        await task;
    } catch(Exception e) {
        logger.info(e);
    }
}
```
instead.

LukeButters (approved):
LGTM 
I wonder if we could add something to the integration tests to look for any unobserved exceptions and cause some sort of failure if any of them look to be within the client.

nathanwoctopusdeploy:
> LGTM I wonder if we could add something to the integration tests to look for any unobserved exceptions and cause some sort of failure if any of them look to be within the client.

We could probably add something but the Finalizer is never guaranteed to run so it would not be guaranteed to catch these issues. It would be better than nothing though. We could potentially try and force GC and wait a short time for the Finalizer to hopefully run at the end of the Test Run to see if any unobserved errors occur and try and find a way to fail something. That might help catch some stuff!! I'll card it!

source/Octopus.Tentacle.Client/Retries/TaskExtensionMethods.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System.Threading.Tasks;
+
+namespace Octopus.Tentacle.Client.Retries
+{
+    static class TaskExtensionMethods
+    {
+        public static void IgnoreUnobservedExceptions(this Task task)
+        {
+            if (task.IsCompleted)
```
sburmanoctopus, line 9: I wonder if we should do the `if (task.IsCompleted)` check after we set the `ContinueWith`.

Because this is all async, there is a small window of opportunity where the task may complete after we do the 'if' check, but before we register the `ContinueWith`.

source/Octopus.Tentacle.Client/Retries/TaskExtensionMethods.cs
nathanwoctopusdeploy: I did a fair bit of reading about it, and this was the approach people took and it seems to work as expected.

I'm not saying awaiting the task wouldn't also work but you may want to ensure it's not a blocking wait (something like `async void LogAndSwallowAnyException`) and I wouldn't log as we aren't really interested in the error.

sburmanoctopus (approved):
LGTM. It's a shame we have to do this. Oh well.

source/Octopus.Tentacle.Client/Retries/TaskExtensionMethods.cs
LukeButters: I think just doing the "what everyone recommends" is a good idea, don't change what ain't broke :D

source/Octopus.Tentacle.Client/Retries/TaskExtensionMethods.cs
nathanwoctopusdeploy: I don't think there is a race condition. I think it's an optimisation to avoid having to add the `ContinueWith` if not needed

image

```
async Task Main()
{
	// Task not completed so obsevrer with ContinueWith
	var t = Fail();
	t.IgnoreUnobservedExceptions();

	// Task already completed
	var t2 = Fail();
	await Task.Delay(TimeSpan.FromSeconds(2));
	t2.IgnoreUnobservedExceptions();

	// Race condition where task completes after IsCompleted check but before ContinueWith is hooked up
	var t3 = Fail();
	await Task.Delay(TimeSpan.FromSeconds(2));
	t3.IgnoreUnobservedExceptionsWithoutCheckingIsCompleted();

	await Task.Delay(TimeSpan.FromSeconds(10));
}

async Task Fail()
{
	await Task.Delay(TimeSpan.FromSeconds(1));
	throw new Exception();
}

static class TaskExtensionMethods
{
	public static void IgnoreUnobservedExceptions(this Task task)
	{
		if (task.IsCompleted)
		{
			if (task.IsFaulted)
			{
				var observedException = task.Exception;
				"Observed as Task was already Complete".Dump();
			}

			return;
		}

		task.ContinueWith(
			t =>
			{
				var observedException = t.Exception;
				"Observed in ContinueWith".Dump();
			},
			TaskContinuationOptions.OnlyOnFaulted | TaskContinuationOptions.ExecuteSynchronously);
	}

	public static void IgnoreUnobservedExceptionsWithoutCheckingIsCompleted(this Task task)
	{
		if(!task.IsCompleted)
		{
			throw new Exception("Expecting Task to be Completed for this to simulate the race condition");
		}
		
		task.ContinueWith(
			t =>
			{
				var observedException = t.Exception;
				"Observed in ContinueWith".Dump();
			},
			TaskContinuationOptions.OnlyOnFaulted | TaskContinuationOptions.ExecuteSynchronously);
	}
}

// You can define other methods, fields, classes and namespaces here
```

sburmanoctopus: Ah, I see. If the task completed after the 'if' but before the `ContinueWith`, then it doesn't matter, because adding the `ContinueWith` will make it run straight away. Like you said, the first part is just an optimisation.

PR #541, opened 2023-07-08
Change how Error Response Caching is configured for the Halibut Runtime when using TentacleClient

nathanwoctopusdeploy:
# Background

Stop the TentacleClient constructor configuring the error response caching for Halibut, as this results in the HalibutRuntime being configured every time the TentacleClient is created rather than once.

# Results

## Before

TentacleClient was being retained in Memory due to the reference created registering the ErrorResponseMessageCachingAction

The delegate handler chain for OverideErrorResponseMessageCachingAction had a new action added to the chain for every TentacleClient created

image

image

## After

Server / other clients will have to perform an additional call to configure Halibut for Tentacle Client usage, which is not ideal but we need to ensure this is only done once.

`TentacleClient.CacheServiceWasNotFoundResponseMessages(halibutRuntime);`

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Client/TentacleClient.cs
+++ b/source/Octopus.Tentacle.Client/TentacleClient.cs
@@ -27,12 +27,27 @@ namespace Octopus.Tentacle.Client
         readonly IClientFileTransferService fileTransferServiceV1;
         readonly IClientCapabilitiesServiceV2 capabilitiesServiceV2;
 
+        public static void CacheServiceWasNotFoundResponseMessages(IHalibutRuntime halibutRuntime)
+        {
+            var innerHandler = halibutRuntime.OverrideErrorResponseMessageCaching;
+            halibutRuntime.OverrideErrorResponseMessageCaching = response =>
+            {
+                if (BackwardsCompatibleCapabilitiesV2Helper.ExceptionTypeLooksLikeTheServiceWasNotFound(response.Error.HalibutErrorType) ||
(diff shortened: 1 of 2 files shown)


nathanwoctopusdeploy:
[sc-53493]

source/Octopus.Tentacle.Client/TentacleClient.cs
```diff
@@ -48,18 +63,12 @@ public class TentacleClient : ITentacleClient
             this.scriptObserverBackOffStrategy = scriptObserverBackOffStrategy;
             this.tentacleClientObserver = tentacleClientObserver;
 
-            var innerHandler = halibutRuntime.OverrideErrorResponseMessageCaching;
```
sburmanoctopus, line 51: Just to make sure I understand this correctly, it looks like every time this constructor is called, we take a reference to the existing inner `OverrideErrorResponseMessageCaching`, then wrap it another one.

Each time this is called, it adds another 'layer' to the onion, right?

And even though this lambda doesn't reference any properties of this class, it still results holds a reference to it, and that results in the memory leak you saw, right?

sburmanoctopus (approved):
Nice find! That one would have sucked to track down in the wild 😁

source/Octopus.Tentacle.Client/TentacleClient.cs
nathanwoctopusdeploy: Yep 🧅

PR #570, opened 2023-08-25
Build a Windows installer for the net6.0 version of Tentacle

nathanwoctopusdeploy:
# Background

This PR builds a Windows installer for the Net 6.0 version of Tentacle.

Tentacle Manager is not supported on net6.0 so the installer only installs Tentacle.

The new MSI's get bundled in the CrossPlatformBundle which is later used by Server and the Deployment pipeline to push the MSI's to a download location available to customer

image

The net6.0 installer adds a shortcut to Tentacle instead of Tentacle Manager and runs Tentacle.exe --help when first installed instead of the Tentacle Manager

Net48 installer won't install without net48 installed

image

Net6.0 installer will 

image

image

## Deployment Process

THe deployment process has been tweaked so it correctly renames the new MSI filename to include the version number before pushing to Aws Storage

image

This works for old and new filenames 

image

## Also

Fixed a bug in the text on the alert when trying to run the x86 installer on an x64 machine

image

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


nathanwoctopusdeploy:
[sc-58904]

.gitignore
```diff
@@ -51,3 +51,4 @@ _artifacts
 _build
 
 .DS_Store
+.env
```
nathanwoctopusdeploy, line 54: Unrelated but it keeps annoying me 😆

source/Octopus.Manager.Tentacle/Octopus.Manager.Tentacle.csproj
```diff
@@ -10,6 +10,7 @@
     <UseWindowsForms>true</UseWindowsForms>
     <UseWPF>true</UseWPF>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <EnableWindowsTargeting>true</EnableWindowsTargeting>
```
nathanwoctopusdeploy, line 13: May not be needed for this PR but this allows Linux to build the net6.0 version of these Windows-specific apps
If we don't have these the build fails

source/Octopus.Tentacle/Commands/RunAgentCommand.cs
```diff
@@ -101,6 +102,7 @@ protected override void Start()
             Environment.SetEnvironmentVariable(EnvironmentVariables.TentacleProgramDirectoryPath, Path.GetDirectoryName(exePath));
             Environment.SetEnvironmentVariable(EnvironmentVariables.AgentProgramDirectoryPath, Path.GetDirectoryName(exePath));
             Environment.SetEnvironmentVariable(EnvironmentVariables.TentacleVersion, appVersion.ToString());
+            Environment.SetEnvironmentVariable(EnvironmentVariables.TentacleNetFrameworkDescription, RuntimeInformation.FrameworkDescription);
```
nathanwoctopusdeploy, line 105: This will allow Server to ask Tentacle, via a script, what framework it is running under, which may be needed for the changes in Server.

By adding it now, we guarantee that all Windows net6.0 installations have this environment variable set so we can rely on its existence.

source/Tentacle.sln
```diff
@@ -54,10 +54,10 @@ Global
 		Release-net6.0-linux-arm64|Any CPU = Release-net6.0-linux-arm64|Any CPU
 		Release-net6.0-linux-musl-x64|Any CPU = Release-net6.0-linux-musl-x64|Any CPU
 		Release-net6.0-linux-x64|Any CPU = Release-net6.0-linux-x64|Any CPU
+		Release-net6.0-osx-arm64|Any CPU = Release-net6.0-osx-arm64|Any CPU
```
nathanwoctopusdeploy, line 57: These were just missing and Visual Studio loves to add missing stuff

installer/Octopus.Tentacle.Installer/Product.wxs
```diff
@@ -42,16 +43,15 @@
     Ensure that they're not trying to run the 32-bit installer on a 64-bit Operating System.
     -->
     <?if $(var.Platform) = x86 ?>
-    <Condition Message="This installer is only supported on 64-bit operating systems. Please download and run the 64-bit installer."><![CDATA[Not VersionNT64]]></Condition>
+    <Condition Message="This installer is not supported on 64-bit operating systems. Please download and run the 64-bit installer."><![CDATA[Not VersionNT64]]></Condition>
```
nathanwoctopusdeploy, line 46: Slightly contradicting error message ;)

build/Build.Pack.cs
```diff
@@ -204,6 +213,10 @@ void BuildMsiInstallerForPlatform(MSBuildTargetPlatform platform, AbsolutePath w
                     wixProjectFile.ReplaceTextInFile("{WixTargetsPath}", wixNugetPackagePath / "tools" / "Wix.targets");
                     wixProjectFile.ReplaceTextInFile("{WixTasksPath}", wixNugetPackagePath / "tools" / "wixtasks.dll");
 
+                    var tentacleInstallerWixProduct = RootDirectory / "installer" / "Octopus.Tentacle.Installer" / "Product.wxs";
```
sburmanoctopus, line 216: I have a faint memory that you can pass arguments into Wix when building it.

If this is true, would that be nicer than manipulating the file?

nathanwoctopusdeploy: If you know how, yep, I was just following our existing pattern

sburmanoctopus: Although, it does look like we already manipulate the files...

sburmanoctopus: Maybe a Sharpening exercise 😁

sburmanoctopus (approved):
LGTM. I'm not super familiar with our install process, but from what I can see here, it looks good 🙂

PR #576, opened 2023-08-30
Add support for testing net6.0 client (TentacleClient) with net48 service (Tentacle)

acodrington:
# Background

We would like to automatically test a net6 client (i.e. Octopus Server using TentacleClient) communicating with a net48 service (i.e. Tentacle running on .NET Framework). This PR adds support for this by adding a `TentacleRuntime` enum, which is then used by the test system to automatically generate extra test cases from existing tests for this scenario.

The extra test cases will only be generated if:
- Tests are being run on **Windows**, and...
- Tests are being run using the **.NET 6.0** runtime

Test runs on other platforms (Linux, Mac) or other runtimes (.NET Framework 4.8) should be unaffected.

These extra test cases will also be tagged with the test category `Net60ClientNet48Tentacle`, which will allow us to split them into another test run if desired.

# Results

Fixes [sc-56679]

## Before
20230904-114127_rider64_DpqRBDPO2k

## After
20230907-161441_rider64_Gy4a3Go4Do

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Tests.Integration/Support/SetupFixtures/WarmTentacleCache.cs
+++ b/source/Octopus.Tentacle.Tests.Integration/Support/SetupFixtures/WarmTentacleCache.cs
@@ -2,9 +2,7 @@
 using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
-using NUnit.Framework;
 using Octopus.Tentacle.Tests.Integration.Support.TentacleFetchers;
-using Octopus.Tentacle.Tests.Integration.Util;
 using Octopus.Tentacle.Util;
 using Serilog;
 
@@ -28,40 +26,32 @@ namespace Octopus.Tentacle.Tests.Integration.Support.SetupFixtures
                 tasks.Add(Task.Run(async () =>
                 {
                     using var l = await concurrentDownloadLimiter.LockAsync(cts.Token);
-                    await GetTentacle(logger, tentacleVersion);
+                    await GetTentacleVersion(logger, tentacleVersion);
                 }));
             }
 
             Task.WhenAll(tasks).GetAwaiter().GetResult();
         }
 
-        private async Task GetTentacle(ILogger logger, Version tentacleVersion)
+        private async Task GetTentacleVersion(ILogger logger, Version tentacleVersion)
         {
             if (PlatformDetection.IsRunningOnWindows && RuntimeDetection.IsDotNet60)
             {
-                using (var net60TempDir = new TemporaryDirectory())
-                {
-                    logger.Information($"Will fetch tentacle {tentacleVersion} ({TentacleRuntime.DotNet6.GetStringValue()}) if it is not already in cache");
-                    await TentacleFetcher.GetTentacleVersion(net60TempDir.DirectoryPath, tentacleVersion, TentacleRuntime.DotNet6, logger, cts.Token);
-                    logger.Information($"Tentacle {tentacleVersion} ({TentacleRuntime.DotNet6.GetStringValue()}) is now in cache");
-                }
-
-                using (var net48TempDir = new TemporaryDirectory())
-                {
-                    logger.Information($"Will fetch tentacle {tentacleVersion} ({TentacleRuntime.Framework48.GetStringValue()}) if it is not already in cache");
-                    await TentacleFetcher.GetTentacleVersion(net48TempDir.DirectoryPath, tentacleVersion, TentacleRuntime.Framework48, logger, cts.Token);
-                    logger.Information($"Tentacle {tentacleVersion} ({TentacleRuntime.Framework48.GetStringValue()}) is now in cache");
-                }
+                await GetTentacleVersionWithRuntime(logger, tentacleVersion, TentacleRuntime.DotNet6);
+                await GetTentacleVersionWithRuntime(logger, tentacleVersion, TentacleRuntime.Framework48);
             }
             else
             {
-                using (var temporaryDirectory = new TemporaryDirectory())
-                {
-                    logger.Information($"Will fetch tentacle {tentacleVersion} if it is not already in cache");
-                    await TentacleFetcher.GetTentacleVersion(temporaryDirectory.DirectoryPath, tentacleVersion, TentacleRuntime.Default, logger, cts.Token);
-                    logger.Information($"Tentacle {tentacleVersion} is now in cache");
-                }
+                await GetTentacleVersionWithRuntime(logger, tentacleVersion, TentacleRuntime.Default);
             }
+        }
+
+        private async Task GetTentacleVersionWithRuntime(ILogger logger, Version tentacleVersion, TentacleRuntime tentacleRuntime)
+        {
+            using var tempDir = new TemporaryDirectory();
+            logger.Information($"Will fetch tentacle {tentacleVersion} ({tentacleRuntime.GetDescription()}) if it is not already in cache");
+            await TentacleFetcher.GetTentacleVersion(tempDir.DirectoryPath, tentacleVersion, tentacleRuntime, logger, cts.Token);
+            logger.Information($"Tentacle {tentacleVersion} ({tentacleRuntime.GetDescription()}) is now in cache");
         }
 
         public void OneTimeTearDown(ILogger logger)
--- a/source/Octopus.Tentacle.Tests.Integration/Support/SetupFixtures/WarmTentacleCache.cs
+++ b/source/Octopus.Tentacle.Tests.Integration/Support/SetupFixtures/WarmTentacleCache.cs
@@ -12,14 +12,14 @@ namespace Octopus.Tentacle.Tests.Integration.Support.SetupFixtures
 {
     public class WarmTentacleCache : ISetupFixture
         {
-            private CancellationTokenSource cts = new CancellationTokenSource();
+            private CancellationTokenSource cts = new();
             
             public void OneTimeSetUp(ILogger logger)
             {
                 logger.Fatal("Downloading all tentacles now");
                 
                 var tasks = new List<Task>();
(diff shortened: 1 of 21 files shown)


source/Octopus.Tentacle.Tests.Integration/Support/TentacleTestCaseSourceAttribute.cs
```diff
@@ -0,0 +1,378 @@
...
+                                args = new object?[] { item };
+                            }
+
+                            parms = new TestCaseParameters(args);
+                        }
+
+                        // If the test is trying to test .net6.0 TentacleClient against .net48 Tentacle,
+                        // then add the 'Net60ClientNet48Tentacle' category so we can potentially run these tests in a different test run
+                        #if !NETFRAMEWORK
+                        if (item is TentacleConfigurationTestCase {TentacleRuntime: TentacleRuntime.Framework48} testCase)
```
acodrington, line 164: This is the logic for adding the custom category. It only gets added if:

1. Running in .net 6.0
2. The net48 Tentacle is explicitly defined in the test case

source/Octopus.Tentacle.Tests.Integration/Support/TentacleTestCaseSourceAttribute.cs
```diff
@@ -0,0 +1,378 @@
...
+using NUnit.Framework.Internal;
+using NUnit.Framework.Internal.Builders;
+
+namespace Octopus.Tentacle.Tests.Integration.Support
+{
+    /// <summary>
+    /// Indicates the source to be used to provide test fixture instances for a test class.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = true, Inherited = false)]
+    public class TentacleTestCaseSourceAttribute : NUnitAttribute, ITestBuilder, IImplyFixture
```
acodrington, line 42: This code was copied almost verbatim from [this PR in Halibut](https://github.com/OctopusDeploy/Halibut/pull/463), which does the exact same thing.

The only differences are in the class name, and the custom category logic that is added, which I've pointed out explicitly with a PR comment.

source/Octopus.Tentacle.Tests.Integration/Support/SetupFixtures/WarmTentacleCache.cs
```diff
@@ -11,38 +11,63 @@
...
+            if (PlatformDetection.IsRunningOnWindows && RuntimeDetection.IsDotNet60)
             {
-                logger.Fatal("Downloading all tentacles now");
-                
-                var tasks = new List<Task>();
-                var concurrentDownloads = TentacleExeFinder.IsRunningInTeamCity() ? 4 : 1;
-                var concurrentDownloadLimiter = new SemaphoreSlim(concurrentDownloads, concurrentDownloads);
-                
-                foreach (var tentacleVersion in TentacleVersions.AllTestedVersionsToDownload)
+                using (var net60TempDir = new TemporaryDirectory())
```
nathanwoctopusdeploy, line 42: nit: repetition (with param changes) for each of the using blocks

source/Octopus.Tentacle.Tests.Integration/Support/TentacleConfigurationTestCase.cs
```diff
@@ -8,30 +8,35 @@ public class TentacleConfigurationTestCase
     {
         public TentacleType TentacleType { get; }
         public SyncOrAsyncHalibut SyncOrAsyncHalibut { get; }
+        public TentacleRuntime TentacleRuntime { get; set; }
```
nathanwoctopusdeploy, line 11: remove set?

source/Octopus.Tentacle.Tests.Integration/Support/TentacleExeFinder.cs
```diff
@@ -1,40 +1,66 @@
...
                 // Desired path to tentacle.
                 // /opt/TeamCity/BuildAgent/work/639265b01610d682/build/outputs/tentaclereal/tentacle/Tentacle
+                
+            string tentacleDir =
+                version switch
+                {
+                    TentacleRuntime.Framework48 => "tentaclereal-net48",
+                    TentacleRuntime.DotNet6 => "tentaclereal-net6.0",
+                    TentacleRuntime.Default => 
+                #if NETFRAMEWORK
```
nathanwoctopusdeploy, line 34: nit: may be github review page but nesting of the .Defaut switch is a bit difficult to read

source/Octopus.Tentacle.Tests.Integration/Support/TentacleExeFinder.cs
```diff
@@ -1,40 +1,66 @@
...
+                "tentaclereal-net48",
```
nathanwoctopusdeploy, line 35: nit- repetition of magic strings

source/Octopus.Tentacle.Tests.Integration/Support/TentacleExeFinder.cs
```diff
@@ -1,40 +1,66 @@
...
+                tentacleExe = Path.Combine(assemblyDir.Parent.Parent.Parent.FullName, tentacleDir, "tentacle", "Tentacle");
                 return AddExeExtension(tentacleExe);
             }
+
+            string runtimeDir =
+                version switch
+                {
+                    TentacleRuntime.Framework48 => "net48",
+                    TentacleRuntime.DotNet6 => "net6.0",
+                    TentacleRuntime.Default => assemblyDir.Name,
```
nathanwoctopusdeploy, line 52: should this not work in the same way as the teamcity switch does?

source/Octopus.Tentacle.Tests.Integration/Support/TentacleConfigurationsAttribute.cs
```diff
@@ -81,20 +81,31 @@ static class TentacleConfigurationTestCases
                 versions.Add(TentacleVersions.Current);
             }
 
+#if NETFRAMEWORK
+            List<TentacleRuntime> runtimes = new List<TentacleRuntime> { TentacleRuntime.Default };
```
nathanwoctopusdeploy, line 85: nit: var

source/Octopus.Tentacle.Tests.Integration/Support/TentacleConfigurationsAttribute.cs
```diff
@@ -81,20 +81,31 @@ static class TentacleConfigurationTestCases
...
+#endif
+#if !NETFRAMEWORK
```
nathanwoctopusdeploy, line 87: nit: #if #else #endif

source/Octopus.Tentacle.Tests.Integration/Support/TentacleFetchers/TentacleBinaryCache.cs
```diff
@@ -24,47 +24,63 @@ public TentacleBinaryCache(ITentacleFetcher tentacleFetcher, ILogger logger)
...
             var pathBase = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData, Environment.SpecialFolderOption.DoNotVerify);
-            var cacheDir = Path.Combine(pathBase, cachDirName, NugetTentacleFetcher.TentacleBinaryFrameworkForCurrentOs());
+            var cacheDir = Path.Combine(pathBase, cachDirName, GetActualRuntimeStringValue(runtime));
             Directory.CreateDirectory(cacheDir);
 
             var tentacleVersionCacheDir = Path.Combine(cacheDir, version);
             return tentacleVersionCacheDir;
         }
 
+        private static string GetActualRuntimeStringValue(TentacleRuntime runtime)
```
nathanwoctopusdeploy, line 71: nit: drop Actual?

source/Octopus.Tentacle.Tests.Integration/Support/TentacleFetchers/TentacleBinaryCache.cs
```diff
@@ -24,47 +24,63 @@ public TentacleBinaryCache(ITentacleFetcher tentacleFetcher, ILogger logger)
...
+        private static readonly ConcurrentDictionary<(Version, TentacleRuntime), SemaphoreSlim> versionLock = new();
 
-        public async Task<string> GetTentacleVersion(string tmp, Version version, CancellationToken cancellationToken)
+        // TODO: Fun times with Runtimes
+        public async Task<string> GetTentacleVersion(string tmp, Version version, TentacleRuntime runtime, CancellationToken cancellationToken)
         {
-            using var _ = await versionLock.GetOrAdd(version, s => new SemaphoreSlim(1, 1)).LockAsync();
+            using var _ = await versionLock.GetOrAdd((version, runtime), s => new SemaphoreSlim(1, 1)).LockAsync();
+            
+            var runtimeString = runtime.GetStringValue();
```
nathanwoctopusdeploy, line 34: nit: could you achieve this by overriding .ToStrig()

nathanwoctopusdeploy: Or should it be "GetDescription" as I see you also rely on .ToString below so wouldn't want to  change the default behavour

nathanwoctopusdeploy (approved):
Nice. LGTM

source/Octopus.Tentacle.Tests.Integration/Support/TentacleFetchers/TentacleBinaryCache.cs
acodrington: We can't override ToString on an `enum`. We could create a type-safe enum for the runtime, then override `ToString`, but I don't think it's worth the rework.

I did rename the extension method to `GetDescription` though, as that makes the intent clearer. I also got rid of my custom attribute, and just use `System.ComponentModel.DescriptionAttribute` instead.

PR #577, opened 2023-08-30
Change the SQL Docker image for the "Test: Windows Docker Image"

nathanwoctopusdeploy:
# Background

Change the SQL Docker image for the "Test: Windows Docker Image"

This original image gets rebuilt quite frequently and appears to be breaking when starting locally. We don't publish tagged versions so I don't think there is a way to use an older image.

Reverting to a much older docker image that Octopus publish gets us back to a working image.

In the long term, we may want to investigate why the original image has stopped working, but this removed an immediate blocker.

## Before

image

## After 

image

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,9 @@ version: "3"
 
 services:
   db:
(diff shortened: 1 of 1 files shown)


sburmanoctopus (approved):
LGTM

PR #578, opened 2023-08-31
Add TentacleConfigTestCase object

acodrington:
# Background

[sc-56679](https://app.shortcut.com/octopusdeploy/story/56679/add-support-to-tentacle-to-test-net6-tentacle-client-with-a-net48-service-tentacle) involves touching (almost) every Tentacle test to introduce a new 'runtime' argument. To avoid double-handling these tests now and in the future, this PR refactors all relevant tests to use a test case object `TentacleConfigTestCase` which consolidates all necessary Tentacle configuration.

# Results

## Before
```
[Test]
[SyncAndAsyncTestCase(TentacleType.Polling, RpcCallStage.InFlight, RpcCall.FirstCall)]
[SyncAndAsyncTestCase(TentacleType.Polling, RpcCallStage.Connecting, RpcCall.FirstCall)]
[SyncAndAsyncTestCase(TentacleType.Listening, RpcCallStage.InFlight, RpcCall.FirstCall)]
[SyncAndAsyncTestCase(TentacleType.Listening, RpcCallStage.Connecting, RpcCall.FirstCall)]
[SyncAndAsyncTestCase(TentacleType.Polling, RpcCallStage.InFlight, RpcCall.RetryingCall)]
[SyncAndAsyncTestCase(TentacleType.Polling, RpcCallStage.Connecting, RpcCall.RetryingCall)]
[SyncAndAsyncTestCase(TentacleType.Listening, RpcCallStage.InFlight, RpcCall.RetryingCall)]
[SyncAndAsyncTestCase(TentacleType.Listening, RpcCallStage.Connecting, RpcCall.RetryingCall)]
public async Task DuringGetCapabilities_ScriptExecutionCanBeCancelled(TentacleType tentacleType, RpcCallStage rpcCallStage, RpcCall rpcCall, SyncOrAsyncHalibut syncOrAsyncHalibut)
{
    ...
}
```
## After
```
[Test]
[TentacleConfigurations(additionalParameterTypes: new object[] { typeof(RpcCallStages), typeof(RpcCalls) })]
public async Task DuringGetCapabilities_ScriptExecutionCanBeCancelled(TentacleConfigurationTestCase tentacleConfigurationTestCase)
{
    ...
}       
```

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Octopus.Tentacle.Tests.Integration/ClientFileTransferRetriesTimeout.cs
```diff
@@ -22,14 +22,12 @@ namespace Octopus.Tentacle.Tests.Integration
...
-            [Values]TentacleType tentacleType, 
-            [Values]bool stopPortForwarderAfterFirstCall, 
-            [Values]SyncOrAsyncHalibut syncOrAsyncHalibut)
+        [TentacleConfigurations(testStopPortForwarderAfterFirstCall: true)]
+        public async Task WhenRpcRetriesTimeOut_DuringUploadFile_TheRpcCallIsCancelled(TentacleConfigurationTestCase tentacleConfigurationTestCase)
         {
             PortForwarder portForwarder = null!;
-            await using var clientAndTentacle = await new ClientAndTentacleBuilder(tentacleType)
-                .WithAsyncHalibutFeature(syncOrAsyncHalibut.ToAsyncHalibutFeature())
+            await using var clientAndTentacle = await new ClientAndTentacleBuilder(tentacleConfigurationTestCase.TentacleType)
```
LukeButters, line 29: I think it would have made sense to have `ClientAndTentacleBuilder` take a `tentacleConfigurationTestCase` or create an extension method which makes that builder from a `tentacleConfigurationTestCase`.

Since now when you add a new value to `tentacleConfigurationTestCase` you will have to go back to every test and make sure it is used.

acodrington: Yeah, that's a good idea. I'm gonna write it on my sharpening list 👍🏻

source/Octopus.Tentacle.Tests.Integration/Support/TentacleConfigurationsAttribute.cs
```diff
@@ -0,0 +1,121 @@
...
+                ? new List<ScriptsInParallelTestCase?>
+                {
+                    // Scripts with different mutex names can run at the same time.
+                    ScriptsInParallelTestCase.FullIsolationDifferentMutex,
+                    // Scripts with the same mutex name can run at the same time if they both has no isolation.
+                    ScriptsInParallelTestCase.NoIsolationSameMutex
+                }
+                : new List<ScriptsInParallelTestCase?> {null};
+
+            return (from tentacleType in tentacleTypes
```
LukeButters, line 101: This link is nice.

source/Octopus.Tentacle.Tests.Integration/Support/TentacleConfigurationsAttribute.cs
```diff
@@ -0,0 +1,121 @@
...
+                    from stopPortForwarderAfterFirstCall in stopPortForwarderAfterFirstCallValues
+                    from rpcCall in rpcCalls
+                    from rpcCallStage in rpcCallStages
+                    from scriptIsolationLevel in scriptIsolationLevels
+                    from scriptsInParallelTestCase in scriptsInParallelTestCases
+                    select new TentacleConfigurationTestCase(
+                        tentacleType,
+                        halibutType,
+                        version,
+                        stopPortForwarderAfterFirstCall,
```
LukeButters, line 113: I don't think most of these bools belong on here.

We had the same problem, of needing additional params, in halibut. This was solved in `LatestClientAndLatestServiceTestCasesAttribute` using `additionalParameters`.

An alternative is to use `ValuesOfType` (which is defined in halibut). You give a type reference to an `IEnumerable`. ValuesOfType doesn't yet have any support for passing parameters. Although a copy of it in tentacle could.

acodrington: Ah cool. didn't notice the solution in `LatestClientAndLatestService...` attribute, that's most of what I wanted. I don't like that you still have to declare multiple copies of the attribute to get all permutations of an enum etc, but coupled with something like the `ValuesOfType` could make it nicer.

Rather than push back this PR to do the rework, I reckon I put it on my sharpening list and do this when I tackle your other suggestion (i.e. `ClientAndTentacleBuilder` takes the whole `TentacleConfigurationTestCase` object)

acodrington: After some more discussion with @LukeButters, I have refactored the test case attribute class to remove most of bools and config values that were specific to very small sets of tests.

There is now an `additionalParameterTypes` field, which takes an array of `Type` objects. The attribute assumes these types are either enums or implement `IEnumerable`, then generates the necessary test cases to cover all permutations.

source/Octopus.Tentacle.Tests.Integration/ClientFileTransferRetriesTimeout.cs
acodrington: This was actually pretty quick to do with a simple find/replace, so it's been done.

source/Octopus.Tentacle.Tests.Integration/Support/TentacleConfigurationsAttribute.cs
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+
+    static class TentacleConfigurationTestCases
+    {
+        public static IEnumerator GetEnumerator(
+            bool testCommonVersions,
+            bool testCapabilitiesInterestingVersions,
+            bool testNoCapabilitiesServiceVersions,
+            bool testScriptIsolationLevel,
```
sburmanoctopus, line 32: Nit: `testScriptIsolationLevel` different name to `testScriptIsolationLevelVersions` above

source/Octopus.Tentacle.Tests.Integration/Support/TentacleConfigurationsAttribute.cs
```diff
@@ -0,0 +1,131 @@
...
+
+            if (testNoCapabilitiesServiceVersions)
+            {
+                versions.AddRange(new []
+                {
+                    TentacleVersions.v6_3_451_NoCapabilitiesService
+                });
+            }
+
+            var testCases =
```
sburmanoctopus, line 80: If we don't have a version, will this result in no test cases?

I saw in `ClientFileTransferRetriesTimeout` that we do not choose a version. And I guess we shouldn't as it does not want to test versions.

But since all our version booleans are false by default, it means that `versions` will be empty in that case.

I did a little (and very ugly) test in LinqPad to see what would happen if we did a join with an empty list. The following results in 0 results:

```
var x = new[]{"One", "Two", "Three"};
var y = new[] { 1, 2, 3 };
var z = new List<string?>();


var r = 
from a in x
from b in y
from c in z
select new {
	a,b,c
};

r.ToList().Dump();
```
When we add something to `z`, it lights up like a Christmas tree and does what we expect.

I could be missing something, but if feels like maybe we should add `TentacleVersions.Current` if nothing else sets the version?

source/Octopus.Tentacle.Tests.Integration/Support/TentacleConfigurationsAttribute.cs
```diff
@@ -0,0 +1,131 @@
...
+            {
+                var additionalEnum = additionalEnums[i];
+                if (additionalEnum.IsEnum)
+                {
+                    enums.And(Enum.GetValues(additionalEnums[i]));
+                }
+                else if (typeof(IEnumerable).IsAssignableFrom(additionalEnum))
+                {
+                    enums.And(ValuesOf.CreateValues(additionalEnum));
+                }
```
sburmanoctopus, line 112: Nit: Should we throw an exception if we aren't an enum or Enumerable? (or an enumerable with no values, based on previous comment 😁 )

source/Octopus.Tentacle.Tests.Integration/Support/TentacleConfigurationsAttribute.cs
sburmanoctopus: If we haven't already, maybe it'll be worth getting a test output from TeamCity from master, and compare it to a TeamCity build for this PR, and ensure the tests line up?

source/Octopus.Tentacle.Tests.Integration/ClientFileTransferRetriesTimeout.cs
```diff
@@ -153,5 +148,14 @@ public class ClientFileTransferRetriesTimeout : IntegrationTest
             // Ensure we actually waited and retried until the timeout policy kicked in
             duration.Elapsed.Should().BeGreaterOrEqualTo(TimeSpan.FromSeconds(14));
         }
+
+        private class BooleanValues : IEnumerable
```
sburmanoctopus, line 152: Nit: If we are keeping this locally, maybe we should name this how we use it. E.g. `StopPortForwarderAfterFirstCallValues`?

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionCanBeCancelledWhenRetriesAreDisabled.cs
```diff
@@ -134,22 +131,26 @@ await syncOrAsyncHalibut.WhenSync(() => scriptServiceV2.EnsureTentacleIsConnecte
...
+        [TentacleConfigurations(additionalParameterTypes: new object[] {typeof(RpcCallStage)})]
+        public async Task DuringStartScript_ScriptExecutionCanBeCancelled(TentacleConfigurationTestCase tentacleConfigurationTestCase, RpcCallStage rpcCallStage)
         {
+            var expectedFlow1 = rpcCallStage switch
+            {
+                RpcCallStage.Connecting => ExpectedFlow.CancelRpcAndExitImmediately,
+                RpcCallStage.InFlight => ExpectedFlow.AbandonRpcThenCancelScriptThenCompleteScript,
+                _ => throw new ArgumentOutOfRangeException()
+            };
+            var expectedFlow = expectedFlow1;
```
sburmanoctopus, line 143: Nit: Do we need the extra variable `expectedFlow1`?

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionCanBeCancelledWhenRetriesAreDisabled.cs
```diff
@@ -134,22 +131,26 @@ await syncOrAsyncHalibut.WhenSync(() => scriptServiceV2.EnsureTentacleIsConnecte
         }
 
         [Test]
-        [SyncAndAsyncTestCase(TentacleType.Polling, RpcCallStage.Connecting, ExpectedFlow.CancelRpcAndExitImmediately)]
-        [SyncAndAsyncTestCase(TentacleType.Listening, RpcCallStage.Connecting, ExpectedFlow.CancelRpcAndExitImmediately)]
-        [SyncAndAsyncTestCase(TentacleType.Polling, RpcCallStage.InFlight, ExpectedFlow.AbandonRpcThenCancelScriptThenCompleteScript)]
-        [SyncAndAsyncTestCase(TentacleType.Listening, RpcCallStage.InFlight, ExpectedFlow.AbandonRpcThenCancelScriptThenCompleteScript)]
-        public async Task DuringStartScript_ScriptExecutionCanBeCancelled(TentacleType tentacleType, RpcCallStage rpcCallStage, ExpectedFlow expectedFlow, SyncOrAsyncHalibut syncOrAsyncHalibut)
+        [TentacleConfigurations(additionalParameterTypes: new object[] {typeof(RpcCallStage)})]
```
sburmanoctopus, line 134: Nit: I like how we can pass in extra source enumerables. But I can see here in this test how we can't use that to pass in an expected flow.

This made me wonder, did we want to consider splitting `additionalParameterTypes` into two separate optional parameters? E.g. something like `additionalSourceTypes` and `additionalParameters`.

Where the first does what it currently does, and the second simply uses the values it is given.  Then this test could look like:

```
[TentacleConfigurations(additionalParameters: new object[] {RpcCallStage.Connecting, ExpectedFlow.CancelRpcAndExitImmediately})]
[TentacleConfigurations(additionalParameters: new object[] {RpcCallStage.InFlight, ExpectedFlow.AbandonRpcThenCancelScriptThenCompleteScript})]
```

This would save having to do the switch below.

Just a thought 🙂

sburmanoctopus: If we do this, then we can remove the switch statements from this file, and other files. I won't bother tagging them all.

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionIsolationMutex.cs
```diff
@@ -16,37 +16,11 @@ namespace Octopus.Tentacle.Tests.Integration
...
-
         [Test]
-        [TestCaseSource(typeof(WithFullIsolationTestCases))]
-        public async Task ScriptIsolationMutexFull_EnsuresTwoDifferentScriptsDontRunAtTheSameTime(
-            TentacleType tentacleType, 
-            Version? tentacleVersion, 
-            ScriptIsolationLevel levelOfSecondScript,
-            SyncOrAsyncHalibut syncOrAsyncHalibut)
+        [TentacleConfigurations(testScriptIsolationLevelVersions: true, additionalParameterTypes: new object[] { typeof(ScriptIsolationLevel)})]
+        public async Task ScriptIsolationMutexFull_EnsuresTwoDifferentScriptsDontRunAtTheSameTime(TentacleConfigurationTestCase tentacleConfigurationTestCase, ScriptIsolationLevel scriptIsolationLevel)
```
sburmanoctopus, line 21: Nit: should we use the existing name of `levelOfSecondScript`?

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionScriptServiceV1IsNotRetried.cs
```diff
@@ -21,12 +21,10 @@ namespace Octopus.Tentacle.Tests.Integration
     public class ClientScriptExecutionScriptServiceV1IsNotRetried : IntegrationTest
     {
         [Test]
-        [TestCaseSource(typeof(TentacleTypesToTest))]
-        public async Task WhenANetworkFailureOccurs_DuringStartScript_WithATentacleThatOnlySupportsV1ScriptService_TheCallIsNotRetried(TentacleType tentacleType, SyncOrAsyncHalibut syncOrAsyncHalibut)
+        [TentacleConfigurations(testNoCapabilitiesServiceVersions: true)]
```
sburmanoctopus, line 24: `TentacleTypesToTest` is just the regular polling/listening/sync/async mix with no version.

But here, we've chosen `testNoCapabilitiesServiceVersions: true`, which now changes the versions.

What was the reason for that?

Whatever we choose, it's throughout this file.

sburmanoctopus (commented):
Looking good. There are some nits, plus a couple of outstanding questions that might need addressing.

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionScriptServiceV1IsNotRetried.cs
acodrington: Yeah this is intentional.

All these tests used to explicitly call `.WithTentacleVersion(TentacleVersions.v6_3_451_NoCapabilitiesService)`, but now that is handled by the attribute bool and the test case `CreateBuilder` method.

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionCanBeCancelledWhenRetriesAreDisabled.cs
acodrington: We certainly could do this, but we only use the switch statement in a handful of tests, and I'm not sure it's worth the added complexity and rework to build this into the attribute.

Let's leave it as-is for now, and if we need to do this kind of thing more in the future, we can add it then.

sburmanoctopus (approved):
LGTM

PR #580, opened 2023-09-05
Fix integration test builds on previous versions of Windows

acodrington:
# Background

We would like to have full confidence in the reliability of Tentacle on older versions of Windows i.e. Windows Server 2012, 2012 R2, and 2016. As such, this PR makes some small changes to our integration test setup to ensure we get reliable test run results.

There are detailed comments below each code change in the PR, to explain why the particular change is being made.

# Results

Fixes [sc-58883]

## Before
20230906-200043_2v8NzMdjoO

## After
20230906-195539_firefox_DPAwqCTyWy

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.CommonTestUtils/Builders/ScriptBuilder.cs
+++ b/source/Octopus.Tentacle.CommonTestUtils/Builders/ScriptBuilder.cs
@@ -41,7 +41,7 @@ namespace Octopus.Tentacle.Tests.Integration.Util.Builders
 touch '{file}'
 ");
             windowsScript.AppendLine($@"
-New-Item '{file}'
+New-Item -type file '{file}'
 ");
             return this;
         }
--- a/source/Octopus.Tentacle.Tests.Integration/Util/RandomTemporaryFileBuilder.cs
+++ b/source/Octopus.Tentacle.Tests.Integration/Util/RandomTemporaryFileBuilder.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.IO;
+using System.Threading.Tasks;
 
 namespace Octopus.Tentacle.Tests.Integration.Util
 {
@@ -38,7 +39,14 @@ namespace Octopus.Tentacle.Tests.Integration.Util
         {
             if (File.Exists)
             {
-                File.Delete();
+                try
+                {
+                    _ = Task.Run(() => File.Delete());
+                }
+                catch
+                {
+                    // We really don't care
+                }
             }
         }
     }
--- a/source/Octopus.Tentacle.Tests.Integration/Util/RandomTemporaryFileBuilder.cs
+++ b/source/Octopus.Tentacle.Tests.Integration/Util/RandomTemporaryFileBuilder.cs
@@ -45,7 +45,9 @@ namespace Octopus.Tentacle.Tests.Integration.Util
                 }
                 catch
                 {
-                    // We really don't care
+                    // We sometimes have tests failing due to this file being locked during teardown.
+                    // As this is class purely for testing, and we regularly recycle our build agents,
+                    // we are happy with a 'best effort' approach here.
                 }
             }
         }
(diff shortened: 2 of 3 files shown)


source/Octopus.Tentacle.CommonTestUtils/Builders/ScriptBuilder.cs
```diff
@@ -41,7 +41,7 @@ public ScriptBuilder CreateFile(string file)
 touch '{file}'
 ");
             windowsScript.AppendLine($@"
-New-Item '{file}'
+New-Item -type file '{file}'
```
acodrington, line 44: In recent versions of Powershell, the `type` arg is optional and a sensible default is chosen by Powershell.

However, in _older_ versions of Powershell, like on Windows Server 2012, omitting the arg brings up a user prompt to specify the value. Obviously, we can't do this in Tentacle (as Powershell is in non-interactive mode), so Tentacle was failing here.

In the future, it would be helpful for test failures to log the Tentacle output, which would've made finding this one a lot easier.

source/Octopus.Tentacle.Tests.Integration/Util/RandomTemporaryFileBuilder.cs
```diff
@@ -38,7 +39,14 @@ public void Dispose()
...
             {
-                File.Delete();
+                try
+                {
+                    _ = Task.Run(() => File.Delete());
+                }
+                catch
+                {
+                    // We really don't care
+                }
```
acodrington, line 49: We had some tests regularly failing, as the temp file was still in use.

After a code review of the test clean up logic in these tests, everything seemed to be getting cleaned up as expected, but the file is still _sometimes_ in use.

As this is class purely for testing, and we regularly recycle our build agents, we are happy with a 'best effort' approach here.

source/Octopus.Tentacle.Tests.Integration/Util/TcpTentacleHelpers/ListeningResponseMessageTcpKiller.cs
```diff
@@ -40,18 +40,26 @@ public IDataTransferObserver DataTransferObserver()
...
+                // Windows Server 2012, where the max size seems to be ~85.
                 // So anything bigger must be the interesting one.
-                if (pauseConnection && size > 45)
+                //
+                // We increased this value from 45 -> 85 to account for Windows Server 2012,
+                // and all the integration test builds seem happy with it.
+                // May need to be adjusted in the future to account for different OSs,
+                // runtimes, lunar cycles, and/or any other random heuristic.
+                const int assumedControlMessageMaxSize = 85;
+
```
acodrington, line 52: Mmmmmmm... magic numbers!

We had some tests reliably failing on 2012/2012 R2 builds, but only when:
* Tentacle was in listening mode
* Test tries to kill the connection

Turns out almost everything was working as expected, _except_ these tests were killing the connection at _very slightly the wrong time_ due to the control messages being a different size on 2012. Killing the connection at this slightly-wrong point somehow resulted in the connection recovering automatically and then completing the RPC call successfully. Fantastic, robust behaviour! Except, of course, we are trying to simulate a flaky network and ensure RPC retries are doing what they are supposed to do, so the tests fail.

We discussed increasing this magic number for *only* Windows Server 2012/2012 R2 integration test builds, but all the other integration test runs are happy with this larger number, so it seems acceptable to change it across the board.

source/Octopus.Tentacle.Tests.Integration/Util/RandomTemporaryFileBuilder.cs
sburmanoctopus: Nit: This is a great explanation. It would be awesome to have an abbreviated version as the comment in code, as I think it would be a better comment.

sburmanoctopus (approved):
LGTM. Excellent work getting the tests green! 😁

PR #582, opened 2023-09-06
Adding extra testing for both net48 and net6.0 installers

sburmanoctopus:
[sc-58906]

# Background

We want to ensure our existing (net48) and new (net6.0) installers work with some automated tests.

# Results

## Before
We have one existing automated test for our installers. It does actually cover both installers already. But all it tested was that the install location directory excluded a certain permission.

## After
We now also test that `Tentacle.exe` was installed, and that it can be run (just a simple command, `help`).

**Locally**
image

**TeamCity**
image

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/build/Build.Tests.cs
+++ b/build/Build.Tests.cs
@@ -1,6 +1,7 @@
 ﻿// ReSharper disable RedundantUsingDirective
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Security.AccessControl;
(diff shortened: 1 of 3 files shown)


build/Build.Tests.cs
```diff
@@ -194,10 +129,134 @@ bool DoesSidHaveRightsToDirectory(string directory, WellKnownSidType sid, params
...
+                    TestWindowsInstaller(installer);
                 }
             });
         });
+
+    string GetTestName(AbsolutePath installerPath) => Path.GetFileName(installerPath).Replace(".msi", "");
+        
+    void TestWindowsInstaller(AbsolutePath installerPath)
+    {
+        Log.Information($"\n--------------------------------------\nTesting Installer {GetTestName(installerPath)}\n--------------------------------------");
```
sburmanoctopus, line 141: Having this loud and proud was helpful during testing, so I left it. But am happy to dumb it down.

acodrington: I like the loud and proud :)

acodrington (approved):
Looks good. General refactor into nicely named methods helps readability too, nice job!

PR #583, opened 2023-09-06
Fix bundling unwanted files in MSI

sburmanoctopus:
[sc-59022]

# Background

We found that the Tentacle Manager was being shipped in the net6.0 MSI, when is shouldn't be.

# Results
Related to https://github.com/OctopusDeploy/OctopusTentacle/issues/584

## Before

We generate the files to include in Wix using Heat.exe. This exe then overrides the `Tentacle.Generated.wxs` file with what should be installed.

Before we run Heat.exe, we copy all the files we need for the MSI being built into the `_build\Installer` folder, and then tell Heat.exe to include everything in there.

But we don't clear that folder between MSI builds. So any file from a previous MSI build will be included in the next MSI (unless it's overridden, which most of them are).

## After

We now clear the `_build\Installer` folder so that the Tentacle Manager files from the previous MSI build are not included in the net6.0 MSIs anymore.

Below is output from a test installation. Note we order by type, and that there is no `Octopus.Manager.Tentacle.exe`
image

Also, note the file size changes before and after.
**Before**
image

**After**
image

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,6 +49,7 @@ tools
 
 _artifacts
 _build
+_test
 
 .DS_Store
 .env
--- a/build/Build.Pack.cs
+++ b/build/Build.Pack.cs
@@ -158,6 +158,7 @@ partial class Build
             {
                 var installerDirectory = BuildDirectory / "Installer";
                 FileSystemTasks.EnsureExistingDirectory(installerDirectory);
+                FileSystemTasks.EnsureCleanDirectory(installerDirectory);
                 
                 if (framework != NetCore)
                 {
--- a/.nuke/build.schema.json
+++ b/.nuke/build.schema.json
@@ -123,7 +123,7 @@
               "TestLinuxPackages",
               "TestOsx",
               "TestWindows",
-              "TestWindowsInstallerPermissions"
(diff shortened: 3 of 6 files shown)


.gitignore
```diff
@@ -49,6 +49,7 @@ tools
 
 _artifacts
 _build
+_test
```
sburmanoctopus, line 52: When tests clean up after themselves, we don't need this. But when they don't, it's a Git mess

build/Build.Pack.cs
```diff
@@ -158,6 +158,7 @@ void PackWindowsInstallers(MSBuildTargetPlatform platform, AbsolutePath wixNuget
             {
                 var installerDirectory = BuildDirectory / "Installer";
                 FileSystemTasks.EnsureExistingDirectory(installerDirectory);
+                FileSystemTasks.EnsureCleanDirectory(installerDirectory);
```
nathanwoctopusdeploy, line 161: Good spot

nathanwoctopusdeploy (approved):

PR #590, opened 2023-09-11
Ensure TestPrincipalUser is thread safe

acodrington:
# Background

`SilentProcessRunnerFixture` appears to be a [flaky test fixture](https://build.octopushq.com/buildConfiguration/TeamFireAndMotion_OctopusTentacle_IntegrationTestNet60Windows2012/9511661?expandPull+Request+Details=true&expandBuildProblemsSection=true&hideProblemsFromDependencies=false&expandBuildTestsSection=true&hideTestsFromDependencies=false&expandBuildChangesSection=true), with various tests occasionally failing with the following error:

```
System.DirectoryServices.AccountManagement.PrincipalExistsException : The account already exists.

  ----> System.Runtime.InteropServices.COMException : The account already exists.
   at System.DirectoryServices.AccountManagement.SDSUtils.ApplyChangesToDirectory(Principal p, StoreCtx storeCtx, GroupMembershipUpdater updateGroupMembership, NetCred credentials, AuthenticationTypes authTypes)
   at System.DirectoryServices.AccountManagement.SDSUtils.InsertPrincipal(Principal p, StoreCtx storeCtx, GroupMembershipUpdater updateGroupMembership, NetCred credentials, AuthenticationTypes authTypes, Boolean needToSetPassword)
   at System.DirectoryServices.AccountManagement.SAMStoreCtx.Insert(Principal p)
   at System.DirectoryServices.AccountManagement.Principal.Save()
   at Octopus.Tentacle.Tests.Integration.Util.TestUserPrincipal..ctor(String username, String password) in C:\BuildAgent\work\639265b01610d682\source\Octopus.Tentacle.Tests.Integration\Util\TestUserPrincipal.cs:line 32
   at Octopus.Tentacle.Tests.Integration.Util.SilentProcessRunnerFixture.SetUpLocal() in C:\BuildAgent\work\639265b01610d682\source\Octopus.Tentacle.Tests.Integration\Util\SilentProcessRunnerFixture.cs:line 27
--COMException
   at System.DirectoryServices.DirectoryEntry.CommitChanges()
   at System.DirectoryServices.AccountManagement.SDSUtils.ApplyChangesToDirectory(Principal p, StoreCtx storeCtx, GroupMembershipUpdater updateGroupMembership, NetCred credentials, AuthenticationTypes authTypes)
```

[...]

PR #591, opened 2023-09-11
Fixing tests for 2012

sburmanoctopus:
[sc-59228]
[sc-59220]

# Background

To support Windows 2012 and Windows 2012R2, we have a build chain that tests Tentacle against this OS. But there are some tests that fail reliably.

# Results

## Before

There were consistent [test failures for 2012] (https://build.octopushq.com/buildConfiguration/TeamFireAndMotion_OctopusTentacle_IntegrationTestNet48onWindows2012/9512172?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true&expandBuildTestsSection=true):
image

and [in 2012R2](https://build.octopushq.com/buildConfiguration/TeamFireAndMotion_OctopusTentacle_IntegrationTestNet48onWindows2012r2/9512177?hideProblemsFromDependencies=false&expandBuildTestsSection=true&hideTestsFromDependencies=false&expandBuildChangesSection=true&expandBuildProblemsSection=true):
image

The causes for these were due to 2 fundamental issues:
- A test concurrency issue that was causing a `The account already exists`
- A username inconsistency issue that was causing the user to be `SYSTEM` where it was actually something like `WIN-VTD8GECL0HO$`

## After

Thankfully, the concurrency issue was fixed in [another PR](https://github.com/OctopusDeploy/OctopusTentacle/pull/590) (yay 😁)

The username logic was fixed in this PR. It looks like this was a known issue, with a known fix. So we simply centralised the fix, and applied it to the failing tests.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Tests.Integration/Support/TestEnvironmentHelper.cs
+++ b/source/Octopus.Tentacle.Tests.Integration/Support/TestEnvironmentHelper.cs
@@ -1,6 +1,5 @@
 ﻿using System.Security.Principal;
 using System;
-using System.Linq;
 using Octopus.Tentacle.Util;
 
 namespace Octopus.Tentacle.Tests.Integration.Support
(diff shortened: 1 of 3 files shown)


source/Octopus.Tentacle.Tests.Integration/Support/TestEnvironmentHelper.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System.Security.Principal;
+using System;
+using System.Linq;
+using Octopus.Tentacle.Util;
+
+namespace Octopus.Tentacle.Tests.Integration.Support
+{
+    public static class TestEnvironmentHelper
```
sburmanoctopus, line 8: The distinction between these is that 
- `CurrentFullUserName` will result in `WORKGROUP\WIN-VTD8GECL0HO$`
- `CurrentUserName` will result in `WIN-VTD8GECL0HO$`

But annoyingly, `Environment.UserName` (i.e. when not Windows) will just be the name (e.g. `WIN-VTD8GECL0HO$`). So I didn't name these properties with things like `CurrentDomainUserName` or anything like that 😑 

I would love some input into a name that helps distinguish these.

Otherwise, I could comment them?

acodrington: I think it's going to be hard to think of a succint name that conveys the different Linux vs Windows behaviour. I'd say a good comment on them would be fine.

acodrington (approved):

PR #594, opened 2023-09-14
Add Integration Test build target for net 6.0 client <-> net48 service

acodrington:
# Background

[Recent changes to test more combinations of server and client versions](https://github.com/OctopusDeploy/OctopusTentacle/pull/576) (i.e. net6.0 client <-> net48 service) have made our Windows (net6.0) Integration test builds take longer than we'd like.

This PR adds another build target called `TestIntegrationNet60ClientNet48Service`, which in combination with changes to the build chain in TeamCity, will allow us to split our integration test builds into separate test runs, allowing quicker and clearer feedback from the builds system.

# Results

* Existing Windows integration test builds (current, 2016, 2012 R2, 2012) take roughly the same time they have been
* New (!) Windows integration test builds for current, 2016, 2012 R2, 2012 versions i.e. `Integration Test: net6.0 on Windows (net48 service)`

## Before / After

20230914-174557_firefox_pkE6CK1OJc

20230914-174953_firefox_9iTnGNQqL7

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/build/Build.Tests.cs
+++ b/build/Build.Tests.cs
@@ -31,11 +31,7 @@ partial class Build
 
     [PublicAPI]
     Target TestIntegration => _ => _
-        .Executes(() => RunIntegrationTests(TestFramework, TestRuntime, "TestCategory!=\"Net60ClientNet48Service\""));
-
-    [PublicAPI]
-    Target TestIntegrationNet60ClientNet48Service => _ => _
-        .Executes(() => RunIntegrationTests(TestFramework, TestRuntime,"TestCategory=\"Net60ClientNet48Service\""));
+        .Executes(() => RunIntegrationTests(TestFramework, TestRuntime, Filter));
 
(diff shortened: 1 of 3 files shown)


nathanwoctopusdeploy:
Have you taken a look at how this was implemented in halibut to split the net48 build chain up. It's the same approach but more generic so doesn't require a nuke config each time we want to split stuff

acodrington:
> Have you taken a look at how this was implemented in halibut to split the net48 build chain up. It's the same approach but more generic so doesn't require a nuke config each time we want to split stuff

Oh nice! I wasn't too concerned about making it more generic for this one time (figured we could do it later if needed), but I'll make some changes as the solution is ready-to-go! 👍🏻

nathanwoctopusdeploy (approved):

PR #604, opened 2023-09-19
Removing usage of CancellationToken.Register

sburmanoctopus:
[sc-58580]

# Background

During Halibut testing, [it was found](https://github.com/OctopusDeploy/Halibut/pull/481) that using `CancellationToken.Register` could cause unwanted memory leaks, and odd patterns when analysing memory.

Therefore, we decided to remove the usage of `CancellationToken.Register`.

# Results
Related to OctopusDeploy/Issues#8229

## Before

We would use `CancellationToken.Register` to see if a task was abandoned.

## After

We now achieve the same outcome using `Task.Delay` and `CancellationTokenSource`.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Octopus.Tentacle.Client/Retries/TaskExtensionMethods.cs
```diff
@@ -23,5 +25,43 @@ public static void IgnoreUnobservedExceptions(this Task task)
...
                 TaskContinuationOptions.OnlyOnFaulted | TaskContinuationOptions.ExecuteSynchronously);
         }
+
+        public static async Task<TaskCompletionResult> WaitTillCompletion(this Task actionTask, TimeSpan abandonAfter, CancellationToken cancellationToken)
+        {
+            var actionTaskCompleted = await actionTask.WaitTillCompletedOrCancelled(cancellationToken);
+
+            if (actionTaskCompleted) return TaskCompletionResult.Completed;
+
+            actionTaskCompleted = await actionTask.WaitTillCompletedOrTimeout(abandonAfter).ConfigureAwait(false);
```
nathanwoctopusdeploy, line 35: nit ConfigureAwait is sometimes used. Should we go consistent and use it or don't

source/Octopus.Tentacle.Client/Retries/TaskExtensionMethods.cs
```diff
@@ -23,5 +25,43 @@ public static void IgnoreUnobservedExceptions(this Task task)
...
+
+            if (actionTaskCompleted) return TaskCompletionResult.Completed;
+
+            actionTask.IgnoreUnobservedExceptions();
+            return TaskCompletionResult.Abandoned;
+
```
nathanwoctopusdeploy, line 41: nit: space

nathanwoctopusdeploy (approved):
LGTM

source/Octopus.Tentacle.Client/Retries/TaskExtensionMethods.cs
sburmanoctopus: Removed

source/Octopus.Tentacle.Client/Retries/TaskExtensionMethods.cs
sburmanoctopus: Yeah, I'm still not 100% sure what our policy on this is.

I think that we do not use it, right?

Either way, I've removed `ConfigureAwait` from this file only.

nathanwoctopusdeploy: Yeh I think we went with we don't need it

PR #607, opened 2023-09-21
Removing TentacleRuntime.Default, replacing it with a calculated value

sburmanoctopus:
[sc-60184]

# Background

We recently [added support for testing net6.0 client with net48 service](https://github.com/OctopusDeploy/OctopusTentacle/pull/576).

This seemed to drastically increase the number of tests run in TeamCity.
image

# Results

## Before
TeamCity would run the tests on different test chains, for different platforms, and aggregate them together. 

As seen here, before we introduced the net6.0/net48 test cases, it ran each of these on 7 different build chains:
image

`TentacleConfigurationsAttribute` was then altered to try different runtimes:
```
#if NETFRAMEWORK
            var runtimes = new List<TentacleRuntime> { TentacleRuntime.Default };
#else
            var runtimes = PlatformDetection.IsRunningOnWindows
                ? new List<TentacleRuntime> {TentacleRuntime.DotNet6, TentacleRuntime.Framework48}
                : new List<TentacleRuntime> {TentacleRuntime.Default};
#endif
```

So if we were a Windows net6.0 build chain, we would now have twice as many test cases.

As seen below, this appeared to be what was happening:
image

But when we use the 'Default' runtime, the test case does not show what the runtime is.

So when TeamCity groups by name, it looks like there are far more test cases now. Whereas really, it's just not grouping them together.

## After
So to make this report more accurately in TeamCity, we made the test case always report the framework in it's name.

While looking into this, it made us wonder if 'Default' was a legitimate `TentacleRuntime`. 

So instead, we decided to remove `Default` as a `TentacleRuntime`, and replace it with a field based on the current runtime. This also helped to simplify the code that would switch on the runtime.

As seen here, the values are now aggregated.
image

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Tests.Integration/Support/ClientAndTentacleBuilder.cs
+++ b/source/Octopus.Tentacle.Tests.Integration/Support/ClientAndTentacleBuilder.cs
@@ -25,7 +25,7 @@ namespace Octopus.Tentacle.Tests.Integration.Support
         IScriptObserverBackoffStrategy scriptObserverBackoffStrategy = new DefaultScriptObserverBackoffStrategy();
         public readonly TentacleType TentacleType;
         Version? tentacleVersion;
-        private TentacleRuntime tentacleRuntime = TentacleRuntime.Default;
+        private TentacleRuntime tentacleRuntime = DefaultTentacleRuntime.Value;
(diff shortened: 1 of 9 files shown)


acodrington (approved):
Nice one, definitely makes things simpler! 👍🏻

PR #609, opened 2023-09-22
If tentacle runtime matches the current runtime then do not show tentacle runtime

sburmanoctopus:
[sc-60184]

# Background

It was decided that there was no need to separate tests that were using the same runtime for server and tentacle.

# Results

## Before

The grouping we tried was not that helpful.
image

## After
We now remove the tentacle runtime name if the current runtime matches (screenshot is 2 short, as it was harvested while tests were running)
image

I also noticed that `RuntimeDetection` was in the `Octopus.Tentacle` assembly, even though it wasn't used there (i.e. it was just used in tests). So that's now moved as well.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Octopus.Tentacle.Tests.Integration/Support/TentacleConfigurationTestCase.cs
```diff
@@ -47,7 +47,12 @@ public override string ToString()
             string version = Version?.ToString() ?? "Latest";
             builder.Append($"{version},");
             builder.Append($"{SyncOrAsyncHalibut}");
-            builder.Append($",{TentacleRuntime.GetDescription()}");
+
+            var tentacleRuntimeDescription = TentacleRuntime.GetDescription();
+            if (tentacleRuntimeDescription != RuntimeDetection.GetCurrentRuntime())
+            {
+                builder.Append($",{tentacleRuntimeDescription}");
+            }
```
acodrington, line 55: It'd be great if we could be more descriptive here e.g. `net6.0client-net48service`, rather than just the framework of the service. The string could be constructed by using the current runtime (there's a helper for this somewhere) and the Tentacle runtime.

sburmanoctopus: Good idea.

I nearly went with the suggested style of `net6.0client-net48service`, but remembered how you and Luke didn't like lengthy test names in Rider.

So I went for a more compact version of C and S. If you prefer it to be more verbose, then I'm cool with that. But I thought you might like it shorter 😁 

image

nathanwoctopusdeploy (approved):

PR #615, opened 2023-09-26
Cleaning up old workspaces

sburmanoctopus:
# Background

Tentacles do not clean up old workspaces if something stops them from doing it naturally. This can lead to a build up of files and directories over time.

# Results

Fixes https://github.com/OctopusDeploy/OctopusTentacle/issues/619

## Before
We had no way of cleaning up old workspaces if Octopus Server could not send the `CompleteScript` command.

## After
We now have a separate background task that periodically looks for old workspaces and deletes them.

Running the task makes sense to only do when we are running the agent. So we only start it during the `RunAgentCommand`.

Services are kept within their own separate scope inside `AutofacServiceFactory`. But we needed to access them in our clean up to find what is currently running. Since we want to minimize change to `ScriptService`, we instead exposed the two script services via a new interface (`IServiceRegistration`)

`ScriptTicket` was never used in `ScriptState`. This was removed during an initial version of this PR. Although that change it's not required anymore, it was left in as the change made sense.

### Testing
**Pointing at an old instance which had 4 orphaned workspaces, it cleaned them all up:**
image
image

**Forcing a failure mid way so that we created an orphaned workspace, the task left it alone:**
image
image

**While running, the new folder was left alone:**
image

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Client/Scripts/ScriptExecutionOrchestrator.cs
+++ b/source/Octopus.Tentacle.Client/Scripts/ScriptExecutionOrchestrator.cs
@@ -461,9 +461,15 @@ namespace Octopus.Tentacle.Client.Scripts
                             clientOperationMetricsBuilder,
                             CancellationToken.None);
 
-                    await actionTask.WaitTillCompletion(onCancellationAbandonCompleteScriptAfter, scriptExecutionCancellationToken);
+                    var actionTaskCompletionResult = await actionTask.WaitTillCompletion(onCancellationAbandonCompleteScriptAfter, scriptExecutionCancellationToken);
+                    if (actionTaskCompletionResult == TaskCompletionResult.Abandoned)
+                    {
+                        throw new OperationAbandonedException(onCancellationAbandonCompleteScriptAfter);
+                    }
+
+                    await actionTask;
                 }
-                catch (Exception ex) when (ex is HalibutClientException or OperationCanceledException)
+                catch (Exception ex) when (ex is HalibutClientException or OperationCanceledException or OperationAbandonedException)
                 {
                     logger.Warn("Failed to cleanup the script working directory on Tentacle");
                     logger.Verbose(ex);
--- a/source/Octopus.Tentacle.Contracts/ScriptTicket.cs
+++ b/source/Octopus.Tentacle.Contracts/ScriptTicket.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Threading;
 
 namespace Octopus.Tentacle.Contracts
 {
@@ -18,7 +17,7 @@ namespace Octopus.Tentacle.Contracts
                 return false;
             if (ReferenceEquals(this, other))
                 return true;
-            return string.Equals(TaskId, other.TaskId);
+            return string.Equals(TaskId, other.TaskId, StringComparison.OrdinalIgnoreCase);
         }
 
         public override bool Equals(object? obj)
@@ -33,7 +32,7 @@ namespace Octopus.Tentacle.Contracts
         }
 
         public override int GetHashCode()
-            => TaskId != null ? TaskId.GetHashCode() : 0;
+            => TaskId != null ? StringComparer.OrdinalIgnoreCase.GetHashCode(TaskId) : 0;
 
         public static bool operator ==(ScriptTicket left, ScriptTicket right)
             => Equals(left, right);
--- a/source/Octopus.Tentacle.Contracts/ScriptTicket.cs
+++ b/source/Octopus.Tentacle.Contracts/ScriptTicket.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Threading;
 
 namespace Octopus.Tentacle.Contracts
 {
@@ -17,7 +18,7 @@ namespace Octopus.Tentacle.Contracts
                 return false;
             if (ReferenceEquals(this, other))
                 return true;
-            return string.Equals(TaskId, other.TaskId, StringComparison.OrdinalIgnoreCase);
+            return string.Equals(TaskId, other.TaskId);
         }
 
         public override bool Equals(object? obj)
@@ -32,7 +33,7 @@ namespace Octopus.Tentacle.Contracts
         }
 
         public override int GetHashCode()
-            => TaskId != null ? StringComparer.OrdinalIgnoreCase.GetHashCode(TaskId) : 0;
+            => TaskId != null ? TaskId.GetHashCode() : 0;
 
         public static bool operator ==(ScriptTicket left, ScriptTicket right)
             => Equals(left, right);
--- a/source/Octopus.Tentacle.Tests.Integration/Support/TentacleBuilder.cs
+++ b/source/Octopus.Tentacle.Tests.Integration/Support/TentacleBuilder.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
@@ -13,13 +14,39 @@ using Serilog;
 
 namespace Octopus.Tentacle.Tests.Integration.Support
 {
-    public abstract class TentacleBuilder<T> where T : TentacleBuilder<T>
+    public interface ITentacleBuilder
+    {
+        ITentacleBuilder WithRunTentacleEnvironmentVariable<TValue>(string environmentVariable, TValue value);
+        ITentacleBuilder WithHomeDirectory(TemporaryDirectory homeDirectory);
+    }
+
+    public abstract class TentacleBuilder<T> : ITentacleBuilder
+        where T : TentacleBuilder<T>
     {
         protected string? ServerThumbprint;
         protected string? TentacleExePath;
         protected string CertificatePfxPath = Certificates.TentaclePfxPath;
         protected string TentacleThumbprint = Certificates.TentaclePublicThumbprint;
-        private readonly Regex listeningPortRegex = new Regex(@"^listen:\/\/.+:(\d+)\/");
+        
+        readonly Regex listeningPortRegex = new (@"^listen:\/\/.+:(\d+)\/");
+        readonly Dictionary<string, string> runTentacleEnvironmentVariables = new();
+
+        TemporaryDirectory? homeDirectory;
+
+        protected TemporaryDirectory HomeDirectory
+        {
+            get
+            {
+                homeDirectory ??= new TemporaryDirectory();
+                return homeDirectory;
+            }
+        }
+
+        public ITentacleBuilder WithHomeDirectory(TemporaryDirectory homeDirectory)
+        {
+            this.homeDirectory = homeDirectory;
+            return this;
+        }
 
         public T WithCertificate(string certificatePfxPath, string tentacleThumbprint)
         {
@@ -34,6 +61,13 @@ namespace Octopus.Tentacle.Tests.Integration.Support
             TentacleExePath = tentacleExe;
 
             return (T)this;
+        }
+
+        public ITentacleBuilder WithRunTentacleEnvironmentVariable<TValue>(string environmentVariable, TValue value)
+        {
+            runTentacleEnvironmentVariables[environmentVariable] = value?.ToString();
+
+            return this;
         }
 
         protected void WithWritableTentacleConfiguration(string configFilePath, Action<IWritableTentacleConfiguration> action)
@@ -81,11 +115,11 @@ namespace Octopus.Tentacle.Tests.Integration.Support
             }
         }
 
-        protected async Task<(Task task, Uri serviceUri)> RunTentacle(
+        async Task<(Task task, Uri serviceUri)> RunTentacle(
             Uri? serviceUri,
             string tentacleExe,
             string instanceName,
-            TemporaryDirectory tempDirectory,
+            TemporaryDirectory tempDirectory, 
             CancellationToken cancellationToken)
         {
             var hasTentacleStarted = new ManualResetEventSlim();
@@ -96,7 +130,10 @@ namespace Octopus.Tentacle.Tests.Integration.Support
             {
                 try
                 {
-                    await RunTentacleCommandOutOfProcess(tentacleExe, new[] {"agent", $"--instance={instanceName}", "--noninteractive"}, tempDirectory,
+                    await RunTentacleCommandOutOfProcess(
+                        tentacleExe, 
+                        new[] {"agent", $"--instance={instanceName}", "--noninteractive"}, 
+                        tempDirectory,
                         s =>
                         {
                             if (s.Contains("Listener started"))
@@ -107,7 +144,9 @@ namespace Octopus.Tentacle.Tests.Integration.Support
                             {
                                 hasTentacleStarted.Set();
                             }
-                        }, cancellationToken);
+                        },
+                        runTentacleEnvironmentVariables, 
+                        cancellationToken);
                 }
                 catch (Exception e)
                 {
@@ -169,20 +208,26 @@ namespace Octopus.Tentacle.Tests.Integration.Support
 
         internal string InstanceNameGenerator()
         {
-            return $"TentacleIT-{Guid.NewGuid().ToString("N")}";
+            return $"TentacleIT-{Guid.NewGuid():N}";
         }
 
-        private async Task RunTentacleCommand(string tentacleExe, string[] args, TemporaryDirectory tmp, CancellationToken cancellationToken)
+        async Task RunTentacleCommand(string tentacleExe, string[] args, TemporaryDirectory tmp, CancellationToken cancellationToken)
         {
-            await RunTentacleCommandOutOfProcess(tentacleExe, args, tmp, s =>
-            {
-            }, cancellationToken);
+            await RunTentacleCommandOutOfProcess(
+                tentacleExe,
+                args, 
+                tmp, 
+                _ => { }, 
+                new Dictionary<string, string?>(), 
+                cancellationToken);
         }
 
-        private async Task RunTentacleCommandOutOfProcess(string tentacleExe,
+        async Task RunTentacleCommandOutOfProcess(
+            string tentacleExe,
             string[] args,
             TemporaryDirectory tmp,
-            Action<string> commandOutput,
+            Action<string> commandOutput, 
+            IReadOnlyDictionary<string, string> environmentVariables,
             CancellationToken cancellationToken)
         {
             var log = new SerilogLoggerBuilder().Build().ForContext<RunningTentacle>();
@@ -199,6 +244,7 @@ namespace Octopus.Tentacle.Tests.Integration.Support
                 var commandResult = await RetryHelper.RetryAsync<CommandResult, CommandExecutionException>(
                     () => Cli.Wrap(tentacleExe)
                         .WithArguments(args)
+                        .WithEnvironmentVariables(environmentVariables)
                         .WithWorkingDirectory(tmp.DirectoryPath)
                         .WithStandardOutputPipe(PipeTarget.ToDelegate(ProcessLogs))
                         .WithStandardErrorPipe(PipeTarget.ToDelegate(ProcessLogs))
--- a/source/Octopus.Tentacle.Tests.Integration/WorkspaceCleanerTests.cs
+++ b/source/Octopus.Tentacle.Tests.Integration/WorkspaceCleanerTests.cs
@@ -1,0 +1,183 @@
+using System;
+using System.IO;
+using System.Threading.Tasks;
+using FluentAssertions;
+using NUnit.Framework;
+using Octopus.Tentacle.CommonTestUtils.Builders;
+using Octopus.Tentacle.Maintenance;
+using Octopus.Tentacle.Scripts;
+using Octopus.Tentacle.Tests.Integration.Support;
+using Octopus.Tentacle.Tests.Integration.Util;
+using Octopus.Tentacle.Tests.Integration.Util.Builders;
+using Octopus.Tentacle.Tests.Integration.Util.Builders.Decorators;
+
+namespace Octopus.Tentacle.Tests.Integration
+{
+    [IntegrationTestTimeout]
+    public class WorkspaceCleanerTests : IntegrationTest
+    {
+        [Test]
+        [TentacleConfigurations(testDefaultTentacleRuntimeOnly: true)]
+        public async Task WhenScriptServiceIsRunning_ThenWorkspaceIsNotDeleted(TentacleConfigurationTestCase tentacleConfigurationTestCase)
+        {
+            var cleanerDelay = TimeSpan.FromMilliseconds(500);
+            var deleteWorkspacesOlderThan = TimeSpan.FromMilliseconds(500);
+
+            var startScriptCommand = new StartScriptCommandV2Builder().WithScriptBody(b => b.Print("Hello")).Build();
+
+            var workspaceDirectory = string.Empty;
+            await using var clientAndTentacle = await tentacleConfigurationTestCase.CreateBuilder()
+                .WithTentacle(b =>
+                {
+                    b.WithRunTentacleEnvironmentVariable(WorkspaceCleanerConfiguration.CleanerDelayEnvironmentVariableName, cleanerDelay)
+                        .WithRunTentacleEnvironmentVariable(WorkspaceCleanerConfiguration.DeleteWorkspacesOlderThanTimeSpanEnvironmentVariableName, deleteWorkspacesOlderThan);
+                })
+                .WithTentacleServiceDecorator(new TentacleServiceDecoratorBuilder()
+                    .DecorateScriptServiceV2With(d =>
+                        d.BeforeCompleteScript(async (_, _) =>
+                            {
+                                Directory.Exists(workspaceDirectory).Should().BeTrue("Workspace should exist before we complete");
+
+                                await Task.Delay(2 * 500, CancellationToken);
+
+                                Directory.Exists(workspaceDirectory).Should().BeTrue("Workspace should not have been cleaned up");
+                            })
+                            .Build())
+                    .Build())
+                .Build(CancellationToken);
+
+            workspaceDirectory = GetWorkspaceDirectoryPath(clientAndTentacle.RunningTentacle.HomeDirectory, startScriptCommand.ScriptTicket.TaskId);
+
+            await clientAndTentacle.TentacleClient.ExecuteScript(startScriptCommand, CancellationToken, null, new InMemoryLog());
+            
+            Directory.Exists(workspaceDirectory).Should().BeFalse("Workspace should be naturally cleaned up after completion");
+        }
+        
+        [Test]
+        [TentacleConfigurations(testDefaultTentacleRuntimeOnly: true)]
+        public async Task WhenTentacleClientCrashesAndTentacleNeverGotToldToComplete_ThenWorkspaceIsConsideredRunning_AndWillNotBeDeleted(TentacleConfigurationTestCase tentacleConfigurationTestCase)
+        {
+            var cleanerDelay = TimeSpan.FromMilliseconds(500);
+            var deleteWorkspacesOlderThan = TimeSpan.FromMilliseconds(500);
+
+            var startScriptCommand = new StartScriptCommandV2Builder().WithScriptBody(b => b.Print("Hello")).Build();
+
+            await using var clientAndTentacle = await tentacleConfigurationTestCase.CreateBuilder()
+                .WithTentacle(b =>
+                {
+                    b.WithRunTentacleEnvironmentVariable(WorkspaceCleanerConfiguration.CleanerDelayEnvironmentVariableName, cleanerDelay)
+                        .WithRunTentacleEnvironmentVariable(WorkspaceCleanerConfiguration.DeleteWorkspacesOlderThanTimeSpanEnvironmentVariableName, deleteWorkspacesOlderThan);
+                })
+                .WithTentacleServiceDecorator(new TentacleServiceDecoratorBuilder()
+                    .DecorateScriptServiceV2With(d =>
+                        d.BeforeCompleteScript((_, _) => throw new NotImplementedException("Force failure to simulate tentacle client crashing, and ensure we do not completion"))
+                        .Build())
+                    .Build())
+                .Build(CancellationToken);
+            
+            await AssertionExtensions
+                .Should(() => clientAndTentacle.TentacleClient.ExecuteScript(startScriptCommand, CancellationToken, null, new InMemoryLog()))
+                .ThrowAsync<NotImplementedException>();
+
+            await Task.Delay(1000, CancellationToken);
+
+            var workspaceDirectory = GetWorkspaceDirectoryPath(clientAndTentacle.RunningTentacle.HomeDirectory, startScriptCommand.ScriptTicket.TaskId);
+
+            Directory.Exists(workspaceDirectory).Should().BeTrue();
+        }
+
+        [Test]
+        [TentacleConfigurations(testDefaultTentacleRuntimeOnly: true)]
+        public async Task WhenTentacleStarts_WithWorkspaceOlderThanThreshold_ThenWorkspaceIsDeleted(TentacleConfigurationTestCase tentacleConfigurationTestCase)
+        {
+            var cleanerDelay = TimeSpan.FromMilliseconds(500);
+            var deleteWorkspacesOlderThan = TimeSpan.FromMilliseconds(500);
+
+            var existingHomeDirectory = new TemporaryDirectory();
+            var existingOrphanedWorkspaceDirectory = GivenExistingOrphanedWorkspaceExists(existingHomeDirectory);
+            File.WriteAllText(ScriptWorkspace.GetLogFilePath(existingOrphanedWorkspaceDirectory), "Existing log file");
+
+            await Task.Delay(1000, CancellationToken);
+
+            await using var clientAndTentacle = await tentacleConfigurationTestCase.CreateBuilder()
+                .WithTentacle(b =>
+                {
+                    b.WithHomeDirectory(existingHomeDirectory)
+                        .WithRunTentacleEnvironmentVariable(WorkspaceCleanerConfiguration.CleanerDelayEnvironmentVariableName, cleanerDelay)
+                        .WithRunTentacleEnvironmentVariable(WorkspaceCleanerConfiguration.DeleteWorkspacesOlderThanTimeSpanEnvironmentVariableName, deleteWorkspacesOlderThan);
+                })
+                .Build(CancellationToken);
+
+            await Wait.For(() => !Directory.Exists(existingOrphanedWorkspaceDirectory), CancellationToken);
+        }
+
+        [Test]
+        [TentacleConfigurations(testDefaultTentacleRuntimeOnly: true)]
+        public async Task WhenTentacleStarts_WithWorkspaceYoungerThanThreshold_ThenWorkspaceIsLeftAlone(TentacleConfigurationTestCase tentacleConfigurationTestCase)
+        {
+            var cleanerDelay = TimeSpan.FromMilliseconds(500);
+            var deleteWorkspacesOlderThan = TimeSpan.FromMinutes(30);
+
+            var existingHomeDirectory = new TemporaryDirectory();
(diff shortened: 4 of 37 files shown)


source/Octopus.Tentacle.Contracts/ScriptTicket.cs
```diff
@@ -18,7 +17,7 @@ public bool Equals(ScriptTicket? other)
                 return false;
             if (ReferenceEquals(this, other))
                 return true;
-            return string.Equals(TaskId, other.TaskId);
+            return string.Equals(TaskId, other.TaskId, StringComparison.OrdinalIgnoreCase);
```
sburmanoctopus, line 20: This was a remnant of a direction I backed away from.

But I kept it in to start the conversation about whether this is actually a legit fix.

If you look at the dictionaries in `ScriptService`, they both compare using `OrdinalIgnoreCase` on the keys (i.e. script tickets as strings).

I'd say, the script ticket `F2F78486-A53E-4BF9-9939-CD64FB07E37D` is the same as `f2f78486-a53e-4bf9-9939-cd64fb07e37d`.

source/Octopus.Tentacle/Scripts/ScriptState.cs
```diff
@@ -16,21 +16,18 @@ public class ScriptState
...
         }
 
-        public ScriptState(ScriptTicket scriptTicket, DateTimeOffset created)
+        public ScriptState(DateTimeOffset created)
         {
-            ScriptTicket = scriptTicket;
             Created = created;
         }
-
-        public ScriptTicket ScriptTicket { get; }
```
sburmanoctopus, line 33: Although this was from a direction that was backed away from, `ScriptTicket` was never used (not even in Octopus Server).

So I kept this, as it wasn't wasted work.

source/Tentacle.sln.DotSettings
```diff
@@ -7,6 +7,8 @@
...
 	<s:String x:Key="/Default/CodeStyle/CodeCleanup/SilentCleanupProfile/@EntryValue">Octopus Deploy</s:String>
+	<s:String x:Key="/Default/CodeStyle/CodeFormatting/CSharpCodeStyle/DEFAULT_INTERNAL_MODIFIER/@EntryValue">Implicit</s:String>
```
sburmanoctopus, line 10: I personally prefer seeing `private` and `internal` on methods. But alas, that is not the Octopus way.

This codebase is very inconsistent. I think largely because the original authors who like removing these modifiers wrote most of it, and Rider/ReSharper has not been configured to respect this.

So I have made this inline with Octopus Server. Maybe later, we can compare more of these settings. But for today, I just wanted to stop ReSharper adding in `private` what it was not supposed to.

source/Octopus.Tentacle.Client/Scripts/ScriptExecutionOrchestrator.cs
```diff
@@ -461,9 +461,15 @@ await asyncHalibutFeature
                             clientOperationMetricsBuilder,
                             CancellationToken.None);
 
-                    await actionTask.WaitTillCompletion(onCancellationAbandonCompleteScriptAfter, scriptExecutionCancellationToken);
+                    var actionTaskCompletionResult = await actionTask.WaitTillCompletion(onCancellationAbandonCompleteScriptAfter, scriptExecutionCancellationToken);
```
sburmanoctopus, line 464: This wasn't part of the change. But when I was trying to get failures to happen in tests, I noticed that the `catch` was never getting hit.

Looking at it closer, it's because `WaitTillCompletion` doesn't actually await the task, it just waits for it to complete in some way.

Everywhere else that uses `WaitTillCompletion` follows this pattern. So I think we need to do it here as well. Otherwise we don't log that we couldn't clean up the working directory.

source/Octopus.Tentacle.Tests.Integration/Support/TentacleBuilder.cs
```diff
@@ -13,13 +14,39 @@
 
 namespace Octopus.Tentacle.Tests.Integration.Support
 {
-    public abstract class TentacleBuilder<T> where T : TentacleBuilder<T>
+    public interface ITentacleBuilder
```
sburmanoctopus, line 17: We needed to apply the environment variables regardless of what type of tentacle it was.

However, while in this space I found it quite awkward having this abstract class with two subclasses. Even the "With" existing methods look a touch awkward (where they return a `T`, where `T` is the type of the subclass...)

Looking at the 2 subclasses, there really isn't much there.

I wonder if it would make things simpler if we just combined all 3 of these into 1 class (this class), and simply had 2 methods like `WithPollingTentacle` and `WithListeningTentacle` instead.

Do we like this idea? (if so, I'll do a separate PR for that)

source/Octopus.Tentacle.Tests.Integration/Support/TentacleBuilder.cs
```diff
@@ -13,13 +14,39 @@
...
         protected string CertificatePfxPath = Certificates.TentaclePfxPath;
         protected string TentacleThumbprint = Certificates.TentaclePublicThumbprint;
-        private readonly Regex listeningPortRegex = new Regex(@"^listen:\/\/.+:(\d+)\/");
+        
+        readonly Regex listeningPortRegex = new (@"^listen:\/\/.+:(\d+)\/");
+        readonly Dictionary<string, string> runTentacleEnvironmentVariables = new();
+
+        TemporaryDirectory? homeDirectory;
+
+        protected TemporaryDirectory HomeDirectory
```
sburmanoctopus, line 36: If we had this all in 1 class, then we would treat this like all the other builders (i.e. null/override and then ensure it exists during `Build`).

So this would get cleaned up if we make this 1 class.

source/Octopus.Tentacle.Tests.Integration/WorkspaceCleanerTests.cs
```diff
@@ -0,0 +1,156 @@
...
+            workspaceDirectory = GetWorkspaceDirectoryPath(clientAndTentacle.RunningTentacle.HomeDirectory, startScriptCommand.ScriptTicket.TaskId);
+
+            await clientAndTentacle.TentacleClient.ExecuteScript(startScriptCommand, CancellationToken, null, new InMemoryLog());
+            
+            Directory.Exists(workspaceDirectory).Should().BeFalse("Workspace should be naturally cleaned up after completion");
+        }
+        
+        [Test]
+        [TentacleConfigurations(testDefaultTentacleRuntimeOnly: true)]
+        public async Task WhenTentacleClientCrashesAndTentacleNeverGotToldToComplete_ThenWorkspaceIsConsideredRunning_AndWillNotBeDeleted(TentacleConfigurationTestCase tentacleConfigurationTestCase)
```
sburmanoctopus, line 58: I nearly didn't put this test case in, but I thought it an interesting one that's at least worth looking at during the PR.

source/Octopus.Tentacle/Maintenance/WorkspaceCleaner.cs
```diff
@@ -0,0 +1,71 @@
...
+            this.log = log;
+
+            scriptService = serviceRegistration.GetService<ScriptService>();
+            scriptServiceV2 = serviceRegistration.GetService<ScriptServiceV2>();
+        }
+
+        public async Task Clean(CancellationToken cancellationToken)
+        {
+            var deleteWorkspacesOlderThanDateTimeUtc = clock.GetUtcTime().DateTime.Subtract(configuration.DeleteWorkspacesOlderThanTimeSpan);
+            log.Info($"Cleaning orphaned workspaces older than {deleteWorkspacesOlderThanDateTimeUtc:g}");
```
sburmanoctopus, line 42: I know we wanted this to be more or less invisible to the user. But it felt wrong to not have some form of logging in a scheduled repeating task.

source/Octopus.Tentacle/Util/IOctopusFileSystem.cs
```diff
@@ -11,6 +13,7 @@ public interface IOctopusFileSystem
         bool DirectoryExists(string path);
         void DeleteFile(string path, DeletionOptions? options = null);
         void DeleteDirectory(string path, DeletionOptions? options = null);
+        Task DeleteDirectory(string path, CancellationToken cancellationToken, DeletionOptions? options = null);
```
sburmanoctopus, line 16: This method existed already. It just wasn't in the interface.

source/Octopus.Tentacle/Maintenance/WorkspaceCleaner.cs
```diff
@@ -0,0 +1,71 @@
...
+                    await workspace.Delete(cancellationToken);
+                    deletedCount++;
+                }
+                catch (Exception e)
+                {
+                    log.Error(e, $"Could not delete orphaned workspace {workspace.WorkingDirectory}");
+                }
+            }
+
+            log.Info($"Deleted {deletedCount} orphaned workspace{(deletedCount != 1 ? "s" : "")}");
```
sburmanoctopus, line 68: It wasn't until I was reviewing this PR, that I wondered if the term `orphaned` is acceptable to use. Does it need to change?

If so, what would be a good alternative? (not just for logs, but throughout the code).

"Uncompleted" doesn't work, because that would also apply to scripts that are still running.

"Will never complete" gets wordy.

"Abandoned" maybe?

source/Octopus.Tentacle/Scripts/ScriptState.cs
nathanwoctopusdeploy: Just checking that removing this won't cause any issues if Tentacle loads up the ScriptState that already has the ScriptTicket property.

source/Octopus.Tentacle.Contracts/ScriptTicket.cs
```diff
@@ -33,7 +32,7 @@ public override bool Equals(object? obj)
         }
 
         public override int GetHashCode()
-            => TaskId != null ? TaskId.GetHashCode() : 0;
+            => TaskId != null ? StringComparer.OrdinalIgnoreCase.GetHashCode(TaskId) : 0;
```
nathanwoctopusdeploy, line 35: Is this backward compatible e.g. before and after the change return the same HashCode?

source/Octopus.Tentacle/Maintenance/WorkspaceCleaner.cs
nathanwoctopusdeploy, line 68: Do you think it's useful to log we did nothing e.g. `Deleted 0 orphaned` at info Level or would it be better to suppress this log noise a bit e.g. just log at info when we did some work otherwise log at verbose etc.

nathanwoctopusdeploy: Adding "Unused" as a possible option

source/Octopus.Tentacle/Scripts/ScriptState.cs
sburmanoctopus: It's just JSON deserialization, so any extra properties in the JSON should be ignored.

But the proof is in the pudding 😁 

So this JSON (which includes the script ticket) loaded successfully into the new object (which does not contain script ticket field anymore)
```
{"ScriptTicket":{"TaskId":"ServerTasks-12881-5e494d62-83fa-4229-9292-05f0fcfb91f5"},"Created":"2023-09-14T03:49:05.5241924+00:00","Started":"2023-09-14T03:49:05.5752389+00:00","Completed":"2023-09-14T03:49:07.8029394+00:00","State":2,"ExitCode":0,"RanToCompletion":true}
```
image

source/Octopus.Tentacle.Contracts/ScriptTicket.cs
sburmanoctopus: No, it won't. 

But that was the bug I thought was worth considering fixing. 

Because to me, the script ticket `ServerTasks-12881-5e494d62-83fa-4229-9292-05f0fcfb91f5` should be the same as `SERVERTASKS-12881-5E494D62-83FA-4229-9292-05F0FCFB91F5`

This Linqpad query shows the differences

```
"A".GetHashCode().Dump();
"a".GetHashCode().Dump();
StringComparer.OrdinalIgnoreCase.GetHashCode("A").Dump();
StringComparer.OrdinalIgnoreCase.GetHashCode("a").Dump();
```
Returned this (i.e. case matters normally):
```
1715730010
574887235
1715730010
1715730010
```

But I'm ok to revert this if backwards compatibility is more important?

nathanwoctopusdeploy: I'm considering the edge case here of 

- Script Execution starts
- Tentacle is upgraded
- RPC Retries are enabled so Server "recovers from the Tentacle Restart"
- Tentacle reports the script is unknown instead of it finished/it was running but never completed (or tentacle didn't know it completed)

It's a small issue, an edge case and probably unlikely anyone will hit the issue so maybe ok?

source/Octopus.Tentacle/Maintenance/WorkspaceCleaner.cs
sburmanoctopus: 🤔 I don't have any strong opinions here. So I'll list some statements to help make a choice 😁 

This only runs every 2 hours (which we can make once a day?), so logging will be minimal.

Regardless, we are already logging at info level that we are going to attempt it. So it might be nice to see what the result was. I know Verbose is also being logged, but really, who looks at Verbose unless something goes wrong.

We could combine these. But having the log entry at the start is helpful to know if we aren't reaching it (plus helps show how long it took).

Personally, I'd either leave it, or combine both entries into 1 entry at the end.

But if you think verbose is superior, I can do that?

sburmanoctopus: Yeah, it's a hard one. Does "Unused" feels a little like it was never used?

nathanwoctopusdeploy: > we are already logging at info level that we are going to attempt it

This could be a candidate for Verbose as well

E.g. how useful are these log messages to users?

I think it's useful to know when it did delete something (just incase this is the cause of a problem) or if there was an error when running the job - so these seem like Info or higher candidate.

We're going to run this thing every 2 hours and we did run it and it did nothing seem like good candidate for Verbose logging.

source/Octopus.Tentacle.Contracts/ScriptTicket.cs
sburmanoctopus: Just to clarify, are you saying we will hit this edge case if we fix the bug, or if we leave it in?

I think it might eventuate if we leave it in. This is because I don't think it affects the original `ScriptService`, as that uses a `ConcurrentDictionary` with `StringComparer.OrdinalIgnoreCase`. So it's just a bug for the new `ScriptServiceV2` (as that uses the domain object itself as the key).

But if we fix it, we fix `ScriptServiceV2` as well?

This might be moot, because I would have thought that whatever casing Octopus Server gives the script ticket would be consistent.

Also, Octopus Server would pass around this object as well. That makes me a little nervous about the change. But I still can't look past the obvious original bug... 😑

source/Octopus.Tentacle/Maintenance/WorkspaceCleaner.cs
sburmanoctopus: Sounds good to me. To summarise:
- Make top log Verbose
- Make bottom log Verbose if 0 (maybe even a new message as well)
- Keep bottom log Info if != 0

nathanwoctopusdeploy:
[sc-59496]

source/Octopus.Tentacle.Tests.Integration/Support/TentacleBuilder.cs
nathanwoctopusdeploy: I'm just resolving these as it is adding extra noise and things to read not directly related to this PR.

source/Octopus.Tentacle.Tests.Integration/Support/TentacleBuilder.cs
nathanwoctopusdeploy: I'm just resolving these as it is adding extra noise and things to read not directly related to this PR

source/Octopus.Tentacle/Scripts/ScriptWorkspaceFactory.cs
```diff
@@ -95,9 +97,56 @@ string FindWorkingDirectory(ScriptTicket ticket)
...
+            var baseWorkingDirectory = GetBaseWorkingDirectory();
+            return fileSystem.GetFullPath(Path.Combine(baseWorkingDirectory, ticket.TaskId));
+        }
+        
+        string GetBaseWorkingDirectory()
+        {
+            return Path.Combine(home.HomeDirectory ?? "", WorkDirectory);
+        }
+
+        public List<IScriptWorkspace> GetUncompletedWorkspaces()
```
nathanwoctopusdeploy, line 121: Am I right in thinking these are not necessarily Uncompleted, it's just all the current Workspace Directories.

nathanwoctopusdeploy: Or Completed = Deleted so Uncompleted = undelete!!

I think the name may be a little awkward!! Or tripped me up anyway!!

source/Octopus.Tentacle/Maintenance/WorkspaceCleanerTask.cs
```diff
@@ -0,0 +1,104 @@
...
+            Stop();
+
+            cancellationTokenSource.Dispose();
+        }
+
+        public void Start()
+        {
+            lock (taskLock)
+            {
+                cleanerTask = Task.Run(RunTask);
```
nathanwoctopusdeploy, line 46: nit: could this allow many cleaner task to be run if start was called more than once

source/Octopus.Tentacle/Maintenance/WorkspaceCleaner.cs
nathanwoctopusdeploy: Just confirming I think this should be Verbose

source/Octopus.Tentacle/Maintenance/WorkspaceCleaner.cs
```diff
@@ -0,0 +1,71 @@
...
+            {
+                try
+                {
+                    if (scriptService.IsRunningScript(workspace.ScriptTicket)) continue;
+                    if (scriptServiceV2.IsRunningScript(workspace.ScriptTicket)) continue;
+
+                    var workspaceLogFilePath = workspace.LogFilePath;
+                    if (!File.Exists(workspaceLogFilePath)) continue;
+
+                    var outputLogFileLastWriteTimeUtc = File.GetLastWriteTimeUtc(workspaceLogFilePath);
```
nathanwoctopusdeploy, line 56: Is there a race condition here that we could handle gracefully rather than log error when the file is deleted after we check it exists but before we read it's properties?

Thought process is to avoid Errors in the logs that might concern users when the cause is just a race condition.

source/Octopus.Tentacle/Maintenance/WorkspaceCleaner.cs
```diff
@@ -0,0 +1,71 @@
...
+                    if (outputLogFileLastWriteTimeUtc >= deleteWorkspacesOlderThanDateTimeUtc) continue;
+
+                    await workspace.Delete(cancellationToken);
+                    deletedCount++;
+                }
+                catch (Exception e)
+                {
+                    log.Error(e, $"Could not delete orphaned workspace {workspace.WorkingDirectory}");
```
nathanwoctopusdeploy, line 64: Error may be concerning to a user, it's not a critical problem with Tentacle, maybe Wanring?

source/Octopus.Tentacle/Maintenance/WorkspaceCleaner.cs
nathanwoctopusdeploy: Maybe none of the above then, Just Deleted x workspaces

source/Octopus.Tentacle.Tests.Integration/WorkspaceCleanerTests.cs
```diff
@@ -0,0 +1,183 @@
...
+                            })
+                            .Build())
+                    .Build())
+                .Build(CancellationToken);
+
+            workspaceDirectory = GetWorkspaceDirectoryPath(clientAndTentacle.RunningTentacle.HomeDirectory, startScriptCommand.ScriptTicket.TaskId);
+
+            await clientAndTentacle.TentacleClient.ExecuteScript(startScriptCommand, CancellationToken, null, new InMemoryLog());
+            
+            Directory.Exists(workspaceDirectory).Should().BeFalse("Workspace should be naturally cleaned up after completion");
```
nathanwoctopusdeploy, line 53: I feel these tests are not guaranteed to work as expected 

e.g. could this

- Run the script
- The process finishes so output.log stops being written to
- The RPC calls take longer than 500 ms to get status and start the complete script call
- The cleaner is actually doing the cleanup.

source/Octopus.Tentacle.Tests.Integration/WorkspaceCleanerTests.cs
```diff
@@ -0,0 +1,183 @@
...
+                })
+                .WithTentacleServiceDecorator(new TentacleServiceDecoratorBuilder()
+                    .DecorateScriptServiceV2With(d =>
+                        d.BeforeCompleteScript(async (_, _) =>
+                            {
+                                Directory.Exists(workspaceDirectory).Should().BeTrue("Workspace should exist before we complete");
+
+                                await Task.Delay(2 * 500, CancellationToken);
+
+                                Directory.Exists(workspaceDirectory).Should().BeTrue("Workspace should not have been cleaned up");
```
nathanwoctopusdeploy, line 43: Is there a way we can guarantee the cleaner has run between 39 and 43? E.g. At 39, create a dummy Work folder and Output.log and at 43 wait until and assert it was cleaned up. That way we guarantee we are testing the cleaner and it's run as expected.

source/Octopus.Tentacle.Tests.Integration/WorkspaceCleanerTests.cs
```diff
@@ -0,0 +1,183 @@
...
+        {
+            var cleanerDelay = TimeSpan.FromMilliseconds(500);
+            var deleteWorkspacesOlderThan = TimeSpan.FromMilliseconds(500);
+
+            var startScriptCommand = new StartScriptCommandV2Builder().WithScriptBody(b => b.Print("Hello")).Build();
+
+            await using var clientAndTentacle = await tentacleConfigurationTestCase.CreateBuilder()
+                .WithTentacle(b =>
+                {
+                    b.WithRunTentacleEnvironmentVariable(WorkspaceCleanerConfiguration.CleanerDelayEnvironmentVariableName, cleanerDelay)
```
nathanwoctopusdeploy, line 68: Nit: an extension method could make this easier to read as it's some long lines!

source/Octopus.Tentacle.Tests.Integration/WorkspaceCleanerTests.cs
```diff
@@ -0,0 +1,183 @@
...
+                        .WithRunTentacleEnvironmentVariable(WorkspaceCleanerConfiguration.DeleteWorkspacesOlderThanTimeSpanEnvironmentVariableName, deleteWorkspacesOlderThan);
+                })
+                .WithTentacleServiceDecorator(new TentacleServiceDecoratorBuilder()
+                    .DecorateScriptServiceV2With(d =>
+                        d.BeforeCompleteScript((_, _) => throw new NotImplementedException("Force failure to simulate tentacle client crashing, and ensure we do not completion"))
```
nathanwoctopusdeploy, line 73: nit do not complete | do not complete the script

source/Octopus.Tentacle.Tests.Integration/WorkspaceCleanerTests.cs
```diff
@@ -0,0 +1,183 @@
...
+            
+            await AssertionExtensions
+                .Should(() => clientAndTentacle.TentacleClient.ExecuteScript(startScriptCommand, CancellationToken, null, new InMemoryLog()))
+                .ThrowAsync<NotImplementedException>();
+
+            await Task.Delay(1000, CancellationToken);
+
+            var workspaceDirectory = GetWorkspaceDirectoryPath(clientAndTentacle.RunningTentacle.HomeDirectory, startScriptCommand.ScriptTicket.TaskId);
+
+            Directory.Exists(workspaceDirectory).Should().BeTrue();
```
nathanwoctopusdeploy, line 86: I'm confused!! Shouldn't this be deleted?

source/Octopus.Tentacle.Tests.Integration/WorkspaceCleanerTests.cs
```diff
@@ -0,0 +1,183 @@
...
+                        d.BeforeCompleteScript((_, _) => throw new NotImplementedException("Force failure to simulate tentacle client crashing, and ensure we do not completion"))
+                        .Build())
+                    .Build())
+                .Build(CancellationToken);
+            
+            await AssertionExtensions
+                .Should(() => clientAndTentacle.TentacleClient.ExecuteScript(startScriptCommand, CancellationToken, null, new InMemoryLog()))
+                .ThrowAsync<NotImplementedException>();
+
+            await Task.Delay(1000, CancellationToken);
```
nathanwoctopusdeploy, line 82: nit: more deterministic way of ensuing the cleaner has run

source/Octopus.Tentacle.Tests.Integration/WorkspaceCleanerTests.cs
nathanwoctopusdeploy: Ahh ok I see, the workspace remains in memory until Tentacle is restarted so we don't clean up!!

I think you have caught a missed requirement here. I think we would want to clean this up! Tentacle could run for weeks or months which would render our cleanup job redundant!!

source/Octopus.Tentacle.Tests.Integration/WorkspaceCleanerTests.cs
```diff
@@ -0,0 +1,183 @@
...
+                {
+                    b.WithHomeDirectory(existingHomeDirectory)
+                        .WithRunTentacleEnvironmentVariable(WorkspaceCleanerConfiguration.CleanerDelayEnvironmentVariableName, cleanerDelay)
+                        .WithRunTentacleEnvironmentVariable(WorkspaceCleanerConfiguration.DeleteWorkspacesOlderThanTimeSpanEnvironmentVariableName, deleteWorkspacesOlderThan);
+                })
+                .Build(CancellationToken);
+
+            await Task.Delay(1000, CancellationToken);
+
+            var existingOrphanedWorkspaceDirectory = GivenExistingOrphanedWorkspaceExists(existingHomeDirectory);
```
nathanwoctopusdeploy, line 159: I think we need to revisit what we call orphaned as we should really be able to test this via script service and not by creating a workspace manually

source/Octopus.Tentacle/Maintenance/WorkspaceCleaner.cs
```diff
@@ -0,0 +1,71 @@
...
+            var deleteWorkspacesOlderThanDateTimeUtc = clock.GetUtcTime().DateTime.Subtract(configuration.DeleteWorkspacesOlderThanTimeSpan);
+            log.Info($"Cleaning orphaned workspaces older than {deleteWorkspacesOlderThanDateTimeUtc:g}");
+
+            var deletedCount = 0;
+            var workspaces = scriptWorkspaceFactory.GetUncompletedWorkspaces();
+            foreach (var workspace in workspaces)
+            {
+                try
+                {
+                    if (scriptService.IsRunningScript(workspace.ScriptTicket)) continue;
```
nathanwoctopusdeploy, line 50: Do we need to be able to ask the services if the script is still running / when it stopped.

If it stopped > 24 hours ago then treat it as needing to be cleaned up.

This also raises the question, should we "CompleteScript" in this scenario so we remove the item from the services collection of running scripts!!

nathanwoctopusdeploy (changes requested):
Looking really good.

The main thing to resolve is the issue you spotted in our approach where the cleaner won't clean anything that is orphaned until after a tentacle restart which doesn't get the results we quite wanted!

source/Octopus.Tentacle/Maintenance/WorkspaceCleaner.cs
sburmanoctopus: Agreed. Done

source/Octopus.Tentacle/Maintenance/WorkspaceCleaner.cs
sburmanoctopus: Deleted

source/Octopus.Tentacle/Scripts/ScriptWorkspaceFactory.cs
sburmanoctopus: > Or Completed = Deleted so Uncompleted = not yet deleted!!

Yeah, that was my thinking. Like you say, when we complete, then the directory is deleted.

I thought about "in progress", but that's also not true, as abandoned requests are no longer in progress.

That's why I went for "Uncompleted", as this is true for in progress, as well as abandoned.

🤔 I can't really think of a better name. We could use "Current" if you'd like. Or simply drop any descriptor (e.g. `GetWorkspaces`).

source/Octopus.Tentacle/Maintenance/WorkspaceCleanerTask.cs
sburmanoctopus: It could in theory. But in practice, this is called from `RunAgentCommand`, and that command would fail much earlier if that is started more than once.

Regardless, I added a fix 🙂

source/Octopus.Tentacle/Maintenance/WorkspaceCleaner.cs
sburmanoctopus: Sounds good. Done

source/Octopus.Tentacle/Maintenance/WorkspaceCleaner.cs
sburmanoctopus: Sure 🙂

source/Octopus.Tentacle.Tests.Integration/WorkspaceCleanerTests.cs
sburmanoctopus: 🤔  Creating the file would help. But that only guarantees that it has at least started running, and not necessarily finished.

The only way I can think of to guarantee it has indeed run to completion is to maybe query the system log itself of the running tentacle and count the log lines that show it has finished (and ensure the count has gone up).

But it seems that we currently only log Info level or higher (and these are logged as Verbose) 😑

For now, I've done the dummy file approach, we can discuss this in more detail when you get back 🙂

source/Octopus.Tentacle.Tests.Integration/WorkspaceCleanerTests.cs
sburmanoctopus: Good one. Done!

We always set these together, so I made 1 method for now. We can split it up if/when needed later.

source/Octopus.Tentacle.Tests.Integration/WorkspaceCleanerTests.cs
sburmanoctopus: 🤦  Done

source/Octopus.Tentacle.Tests.Integration/WorkspaceCleanerTests.cs
sburmanoctopus: Have done the same thing as above for now, until we determine if we want to do more (e.g. log querying solution)

source/Octopus.Tentacle.Tests.Integration/WorkspaceCleanerTests.cs
sburmanoctopus: I fixed this up. Now, if it finds a running script, it also queries the `State`.

You mentioned maybe asking "when it stopped". But I decided against that because if Tentacle dies, then the in-memory date would die. So we would still need a way of knowing how old something is (i.e. looking at the log date), so we may as well use the same mechanism for both. Also, this way means not adding extra code to the existing script service 😁

source/Octopus.Tentacle.Tests.Integration/WorkspaceCleanerTests.cs
sburmanoctopus: Actually, after I fixed the "Is Running" issue, the "dummy folder" solution can't work anymore. 

Mainly because as soon as the script finishes running, but before "complete script" is called, it now meets the criteria of a workspace that has to be cleaned up.

What stops it from being cleaned up normally is that we wait 1 day before cleaning up candidates. But we had purposely decreased that time to 500ms in this test in order to clean up the "dummy folder". I.e. they both use the same mechanism.

We could add yet another threshold for files that have just completed. But that feels excessive.

For now, I've put it back the way it was, and we can discuss further when you're back.

source/Octopus.Tentacle.Tests.Integration/WorkspaceCleanerTests.cs
sburmanoctopus: Indeed. Although, I am finding it hard to get this to work.

Now that we've fixed the "Is Running" issue, we can create an abandoned directory by killing a tentacle client.

That main test case is now covered in the renamed `WhenTentacleClientCrashesAndTentacleNeverGotToldToComplete_ThenWorkspaceShouldGetDeletedWhenScriptFinishesRunning` test.

But creating a directory like this would for the other tests has issues, as we would have to reuse the same tentacle (or create a new one that doesn't dispose until we finish).

I think this would be a good one to talk through when you're back.

source/Octopus.Tentacle.Contracts/ScriptTicket.cs
sburmanoctopus: This was made into a [new story.](https://app.shortcut.com/octopusdeploy/story/60772/make-scriptticket-ignore-case-when-doing-comparisons)

source/Octopus.Tentacle.Tests.Integration/WorkspaceCleanerTests.cs
sburmanoctopus: This ended up being a problem. So we fixed the issue.

source/Octopus.Tentacle.Tests.Integration/WorkspaceCleanerTests.cs
sburmanoctopus: We reviewed this together via Zoom, and made the appropriate changes

source/Octopus.Tentacle/Maintenance/WorkspaceCleaner.cs
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        public async Task Clean(CancellationToken cancellationToken)
+        {
+            var deleteWorkspacesOlderThanDateTimeUtc = clock.GetUtcTime().DateTime.Subtract(configuration.DeleteWorkspacesOlderThanTimeSpan);
+            log.Verbose($"Cleaning workspaces older than {deleteWorkspacesOlderThanDateTimeUtc:g}");
+
+            var deletedCount = 0;
+            var workspaces = scriptWorkspaceFactory.GetUncompletedWorkspaces();
+            foreach (var workspace in workspaces)
```
nathanwoctopusdeploy, line 46: nit: we could `cancellationToken.ThrowIfCancellationRerquested()` to help cancel quickly

nathanwoctopusdeploy (approved):
Lovely job. LGTM

source/Octopus.Tentacle/Maintenance/WorkspaceCleaner.cs
sburmanoctopus: @nathanwoctopusdeploy , after doing some manual testing, I came to realize that [File.GetLastWriteTimeUtc](https://learn.microsoft.com/en-us/dotnet/api/system.io.file.getlastwritetimeutc?view=net-7.0) doesn't actually throw an exception when the file doesn't exist. It returns a date from 1601.

So the try/catch I had around here was redundant. 

I think instead, we simply check to see if the file still exists after we call `File.GetLastWriteTimeUtc` (because if it's not there, then it shows the race condition is happening).

Let me know if you think we should put the try/catch back, but I think this is a better solution.

PR #616, opened 2023-09-27
Attempt to fix what looks like a timing issue in ClientScriptExecutionCanBeCancelledWhenRetriesAreEnabled

nathanwoctopusdeploy:
# Background

Attempt to fix what looks like a timing issue in ClientScriptExecutionCanBeCancelledWhenRetriesAreEnabled

## Before

Occasional flakey tests results e.g. https://build.octopushq.com/buildConfiguration/TeamFireAndMotion_OctopusTentacleNet48_SensibleDefaultsChainFullChain/9718957?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildTestsSection=true&expandBuildProblemsSection=true

## After

It may not be fixed but there are a series of green runs 

image

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionCanBeCancelledWhenRetriesAreEnabled.cs
+++ b/source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionCanBeCancelledWhenRetriesAreEnabled.cs
@@ -78,7 +78,7 @@ namespace Octopus.Tentacle.Tests.Integration
                                     await syncOrAsyncHalibut.WhenSync(() => scriptServiceV2.EnsureTentacleIsConnectedToServer(Logger))
                                         .WhenAsync(async () => await asyncScriptServiceV2.EnsureTentacleIsConnectedToServer(Logger));
 
-                                    PauseOrStopPortForwarder(rpcCallStage, portForwarder.Value, responseMessageTcpKiller, rpcCallHasStarted);
+                                    await PauseOrStopPortForwarder(rpcCallStage, portForwarder.Value, responseMessageTcpKiller, rpcCallHasStarted);
                                     if (rpcCallStage == RpcCallStage.Connecting && tentacleConfigurationTestCase.TentacleType == TentacleType.Polling)
                                     {
                                         await service.EnsurePollingQueueWontSendMessageToDisconnectedTentacles(Logger);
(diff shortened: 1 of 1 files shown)


source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionCanBeCancelledWhenRetriesAreEnabled.cs
```diff
@@ -504,19 +504,29 @@ public async Task DuringCompleteScript_ScriptExecutionCanBeCancelled(TentacleCon
...
-                responseMessageTcpKiller.PauseConnectionOnNextResponse(() => rpcCallHasStarted.Value = true);
+#pragma warning disable CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
+                responseMessageTcpKiller.PauseConnectionOnNextResponse(() => SetRpcCallWeAreInterestedInAsStarted(rpcCallHasStarted));
+#pragma warning restore CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
+            }
+
+            static async Task SetRpcCallWeAreInterestedInAsStarted(Reference<bool> rpcCallHasStarted)
+            {
+                // Allow the port forwarder some time to stop
+                await Task.Delay(TimeSpan.FromSeconds(5));
```
sburmanoctopus, line 528: I know you would have done this if it were possible, but is there a way we can detect that the port forwarder has actually stopped?

sburmanoctopus (approved):
LGTM

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionCanBeCancelledWhenRetriesAreEnabled.cs
nathanwoctopusdeploy: I didn't look into it. It would be great to be able to know precisely when everything is stopped but it also sounds like a possible time sync and we are at the mercy of the Operating System and how it is actually dealing with TCP Connections. 

I'm sure this area could be dramatically improved, and all these nasty, prone to flake or fragile with change tests could be improved.

At the moment I'm going for the simplest (unless it's just horrible) approach to try and squish the flakey tests we have!!

PR #618, opened 2023-09-28
Add support for .Net 6 Windows tentacle upgrader

samdanaei:
# Background

The Windows auto-upgrade installer (managed by Octopus Server) requires an `Octopus.Tentacle.Upgrader` file to kick off the auto-upgrade process. It requires .Net Framework to be installed on the user's machine. 

Now that we are rolling out support for .Net 6 standalone tentacles on Windows, we need to include standalone versions of the `Octopus.Tentacle.Upgrader` so that customers don't need to have .Net Framework installed on their Windows machines to allow Octopus server to install and run tentacles.

The `Octopus.Tentacle.CrossPlatformBundle` should now include two new files:
- Octopus.Tentacle.Upgrader-net6.0-win-x86.exe
- Octopus.Tentacle.Upgrader-net6.0-win-x64.exe

[sc-59761]

# Results

It should get rid of the following error when the server tries to upgrade a .Net 6 tentacle:

image

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Upgrader/ServiceBouncer.cs
+++ b/source/Octopus.Tentacle.Upgrader/ServiceBouncer.cs
@@ -6,7 +6,6 @@ using System.Linq;
 using System.ServiceProcess;
 using System.Threading;
 using Microsoft.Win32;
-#pragma warning disable CA1416
 
 namespace Octopus.Tentacle.Upgrader
 {
--- a/source/Octopus.Tentacle.Upgrader/ServiceBouncer.cs
+++ b/source/Octopus.Tentacle.Upgrader/ServiceBouncer.cs
@@ -6,6 +6,7 @@ using System.Linq;
 using System.ServiceProcess;
 using System.Threading;
 using Microsoft.Win32;
+#pragma warning disable CA1416
 
 namespace Octopus.Tentacle.Upgrader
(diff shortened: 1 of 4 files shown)


source/Octopus.Tentacle.Upgrader/ServiceBouncer.cs
```diff
@@ -6,6 +6,7 @@
 using System.ServiceProcess;
 using System.Threading;
 using Microsoft.Win32;
+#pragma warning disable CA1416
```
evolutionise, line 9: Does this not get handled by the global suppression?

evolutionise (approved):
LGTM, one minor question

source/Octopus.Tentacle.Upgrader/ServiceBouncer.cs
samdanaei: Thanks, good pickup. I've applied the CA1416 suppression to the whole project, since this project is only for Windows, and that's what the warning was for.

samdanaei: _It wasn't being suppressed before as the global suppressions were only being done for certain members._

PR #622, opened 2023-10-05
Update OctopusClient version in Tentacle

scme0:
# Background

Before we cam complete our work on [#project-k8s-agent](https://octopusdeploy.slack.com/archives/C05KK1G85J5) we need to update Octopus Client in Tentacle as it's currently very out of date.

[sc-61268]

# Results

I have updated to the latest version of OctopusClient which has now made the Async methods which don't take a `CancellationToken`, `obsolete`. This means the Tentacle build fails.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Manager.Tentacle/Dialogs/ServerConnectionDialog.xaml.cs
+++ b/source/Octopus.Manager.Tentacle/Dialogs/ServerConnectionDialog.xaml.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 using System.Windows;
 using System.Windows.Controls;
@@ -35,7 +36,7 @@ namespace Octopus.Manager.Tentacle.Dialogs
             Loaded += async (a, e) =>
             {
                 model.ContributeSensitiveValues(logger);
-                await model.VerifyCredentials(logger);
+                await model.VerifyCredentials(logger, CancellationToken.None);
                 if (model.HaveCredentialsBeenVerified)
                 {
                     setProgressBarToStatus(false, true);
--- a/source/Octopus.Manager.Tentacle/Dialogs/ServerConnectionDialog.xaml.cs
+++ b/source/Octopus.Manager.Tentacle/Dialogs/ServerConnectionDialog.xaml.cs
@@ -2,7 +2,6 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
-using System.Threading;
 using System.Threading.Tasks;
 using System.Windows;
 using System.Windows.Controls;
@@ -36,7 +35,7 @@ namespace Octopus.Manager.Tentacle.Dialogs
             Loaded += async (a, e) =>
             {
                 model.ContributeSensitiveValues(logger);
-                await model.VerifyCredentials(logger, CancellationToken.None);
+                await model.VerifyCredentials(logger);
                 if (model.HaveCredentialsBeenVerified)
                 {
                     setProgressBarToStatus(false, true);
--- a/source/Octopus.Manager.Tentacle/TentacleConfiguration/SetupWizard/TentacleSetupWizardModel.cs
+++ b/source/Octopus.Manager.Tentacle/TentacleConfiguration/SetupWizard/TentacleSetupWizardModel.cs
@@ -4,7 +4,6 @@ using System.Collections.ObjectModel;
 using System.IO;
 using System.Linq;
 using System.Net;
-using System.Threading;
 using System.Threading.Tasks;
 using FluentValidation;
 using Octopus.Client;
@@ -521,9 +520,7 @@ namespace Octopus.Manager.Tentacle.TentacleConfiguration.SetupWizard
                 OnPropertyChanged();
                 if (!string.IsNullOrEmpty(selectedSpace))
 #pragma warning disable 4014 // we want this to be async
-                    LoadSpaceData(async (client, ct) => await LoadSpaceSpecificData(client, ct), CancellationToken.None)
-                        .GetAwaiter()
-                        .GetResult();
+                    LoadSpaceData(async client => await LoadSpaceSpecificData(client));
 #pragma warning restore 4014
             }
         }
@@ -559,7 +556,7 @@ namespace Octopus.Manager.Tentacle.TentacleConfiguration.SetupWizard
 
         public ProxyWizardModel ProxyWizardModel => proxyWizardModel;
 
-        public async Task VerifyCredentials(ILog logger, CancellationToken cancellationToken)
+        public async Task VerifyCredentials(ILog logger)
         {
             try
             {
@@ -597,7 +594,7 @@ namespace Octopus.Manager.Tentacle.TentacleConfiguration.SetupWizard
                         PotentialSpaces = new string[] { };
                         AreSpacesSupported = false;
 
-                        await LoadDataFromSpace(logger.Info, repository, cancellationToken);
+                        await LoadDataFromSpace(logger.Info, repository);
                     }
 
                     logger.Info("Credentials verified");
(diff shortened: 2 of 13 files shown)


source/Octopus.Manager.Tentacle/Dialogs/ServerConnectionDialog.xaml.cs
```diff
@@ -35,7 +36,7 @@ public ServerConnectionDialog(TentacleSetupWizardModel model)
             Loaded += async (a, e) =>
             {
                 model.ContributeSensitiveValues(logger);
-                await model.VerifyCredentials(logger);
+                await model.VerifyCredentials(logger, CancellationToken.None);
```
scme0, line 39: I am really unsure on what we can do here other than pass in `CancellationToken.None`. Is this acceptable?

source/Octopus.Manager.Tentacle/TentacleConfiguration/SetupWizard/TentacleSetupWizardModel.cs
```diff
@@ -519,9 +520,9 @@ public string SelectedSpace
                 selectedSpace = value;
                 OnPropertyChanged();
                 if (!string.IsNullOrEmpty(selectedSpace))
-#pragma warning disable 4014 // we want this to be async
-                    LoadSpaceData(async client => await LoadSpaceSpecificData(client));
-#pragma warning restore 4014
+                {
+                    var _ = LoadSpaceData(LoadSpaceSpecificData, CancellationToken.None);
+                }
```
scme0, line 525: Unsure on how to best do this but just calling this and letting it run in the background seems like the accepted solution online.

source/Octopus.Tentacle/Commands/DeregisterMachineCommand.cs
```diff
@@ -48,22 +51,30 @@ public class DeregisterMachineCommand : AbstractStandardCommand
...
             base.Start();
-            StartAsync().GetAwaiter().GetResult();
+            StartAsync(cancellationTokenSource.Token).GetAwaiter().GetResult();
+        }
+
+        protected override void Stop()
+        {
+            base.Stop();
+
+            cancellationTokenSource?.Cancel();
```
scme0, line 63: I'm not sure about this - If I'm using `GetAwaiter().GetResult();` does it make much sense to create a token and cancel it on `Stop()`? As the `Start()` method won't complete until `StartAsync()` has completed.

source/Octopus.Manager.Tentacle/Dialogs/ServerConnectionDialog.xaml.cs
APErebus: As we are no providing any way to cancel the operation, then I think this is acceptable

source/Octopus.Manager.Tentacle/TentacleConfiguration/SetupWizard/TentacleSetupWizardModel.cs
APErebus: Hrm, this does seem a bit dodgy... I feel like this might have problems if there is an exception (or the exception might just get swallowed 🤔 )

source/Octopus.Tentacle/Commands/DeregisterMachineCommand.cs
APErebus: Hrm, it depends on how `Stop()` get's called. Does it get called in the same method chain as the `Start()`?

What is the purpose of `Stop()`? Is it to stop a long-running command?

scme0: Start and Stop are dictated by the `ICommandHost` being used. 

The implementations are:
Screenshot 2023-10-06 at 10 04 38

I guess if we wanted to keep the existing functionality, we could just pass in `CancellationToken.None`.

source/Octopus.Manager.Tentacle/TentacleConfiguration/SetupWizard/TentacleSetupWizardModel.cs
scme0: The actual work in LoadSpaceData is done in a `try...catch...finally` so I think it's ok in terms of exceptions...

source/Octopus.Manager.Tentacle/TentacleConfiguration/SetupWizard/TentacleSetupWizardModel.cs
```diff
@@ -676,16 +679,16 @@ Task<IOctopusAsyncClient> CreateClient()
...
             var spaceRepository = await new SpaceRepositoryFactory().CreateSpaceRepository(client, SelectedSpace);
             await LoadDataFromSpace(_ =>
             {
                 /*users aren't actually interested in these progress messages, and we have nowhere to display them*/
-            }, spaceRepository);
+            }, spaceRepository, cancellationToken);
         }
 
-        async Task LoadSpaceData(Func<IOctopusAsyncClient, Task> loadAction)
+        async Task LoadSpaceData(Func<IOctopusAsyncClient, CancellationToken, Task> loadAction, CancellationToken cancellationToken)
```
scme0, line 691: @APErebus if you look at this method, you'll see how it's handling exceptions.

Related to this thread: https://github.com/OctopusDeploy/OctopusTentacle/pull/622#discussion_r1348082515

source/Octopus.Manager.Tentacle/TentacleConfiguration/SetupWizard/TentacleSetupWizardModel.cs
scme0: Also the way it was written before was effectively the same - it ran LoadSpaceData as an async method without awaiting the task.

source/Octopus.Tentacle/Commands/DeregisterMachineCommand.cs
zentron: The stop should be getting called from a different thread that is listening to termination signals. 
In theory the `Start` method is blocking in both scenarios, really the `Start()` method itself should be async as well all the way up to the main entrypoint.

The task cancellation (e.g. `Ctrl+C`) should be signaled on its own thread and therefore setting this cancellation should  provide the signal to all the later tasks using that cancellation token. Ultimately the process will still be killed, but the signal provides an opportunity for a "graceful shutdown".

zentron: Since it probably wasn't graceful before its probably not much different to _not_ provide the cancellation token, but then that loses some of the nice benefits of the async+cancellation token capabilities.

source/Octopus.Tentacle/Commands/RegisterMachineCommandBase.cs
```diff
@@ -79,11 +82,19 @@ public abstract class RegisterMachineCommandBase<TRegistrationOperationType> : A
 
         protected override void Start()
         {
+            cancellationTokenSource = new CancellationTokenSource();
```
zentron, line 85: Feels like the cancellation should move up to potentially the `OctopusProgram` level to be passed down and handled at that level rather at the command level

source/Octopus.Manager.Tentacle/Dialogs/ServerConnectionDialog.xaml.cs
zentron: yeah if we dont have an existing token handing around at this context perhaps a CancellationToken with a timeout might be a little safer.  Technically the Http requests themselves should have a timeout as well so its probably not entirely needed, but even a verification of 5 minutes would safe us from any weird hanging results.

scme0:
I have reverted all the extra changes and just added a `CancellationToken.None` to all methods requiring a CancellationToken. This is a stop-gap which maintains the existing functionality. @OctopusDeploy/team-server-at-scale will need to create a ticket to follow this up with a more thought out and complete solution.

acodrington (approved):
After reviewing the changes between `11.3.3453` and `14.3.868`, this looks like a fairly safe upgrade. See [this slack post](https://octopusdeploy.slack.com/archives/C04GAH78491/p1696566050073359?thread_ts=1696383212.787839&cid=C04GAH78491) for more details.

SaST will look into properly supporting CancellationTokens and async in Tentacle in the future, but using `CancellationToken.None` as a stop gap measure is adequate as we currently do not support cancelling these operations in Tentacle.

acodrington:
[sc-61163]

PR #624, opened 2023-10-10
Add token support to Tentacle

scme0:
# Background

As part of work on the Kubernetes Agent work, we are updating Tentacle to take a JWT Token for authentication instead of Username and Password or API Key.

[sc-60667]

# Results

I've added a `--bearerToken` command line option so that we can specify it instead of an API Key.

I've manually tested this by installing a Tentacle Target in kubernetes for both linux and windows so the `configure-tentacle.sh` and `.ps1` scripts have been tested 👍 

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Octopus.Tentacle/Octopus.Tentacle.csproj
```diff
@@ -41,7 +41,7 @@
 		<DefineConstants>$(DefineConstants);HTTP_CLIENT_SUPPORTS_SSL_OPTIONS;REQUIRES_EXPLICIT_LOG_CONFIG;REQUIRES_CODE_PAGE_PROVIDER;USER_INTERACTIVE_DOES_NOT_WORK;DEFAULT_PROXY_IS_NOT_AVAILABLE;HAS_NULLABLE_REF_TYPES</DefineConstants>
 	</PropertyGroup>
 	<ItemGroup>
-		<PackageReference Include="Octopus.Client" Version="14.3.884" />
+		<PackageReference Include="Octopus.Client" Version="14.3.892-scme-add-support-for-token-authentication" />
```
APErebus, line 44: This should be a proper release right?

source/Octopus.Tentacle/Commands/RegisterMachineCommandBase.cs
```diff
@@ -218,7 +218,7 @@ Uri GetActiveTentacleAddress()
                     serverCommsPort ??= serverCommsAddressUri.Port;
                 }
 
-                return new Uri($"https://{serverCommsAddressUri.Host}:{serverCommsPort}");
+                return new Uri($"https://{serverCommsAddressUri?.Host}:{serverCommsPort}");
```
APErebus, line 221: I'm assuming this is possibly null as `api.ServerUri` might be null? If so, would it not be better to throw earlier and make sure it's not null at this point? having a null host here doesn't make sense

APErebus (changes requested):
2 minor things, but the rest looks good!

source/Octopus.Tentacle/Commands/RegisterMachineCommandBase.cs
scme0: Good point 👍 Fixed.

source/Octopus.Tentacle/Octopus.Tentacle.csproj
scme0: Yep, fixed now 👍

APErebus (approved):
Excellent! Nice work!

PR #625, opened 2023-10-11
Add optional ExecutionProperties to StartScriptCommandV2

APErebus:
# Background

Adds a new optional field to the `StartScriptCommandV2`, the `ExecutionProperties`. The intention is that this is for extra properties to be passed from Server relating to the Tentacles execution of the script.

This is a preliminary PR for the changes relating to the Octopus Kubernetes Agent, where we will need to pass container image (execution container) when the scripts are being executed via Kubernetes Jobs.

# Results

Adds the new optional `Dictionary<string,string> ExecutionProperties { get; }`. This is not currently used for running scripts via the shell.

I have tested this locally and a Server sending the command without this property is successfully executed. As this is an additional, optional property, this should be forwards and backwards compatible.

Shortcut story: [sc-61739]

# How to review this PR

Is the name `ExecutionProperties` an acceptable name? I didn't want to use "arguments" as we already have that for the scripts args.

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.CommonTestUtils/Builders/StartScriptCommandV2Builder.cs
+++ b/source/Octopus.Tentacle.CommonTestUtils/Builders/StartScriptCommandV2Builder.cs
@@ -21,6 +21,7 @@ namespace Octopus.Tentacle.CommonTestUtils.Builders
         string taskId = Guid.NewGuid().ToString();
         ScriptTicket scriptTicket = new ScriptTicket(Guid.NewGuid().ToString());
         TimeSpan? durationStartScriptCanWaitForScriptToFinish;
+        Dictionary<string, string>? executionProperties;
 
         public StartScriptCommandV2Builder WithScriptBody(string scriptBody)
         {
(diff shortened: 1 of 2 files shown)


scme0 (approved):
🚀

APErebus:
@zentron This is the "property bag" approach we discussed earlier

zentron:
easy and simple. Might be already done, but perhaps a quick backwards/forwards compatibility check to tick this off.

Probably bring in someone from the `Execution@Scale` team for any input. Its a bit of a cheat way of not breaking compatibility by having a fairly stringly tied contract although it allows us to move forward without having to break tentacle or server.

sburmanoctopus (commented):
These changes look fine 🙂 

Although, we were wondering if we wanted to go the generic all-purpose property bag, or whether a more specific contract would be preferred.

A tighter contract would mean extra changes would be needed as more properties are required, but it would make it easier to understand what is going on (and harder to misuse).

We also discussed whether it would make sense to have a separate command and/or service for the Kubernetes agent, but that will depend on how much work that would involve to make that happen.

After a meeting with @APErebus, he said that he would bring these thoughts to the team.

APErebus:
I'm closing this PR as we have moved a _long_ way from this design

PR #627, opened 2023-10-11
Convert IWorkspaceCleanerTask to Lazy<IWorkspaceCleanerTask>

samdanaei:
# Background

From @sburmanoctopus: 

> The newly introduced `WorkspaceCleanerTask`, through its dependency tree, eventually takes a dependency on the `IKeyValueStore`.
> 
> This object validates some of the configuration. So when it does that, it might not be valid (e.g. in E2E tests), and throws an exception.
> 
> This, in turn, causes the `WorkspaceCleanerTask` to also throw a dependency resolution exception.
> 
> Looking at the `RunAgentCommand` (which is what creates the cleaner task), it seems like this isn't the first time this problem has happened.
> 
> And the go-to move we have done several times in this class already is to add the dependency using a Lazy.
> 
> This allows the class to be created when its dependencies are not legit yet. Which is what we need when calling commands like 'help'.
> 

Additionally, this PR allows a version bump to happen to a 7.1.x version in Octopus Server. Tests may be added with a separate PR. See [sc-61789].

# Results

## Before

```
> .\Tentacle.exe agent --help
===============================================================================
An error occurred during the activation of a particular registration. See the inner exception for details. Registration: Activator = RunAgentCommand (ReflectionActivator), Services = [Octopus.Tentacle.Startup.ICommand], Lifetime = Autofac.Core.Lifetime.CurrentScopeLifetime, Sharing = None, Ownership = OwnedByLifetimeScope ---> An error occurred during the activation of a particular registration. See the inner exception for details. Registration: Activator = WorkspaceCleanerTask (ReflectionActivator), Services = [Octopus.Tentacle.Maintenance.IWorkspaceCleanerTask]

...

--Inner Exception--
There is more than one instance of Tentacle configured on this machine. Please pass --instance=INSTANCENAME when invoking this command to target a specific instance. Available instances: e2e-8DBC4407FC18339, e2e-8DBC5B9D575A5A7, e2e-8DBC6508681DD80, e2e-8DBC652D0477A30, e2e-8DBC6568588DC41, e2e-8DBC65CF4042424, e2e-8DBC6666255D89B, e2e-8DBC97B01EFF415, e2e-8DBC97BD98567CE, e2e-8DBC9AD048274A0, e2e-8DBC9AF1FFB9477, sast-net6-tentacle-upgrade-script-Tentacle0, Tentacle0001.
Octopus.Tentacle.ControlledFailureException
   at Octopus.Tentacle.Configuration.Instances.ApplicationInstanceStore.GetDefaultRegistryRecord() in ApplicationInstanceStore.cs:line 172
   at Octopus.Tentacle.Configuration.Instances.ApplicationInstanceStore.LoadInstanceDetails(String instanceName) in ApplicationInstanceStore.cs:line 54
   at Octopus.Tentacle.Configuration.Instances.ApplicationInstanceSelector.LocateApplicationPrimaryConfiguration() in ApplicationInstanceSelector.cs:line 125
   at Octopus.Tentacle.Configuration.Instances.ApplicationInstanceSelector.LoadInstance() in ApplicationInstanceSelector.cs:line 67
   at Octopus.Tentacle.Configuration.Instances.ApplicationInstanceSelector.LoadCurrentInstance() in ApplicationInstanceSelector.cs:line 57
   at Octopus.Tentacle.Configuration.ConfigurationModule.<>c.<Load>b__3_2(IComponentContext c) in ConfigurationModule.cs:line 69
   at Autofac.Builder.RegistrationBuilder.<>c__DisplayClass0_0`1.<ForDelegate>b__0(IComponentContext c, IEnumerable`1 p)   at Autofac.Core.Activators.Delegate.DelegateActivator.ActivateInstance(IComponentContext context, IEnumerable`1 parameters)
   at Autofac.Core.Resolving.InstanceLookup.Activate(IEnumerable`1 parameters)
```

## After

```
>  .\Tentacle.exe agent --help
Usage: Tentacle agent [<options>]

Where [<options>] is any of:

      --instance=VALUE       Name of the instance to use
      --config=VALUE         Configuration file to use
      --wait=VALUE           Delay (ms) before starting
      --console              Don't attempt to run as a service, even if the
                               user is non-interactive

Or one of the common options:

      --help                 Show detailed help for this command
```
# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle/Commands/RunAgentCommand.cs
+++ b/source/Octopus.Tentacle/Commands/RunAgentCommand.cs
@@ -28,7 +28,7 @@ namespace Octopus.Tentacle.Commands
         readonly Lazy<IProxyInitializer> proxyInitializer;
         readonly IWindowsLocalAdminRightsChecker windowsLocalAdminRightsChecker;
         readonly AppVersion appVersion;
-        readonly IWorkspaceCleanerTask workspaceCleanerTask;
+        readonly Lazy<IWorkspaceCleanerTask> workspaceCleanerTask;
         int wait;
         bool halibutHasStarted;
 
@@ -46,7 +46,7 @@ namespace Octopus.Tentacle.Commands
             IWindowsLocalAdminRightsChecker windowsLocalAdminRightsChecker,
             AppVersion appVersion,
             ILogFileOnlyLogger logFileOnlyLogger,
-            IWorkspaceCleanerTask workspaceCleanerTask) : base(selector, log, logFileOnlyLogger)
+            Lazy<IWorkspaceCleanerTask> workspaceCleanerTask) : base(selector, log, logFileOnlyLogger)
         {
             this.halibut = halibut;
             this.configuration = configuration;
@@ -122,7 +122,7 @@ namespace Octopus.Tentacle.Commands
             halibut.Value.Start();
             halibutHasStarted = true;
 
-            workspaceCleanerTask.Start();
+            workspaceCleanerTask.Value.Start();
 
             Runtime.WaitForUserToExit();
         }
@@ -146,7 +146,7 @@ namespace Octopus.Tentacle.Commands
                 halibut.Value.Stop();
             }
 
-            workspaceCleanerTask.Stop();
+            workspaceCleanerTask.Value.Stop();
         }
     }
 }
--- a/source/Octopus.Tentacle/Commands/RunAgentCommand.cs
+++ b/source/Octopus.Tentacle/Commands/RunAgentCommand.cs
@@ -18,19 +18,23 @@ namespace Octopus.Tentacle.Commands
 {
     public class RunAgentCommand : AbstractStandardCommand
     {
+        // The lazy dependencies allow the class to be created when its dependencies are not initialised yet.
+        // This is needed when calling commands like 'help' or providing feedback on validation (e.g. when no --instance parameter provided)
         readonly Lazy<IHalibutInitializer> halibut;
         readonly Lazy<IWritableTentacleConfiguration> configuration;
         readonly Lazy<IHomeConfiguration> home;
         readonly Lazy<IProxyConfiguration> proxyConfiguration;
+        readonly Lazy<IProxyInitializer> proxyInitializer;
+        readonly Lazy<IWorkspaceCleanerTask> workspaceCleanerTask;
+        
         readonly ISleep sleep;
         readonly ISystemLog log;
         readonly IApplicationInstanceSelector selector;
-        readonly Lazy<IProxyInitializer> proxyInitializer;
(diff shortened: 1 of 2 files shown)


source/Octopus.Tentacle/Commands/RunAgentCommand.cs
```diff
@@ -46,7 +46,7 @@ public class RunAgentCommand : AbstractStandardCommand
             IWindowsLocalAdminRightsChecker windowsLocalAdminRightsChecker,
             AppVersion appVersion,
             ILogFileOnlyLogger logFileOnlyLogger,
-            IWorkspaceCleanerTask workspaceCleanerTask) : base(selector, log, logFileOnlyLogger)
+            Lazy<IWorkspaceCleanerTask> workspaceCleanerTask) : base(selector, log, logFileOnlyLogger)
```
sburmanoctopus, line 49: Nit: Like we discussed yesterday, I think it's worth putting a comment here stating why we do the `Lazy` thing (for future travellers 😁)

Also, do we think it's worth grouping all the `Lazy` dependencies together? It would make the comment more effective. I don't think it will have any flow-on affects 🤔 I'll leave the choice to you 😁

source/Octopus.Tentacle/Commands/RunAgentCommand.cs
```diff
@@ -146,7 +146,7 @@ protected override void Stop()
                 halibut.Value.Stop();
             }
 
-            workspaceCleanerTask.Stop();
+            workspaceCleanerTask.Value.Stop();
```
sburmanoctopus, line 149: If `RunAgentCommand.Stop` is called before the dependencies are setup, we could crash in the same way we did in the constructor.

Do we think this is possible? Or are you happy that this is a situation that cannot happen?

sburmanoctopus (approved):
LGTM. Some minor questions and comments, but this will fix the issue. Nice one 😁

source/Octopus.Tentacle/Commands/RunAgentCommand.cs
samdanaei: I had created a separate story (sc-61799) to have another discussion and look around tidying up the Lazy dependencies. A comment is definitely good though, for the time being 👍

source/Octopus.Tentacle/Commands/RunAgentCommand.cs
samdanaei: I think the safest way to go for now is just to have a check, like `halibutHasStarted`. I do think it would be safe to assume that if the "has started" flag is true, then there should be a `Value`. We'll take another look with sc-61799. Let me know if you have any concerns though 👂

evolutionise (approved):

PR #629, opened 2023-10-13
Ignore development scripts

APErebus:
I have some development scripts for registering a tentacle locally but I don't want these checked in (as they have API keys in them), so I want to ignore them


acodrington (approved):
👍🏻

PR #630, opened 2023-10-16
Copy Client & Contracts to local packages with unique timestamped version

APErebus:
If you are making changes to the Clients and Contracts libraries and want to use them locally in Server, to make this easier we add a unique timestamp and target to all Local builds.

Also adds a new Target to build and copy these two packages to the LocalPackages directory


scme0 (approved):
Nice! This will make local dev a lot easier 👍

PR #631, opened 2023-10-16
Fixing script ticket casing equality

sburmanoctopus:
[sc-60772]

# Background

When comparing the [ScriptService](https://github.com/OctopusDeploy/OctopusTentacle/blob/c8c61358a053365e30b6df33d0248252db6bc856/source/Octopus.Tentacle/Services/Scripts/ScriptService.cs#L11) and [ScriptServiceV2](https://github.com/OctopusDeploy/OctopusTentacle/blob/c8c61358a053365e30b6df33d0248252db6bc856/source/Octopus.Tentacle/Services/Scripts/ScriptServiceV2.cs), it was noticed that `ScriptService` was doing lookups on the `TaskId` of the `ScriptTicket` ignoring case. But `ScriptServiceV2` was doing lookups on the `ScriptTicket` itself (therefore using `GetHashCode`, which respects casing). These were inconsistent. 

# Results

## Before

If someone made a request of the V2 API using a `ScriptTicket` with different casing, it would not return the correct result.

## After

Now, we ignore casing for `GetHashCode` and `Equals` for a `ScriptTicket`, as different casing still means the same `ScriptTicket`. 

As a result, V2 requests now behave correctly.

We looked through Octopus Server and Tentacle, and could not find anywhere that this change would cause a problem.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Tests/Contracts/ScriptTicketFixture.cs
+++ b/source/Octopus.Tentacle.Tests/Contracts/ScriptTicketFixture.cs
@@ -1,0 +1,32 @@
+﻿using FluentAssertions;
+using NUnit.Framework;
+using Octopus.Tentacle.Contracts;
+
+namespace Octopus.Tentacle.Tests.Contracts
+{
+    [TestFixture]
+    public class ScriptTicketFixture
+    {
+        [Test]
+        public void GetHashCodeShouldIgnoreCasing()
+        {
+            var originalScriptTicket = new ScriptTicket("MixedCasing");
+            var upperCaseScriptTicket = new ScriptTicket(originalScriptTicket.TaskId.ToUpper());
+            var lowerCaseScriptTicket = new ScriptTicket(originalScriptTicket.TaskId.ToLower());
+
+            upperCaseScriptTicket.GetHashCode().Should().Be(originalScriptTicket.GetHashCode());
+            lowerCaseScriptTicket.GetHashCode().Should().Be(originalScriptTicket.GetHashCode());
+        }
+
+        [Test]
+        public void EqualityShouldIgnoreCasing()
+        {
+            var originalScriptTicket = new ScriptTicket("MixedCasing");
+            var upperCaseScriptTicket = new ScriptTicket(originalScriptTicket.TaskId.ToUpper());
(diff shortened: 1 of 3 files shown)


source/Octopus.Tentacle.Tests/Contracts/ScriptTicketFixture.cs
```diff
@@ -0,0 +1,32 @@
...
+        }
+
+        [Test]
+        public void EqualityShouldIgnoreCasing()
+        {
+            var originalScriptTicket = new ScriptTicket("MixedCasing");
+            var upperCaseScriptTicket = new ScriptTicket(originalScriptTicket.TaskId.ToUpper());
+            var lowerCaseScriptTicket = new ScriptTicket(originalScriptTicket.TaskId.ToLower());
+
+            upperCaseScriptTicket.Should().Be(originalScriptTicket);
```
gb-8, line 28: Since ScriptTicket has a custom equality implementation, I think we should be explicitly testing all the ways equality can be tested.

This test relies on whichever equality test `FluentAssertions` is choosing to use. It would be easier to see if we are covering all required cases if we had an 'act' part that used the explicit method (`Equals(...)`, `==`, `!=`, etc.), and then assert on the result.

gb-8 (commented):
The equality and hashcode methods look sound.
Aside from the test cases I commented on, I would love to have a quick chat to make sure I am across all the usages/impacts, so I can understand risk etc.
Could we have a quick chat after R&D weekly?

source/Octopus.Tentacle.Tests/Contracts/ScriptTicketFixture.cs
sburmanoctopus: Good call. Fixed!

gb-8 (approved):
As per discussion with @sburmanoctopus:
 - Change looks good and appropriate.
 - Test coverage looks good.
 - It's hard to introduce additional safety checks with this change, so we will be relying on existing monitoring that has recently been introduced to be able to detect any problems as the change rolls out.

acodrington (approved):
Tests look good, coverage is appropriate. As a manual review of possible impacts came up clean, I think this is good to go.

PR #632, opened 2023-10-17
Disposing of the CancellationTokenSource used in script service V2

sburmanoctopus:
[sc-60771]

# Background

It was noticed that `ScriptServiceV2` was not disposing the `CancellationTokenSource` that it was creating.

# Results

Related to OctopusDeploy/Issues#8229

## Before

Since we were not disposing the cancelation token source, this would build up over time.

## After

When we remove the `RunningScriptWrapper`, we now also dispose it (along with the `CancellationTokenSource` it is wrapping).

To make this safer, we moved creation of the `CancellationTokenSource` into `RunningScriptWrapper` itself.

This does mean that we create a `CancellationTokenSource` earlier. But 99.9% of the time we will end up going through the same code path as before. Ignoring exceptions, the only time we don't go down that path is if the Tentacle has restarted, and we loaded the state from disk. In which case, it doesn't matter if we create a `CancellationTokenSource`

# How to review this PR

Have a think about if there are any issues with how I've refactored it. I don't not believe there is, but am keen to find ways I didn't think of.

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle/Services/Scripts/ScriptServiceV2.cs
+++ b/source/Octopus.Tentacle/Services/Scripts/ScriptServiceV2.cs
@@ -72,11 +72,9 @@ namespace Octopus.Tentacle.Services.Scripts
                     runningScript.ScriptStateStore.Create();
                 }
 
-                var cancellationTokenSource = new CancellationTokenSource();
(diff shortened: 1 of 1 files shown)


APErebus (approved):
Excellent!

PR #633, opened 2023-10-19
Fix Solution for VS Users and Nuke

sburmanoctopus:
[sc-62554]

# Background

Every time a Visual Studio user would open the Tentacle solution, VS would alter it.

If the unsuspecting VS user was not paying attention, this change would often make its way into a PR, and the build server would no longer compile. Often leaving the poor VS developer confused about what they did wrong.

# Results

We applied to same fix as [this PR](https://github.com/OctopusDeploy/Halibut/pull/512) for Halibut.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


gb-8 (approved):
Verified and approved 👍

PR #634, opened 2023-10-23
Support async service implementations

APErebus:
# Background

As part of the Kubernetes Agent work, we want to start using async service implementations (not just the async client in Server).

There are two things we need to do to support this:

1. Change how service contracts are registered
2. Add a command-line flag/environment variable to enable the `AsyncHalibutFeature` on the Halibut Runtime.

# Results

Enabling the AsyncHalibutFeature means that the Halibut Runtime, when handling the incoming message, will choose to [use the async path](https://github.com/OctopusDeploy/Halibut/blob/main/source/Halibut/HalibutRuntime.cs#L196), where it will [try and execute service contract methods with an `Async` suffix](https://github.com/OctopusDeploy/Halibut/blob/main/source/Halibut/ServiceModel/ServiceInvoker.cs#L42)

I've also updated how services are registered with the `AutofacTypeRegistry`. Previously, we used implemented interface names on the types implementing the `[Service]` attribute and they resolved the service from the Autofac container by name. However, for async services, the contract being sent over the wire is the synchronous contract, when our async service implementation doesn't implement. Thus it won't be registered with the correct name. To fix this, I've changed the `ServiceAttribute` to take the `ServiceContractType`, forcing the developer to specific the type that is the contract.

An example async service + contract might be something like this:

```C#
public interface IScriptServiceV3Alpha
{
	ScriptStatusResponseV3Alpha StartScript(StartScriptCommandV3Alpha command);
}

public interface IAsyncScriptServiceV3Alpha
{
	Task<ScriptStatusResponseV3Alpha> StartScriptAsync(StartScriptCommandV3Alpha command, CancellationToken cancellationToken);
}

[Service(typeof(IScriptServiceV3Alpha))]
public class ScriptServiceV3Alpha : IAsyncScriptServiceV3Alpha
{
    public Task<ScriptStatusResponseV3Alpha> StartScriptAsync(StartScriptCommandV3Alpha command, CancellationToken cancellationToken)
	{
		// implementation here
	}
}
```

I'm happy to take feedback on this approach. It feels like making the contract more explicit is a good improvement.

Shortcut story: [sc-62808]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Octopus.Tentacle/Commands/RunAgentCommand.cs
```diff
@@ -70,6 +71,7 @@ public class RunAgentCommand : AbstractStandardCommand
                 // There's actually nothing to do here. The CommandHost should have already been determined before Start() was called
                 // This option is added to show help
             });
+            Options.Add("async-halibut", "Enables async Halibut support for running services asynchronously", _ => enableAsyncHalibut = true);
```
APErebus, line 74: I thought this might be a good thing to add, but if people feel like an environment variable is sufficient, then I can remove it.

source/Octopus.Tentacle/Variables/EnvironmentVariables.cs
```diff
@@ -20,5 +20,6 @@ public static class EnvironmentVariables
         public const string TentacleExecutablePath = "TentacleExecutablePath";
         public const string TentacleProgramDirectoryPath = "TentacleProgramDirectoryPath";
         public const string AgentProgramDirectoryPath = "AgentProgramDirectoryPath";
+        public const string TentacleEnableAsyncHalibut = "TentacleEnableAsyncHalibut";
```
APErebus, line 23: I don't love the naming scheme here... `OCTOPUS__Tentacle__EnableAsyncHalibut` probably fits with our current approach, but I wasn't sure if that was Server-specific

scme0:
This looks pretty good but my one concern is that, in the future when synchronous contracts are no longer required or used, will we continue to have to implement them just to use the name in the `[Service(typeof(...))]` Attribute? It feels like the Async contract should be able to "Just work" without needing to know about the synchronous equivalent.

We can probably kick that can down the road if you think it's not worth worrying about now but just a thought.

APErebus:
> This looks pretty good but my one concern is that, in the future when synchronous contracts are no longer required or used, will we continue to have to implement them just to use the name in the `[Service(typeof(...))]` Attribute? It feels like the Async contract should be able to "Just work" without needing to know about the synchronous equivalent.
> 
> We can probably kick that can down the road if you think it's not worth worrying about now but just a thought.

Right now there is some limitations where you can't have `Task<T>` in the contract definition. Maybe that's something that should change, but I don't know what the possible side-affect that would be.

I thought this might be the best approach while we still are mixing async & async services, but definitely happy to have feedback

source/Octopus.Tentacle/Communications/AutofacServiceFactory.cs
```diff
@@ -17,48 +17,48 @@ public class AutofacServiceFactory : IServiceFactory, IServiceRegistration, IDis
...
-            var interfaces = serviceType.GetInterfaces();
-            if (serviceType.IsInterface || interfaces.IsNullOrEmpty())
-            {
-                throw new InvalidServiceTypeException(serviceType);
-            }
-            
-            foreach (var face in interfaces)
+            //track the contract type to their known service implementations
+            //the contract type is the one that is sent across the wire (typically an IService sync contract)
+            knownServices[knownService.ServiceContractType.Name] = knownService;
```
MissedTheMark, line 43: Instead of having a dictionary of known services, can we just leverage the container and use [named services](https://autofac.readthedocs.io/en/latest/advanced/keyed-services.html#named-services) instead?
I do realise that you're following the pattern used prior to your change though.

MissedTheMark: Oh wait, looks like it's needed for the `RegisteredServiceTypes` property.  We'd need a common service interface to retrieve them without that dictionary.  Bugger.
Never mind then, as you were!

source/Octopus.Tentacle/Communications/TentacleCommunicationsModule.cs
```diff
@@ -22,11 +23,17 @@ protected override void Load(ContainerBuilder builder)
...
-                var halibutRuntime = new HalibutRuntimeBuilder()
+                var halibutRuntimeBuilder = new HalibutRuntimeBuilder()
                     .WithServiceFactory(services)
                     .WithServerCertificate(configuration.TentacleCertificate)
-                    .WithMessageSerializer(serializerBuilder => serializerBuilder.WithLegacyContractSupport())
-                    .Build();
+                    .WithMessageSerializer(serializerBuilder => serializerBuilder.WithLegacyContractSupport());
+
+                //if we have the environment variable to enable async halibut, enable it :)
+                if (Environment.GetEnvironmentVariable(EnvironmentVariables.TentacleEnableAsyncHalibut) == bool.TrueString)
```
sburmanoctopus, line 32: Maybe TryParse the bool instead of comparing to the string, as `"true" == bool.TrueString` is false

sburmanoctopus: Nit: my personal preference, if it works for you, it would be awesome to have brackets around the new line of the 'if' (old habits 😁 ). But I know this is mainly a Steve preference...

APErebus: hehe, I don't mind changing that. I have no strong opinion either way :)

APErebus: Fixed in 2b4689b6

source/Octopus.Tentacle/Communications/TentacleCommunicationsModule.cs
```diff
@@ -22,11 +23,19 @@ protected override void Load(ContainerBuilder builder)
...
                     .WithServiceFactory(services)
                     .WithServerCertificate(configuration.TentacleCertificate)
-                    .WithMessageSerializer(serializerBuilder => serializerBuilder.WithLegacyContractSupport())
-                    .Build();
+                    .WithMessageSerializer(serializerBuilder => serializerBuilder.WithLegacyContractSupport());
+
+                //if we have the environment variable to enable async halibut, enable it :)
+                if (bool.TryParse(Environment.GetEnvironmentVariable(EnvironmentVariables.TentacleEnableAsyncHalibut), out var isEnabled) && isEnabled)
+                {
+                    halibutRuntimeBuilder.WithAsyncHalibutFeatureEnabled();
```
sburmanoctopus, line 34: 🤔 I was just wondering whether we want to have a feature toggle at all, and just turn on Halibut async always.

We [have a story](https://app.shortcut.com/octopusdeploy/story/56576/make-tentacle-use-async-halibut) to do this work at some point.

But thinking it through, this is the safest way to unblock you, and when we get around to do that story, we can remove this environment variable then.

source/Octopus.Tentacle/Communications/AutofacServiceFactory.cs
```diff
@@ -17,48 +17,48 @@ public class AutofacServiceFactory : IServiceFactory, IServiceRegistration, IDis
...
                 .AsSelf()
                 .SingleInstance();
 
-            var interfaces = serviceType.GetInterfaces();
-            if (serviceType.IsInterface || interfaces.IsNullOrEmpty())
-            {
-                throw new InvalidServiceTypeException(serviceType);
-            }
-            
-            foreach (var face in interfaces)
```
sburmanoctopus, line 46: Yes, I saw this once before, how it's extra complication to allow one service to have multiple contracts (which we don't use). So I'm happy to see this unrequired loop go 😁

source/Octopus.Tentacle/Communications/AutofacServiceFactory.cs
```diff
@@ -17,48 +17,48 @@ public class AutofacServiceFactory : IServiceFactory, IServiceRegistration, IDis
...
+            //track the contract type to their known service implementations
+            //the contract type is the one that is sent across the wire (typically an IService sync contract)
+            knownServices[knownService.ServiceContractType.Name] = knownService;
+
+            //the implementation type doesn't need to implement the service contract, but it's _implied_ that it should have _some_ interfaces
+            if (knownService.ServiceImplementationType.IsInterface || knownService.ServiceImplementationType.GetInterfaces().IsNullOrEmpty())
```
sburmanoctopus, line 46: I feel like this validation would be better placed inside the `KnownService` constructor itself. That way, you can never have an invalid `KnownService`

source/Octopus.Tentacle/Communications/TentacleCommunicationsModule.cs
APErebus: Yeah, I wasn't sure what your plans are (outside of we are just turning it on), so I figured this was the safest then you can just trash that environment variable :)

sburmanoctopus (approved):
LGTM

source/Octopus.Tentacle/Communications/AutofacServiceFactory.cs
APErebus: @sburmanoctopus  I've made this change in 1a95c8d9

sburmanoctopus: Nice one!

While looking at that, I wonder if we should also validate that `serviceContractType` is an interface?

I don't want to keep moving the goalposts, so feel free to ignore this idea 🙂

APErebus: That's a good idea 👍

APErebus: @sburmanoctopus added expanded validation in 21429d6f and changed the exception type and cleaned up the message

sburmanoctopus: Nice, LGTM 🙂

PR #635, opened 2023-10-24
Add IScriptServiceV3Alpha contract

APErebus:
# Background

To support the Kubernetes Agent, which will be adding the ability to execute scripts via Kubernetes Jobs, we need to change the `StartScriptCommand` contract. As a result, we have decided to create a new `IScriptServiceV3Alpha`.

# Results

The main requirement for the contract change is that we need information from Server when running scripts as k8s Jobs. We need to know the container image and any external feed credentials/url.

To support this, I've added a new `IScriptExecutionContext ExecutionContext { get; }`  property to the `StartScriptCommandV3Alpha`.
There is one implementation of this interface, `LocalShellScriptExecutionContext` for the existing execution process (a local shell). In a future PR, there will be a `KubernetesJobScriptExecutionContext` (or something similiar). I didn't version this interface as the namespace it's in is versioned.

A future PR will be the initial implementation of this service, which will be a direct copy of the V2 service initially.

Shortcut story: [sc-62810]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Octopus.Tentacle.Contracts/ScriptServiceV3Alpha/ScriptCommandV3Alpha.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace Octopus.Tentacle.Contracts.ScriptServiceV3Alpha
+{
+    public abstract class ScriptCommandV3Alpha
```
APErebus, line 3: I create a base command to enforce that all commands must have the ScriptTicket property

source/Octopus.Tentacle.Contracts/ScriptServiceV3Alpha/StartScriptCommandV3Alpha.cs
```diff
@@ -0,0 +1,100 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Newtonsoft.Json;
+
+namespace Octopus.Tentacle.Contracts.ScriptServiceV3Alpha
+{
+    public class StartScriptCommandV3Alpha : ScriptCommandV3Alpha
```
sburmanoctopus, line 8: Just thinking out aloud, how much do we want to force customers to specify the `scriptExecutionContext`?

To start with the K8 version will be very rarely used. 

Even when it takes off any everyone thinks it's amazing, I would guess the local shell context is more commonly used.

This makes me wonder, should we make it easier to create these types of commands somehow?

Either by 
- making a constructor that doesn't take the context, and then default it to the local shell context
- some factory style methods?

We don't have to do any of this, just some thoughts 🙂

sburmanoctopus (approved):
LGTM

source/Octopus.Tentacle.Contracts/ScriptServiceV3Alpha/ScriptCommandV3Alpha.cs
APErebus: Yeah, that's fine. I don't feel strongly about it, so happy to defer.

source/Octopus.Tentacle.Contracts/ScriptServiceV3Alpha/StartScriptCommandV3Alpha.cs
APErebus: I believe in the server codebase we have some builders that are used for this purpose. In this case, we would default the context to the local shell execution, and then have methods for setting the required k8s context information, which would then build it up.

source/Octopus.Tentacle.Contracts/ScriptServiceV3Alpha/ScriptCommandV3Alpha.cs
APErebus: Addressed in c679e679

PR #637, opened 2023-10-24
Refactor TentacleClient & script orchestration

APErebus:
# Background

Currently, script execution orchestration in Tentacle.Client is handled by the `ScriptExecutorOrchestrator`. This checks the capabilities of the Tenacle (via the capabilities service, then uses this to decide which service to use), This code has a lot of `if/else` statements for the (current) two versions of the script service.

As part of the `ScriptServiceV3Alpha` work (starting with #635), this design doesn't seem overly scalable. This PR seeks to address this by refactoring the script orchestration into an orchestrator _per_ service version.

# Results

I have refactored the aforementioned `ScriptExecutorOrchestrator` into a `IScriptOrchestratorFactory`, which constructs a given `IScriptOrchestrator`, which has a single method, `ExecuteScript`.

Now, the `ScriptOrchestratorFactory` is the one that makes the service call to determine the capabilities of the Tentacle and given that, it then decides which service script orchestrator to call (e.g. `ScriptServiceV1Orchestrator` or `ScriptServiceV2Orchestrator`).

However, the original orchestrator code contained all the logic for monitoring a script, cancelling it if requested and completing it when it was done.

To avoid duplicating this logic into each orchestrator (so we can be sure they are behaving the same), each orchestrator inherits from an abstract `ObservingScriptOrchestrator` that has abstract methods for delegating the specific service calls for each service implementation.

One issue is how we mix the different contract command and response types. I handle this by have two generic parameters in the base orchestrator: `TStartCommand` and `TScriptStatusResponse` and a number of abstract `Map` methods.

The main `ExecuteScript` call takes the _latest_ version's `StartScriptCommand` (currently V2) and it's up to the specific service orchestrator to map that to a model they then send across the wire. The base orchestrator doesn't know or care about the specific types, it just passes around the generic args and the inheritors then have their expected types.

When the V3 service comes, we will just need to update the "public" contract of TentacleClient to use the V3 start command and update the orchestrators main command and then just tweak a couple of map commands.

Shortcut story: [sc-62862]

# How to review this PR

I'm keen on getting feedback about the approach of the factory & orchestrators

Quality :heavy_check_mark:

# Pre-requisites


acodrington (approved):
Nicely done, this looks great.

These changes will make it easier to introduce new versions of script service in the future, whilst not changing any behaviour for existing services. 👍🏻

PR #643, opened 2023-10-30
Fix namespace for Code Analysis Attributes

scme0:
# Background

While working on #project-k8s-agent, I noticed that the `NotNullWhenAttribute` wasn't working correctly. I would use the attribute but the resulting value wouldn't be seen as not null when it should have been.

I realised that it was because the code analysis attributes we have added into Tentacle (to give backward compatibilioty to net452) needs to have the same namespace as the original one. It looks like our other usages have been straddling two different implementations depending on if they are building with net6.0 or net452:
Screenshot 2023-10-30 at 17 30 54

# Results

By updating the implementation in our code base, there is no confusion between what attributes should be used. We can just use this implementation and it will work correctly with static code analysis.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


scme0:
Closing as not required

PR #647, opened 2023-11-02
Co-operatively cancel Tentacle Client Operations

nathanwoctopusdeploy:
# Background

[sc-58429]

Related to: https://github.com/OctopusDeploy/Halibut/pull/525

Before Async Halibut was introduced, an RPC the was transferring (reading or writing over a TCP Connection) could not be cancelled in Halibut, only an RPC that was connecting could be cancelled.

Async Halibut, along with https://github.com/OctopusDeploy/Halibut/pull/525 has added the ability to cancel RPC whether connecting or transferring. This allows the caller of Halibut to cancel an RPC with a cancellation token co-operatively.

Before Async Halibut, abandoning a Transferring RPC was the only way to "cancel" it. That is, run it as a different Task and walk away from the running Task if a cancellation was performed. While this method would allow the caller to regain control flow, it left the Task still executing and left IO operations on TCP Connection from Tentacle Client to Tentacle active.

# Results

## Before

Tentacle Client only supported the cancellation of connecting RPC. If an RPC was transferred, cancellation could not be performed, and the caller would have to wait for the RPC to complete.

Tentacle Client used the abandon technique for RPC Retry Timeouts, as it was the only reliable way to timeout the Retries and return control flow to the caller. 

## After

Tentacle Client will co-operatively cancel RPC to the Socket, stopping active reads and writes over TCP Connections.

Tentacle Client will co-operatively cancel RPC to the Socket when RPC Retries Timeout.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Octopus.Tentacle.Contracts/Octopus.Tentacle.Contracts.csproj
```diff
@@ -28,7 +28,7 @@
     </Otherwise>
   </Choose>
   <ItemGroup>
-    <PackageReference Include="Halibut" Version="7.0.144" />
+    <PackageReference Include="Halibut" Version="7.0.205-pull-525" />
```
nathanwoctopusdeploy, line 31: TODO - Update once https://github.com/OctopusDeploy/Halibut/pull/525 merges

source/Octopus.Tentacle.Client/Scripts/ObservingScriptOrchestrator.cs
```diff
@@ -73,8 +73,7 @@ public async Task<ScriptExecutionResult> ExecuteScript(StartScriptCommandV3Alpha
             {
                 if (scriptExecutionCancellationToken.IsCancellationRequested)
                 {
-                    // We don't want to cancel this operation as it is responsible for stopping the script executing on the Tentacle
-                    lastStatusResponse = await Cancel(lastStatusResponse, CancellationToken.None).ConfigureAwait(false);
+                    lastStatusResponse = await Cancel(lastStatusResponse, scriptExecutionCancellationToken).ConfigureAwait(false);
```
nathanwoctopusdeploy, line 76: The fact CancellationToken.None was passed into the method from this abstract class made it hard to understand the flow in the implementations. The actual cancellation token is now passed, and the implementations have been changed to use CancellationToken.None instead of the passed cancellation token. where appropriate. This makes reading the logic in the implementation much clearer.

source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
```diff
@@ -103,86 +109,68 @@ internal class RpcCallExecutor
...
         {
-            return await rpcCallNoRetriesHandler.ExecuteWithNoRetries(
-                    async ct =>
-                    {
-                        // Wrap the action in a task so it doesn't block on sync Halibut calls, and cancellation token is respected.
-                        return await Task.Run(async () => 
-                        {
-                            var rpcCallMetricsBuilder = RpcCallMetricsBuilder.StartWithoutRetries(rpcCall);
-                            var start = DateTimeOffset.UtcNow;
+            // Wrap the action in a task so it doesn't block on sync Halibut calls, and cancellation token is respected.
```
sburmanoctopus, line 115: Just wondering, is this comment still legit now? I.e., can we remove it, as well as wrapping it in a task?

It might be more retesting than we have appetite for, but thought I'd ask.

source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
```diff
@@ -103,86 +109,68 @@ internal class RpcCallExecutor
...
         {
-            await rpcCallNoRetriesHandler.ExecuteWithNoRetries(
-                    async ct =>
-                    {
-                        // Wrap the action in a task so it doesn't block on sync Halibut calls, and cancellation token is respected.
-                        await Task.Run(async () =>
-                        {
-                            var rpcCallMetricsBuilder = RpcCallMetricsBuilder.StartWithoutRetries(rpcCall);
-                            var start = DateTimeOffset.UtcNow;
+            // Wrap the action in a task so it doesn't block on sync Halibut calls, and cancellation token is respected.
```
sburmanoctopus, line 150: Same question as above.

source/Octopus.Tentacle.Client/Scripts/HalibutClientExceptionExtensionMethods.cs
```diff
@@ -0,0 +1,28 @@
...
+    public static class HalibutClientExceptionExtensionMethods
+    {
+        public static bool IsHalibutWrappedConnectingRequestCancelledException(this HalibutClientException? ex)
+        {
+            if (ex is null)
+            {
+                return false;
+            }
+
+            if (ex.Message.Contains("The Request was cancelled while Connecting"))
```
sburmanoctopus, line 15: I suspect you've explored all this, but is there not a type-safe way to detect this? Or even a string in the message (or other property) that has a text-based version of the type instead?

Possibly just wishful thinking 🙂

source/Octopus.Tentacle.Client/Scripts/ScriptServiceV2Orchestrator.cs
```diff
@@ -76,163 +77,125 @@ protected override async Task<ScriptStatusResponseV2> StartScript(StartScriptCom
...
+            catch (Exception e) when (scriptExecutionCancellationToken.IsCancellationRequested)
             {
-                // If we are not retrying and we managed to cancel execution it means the request was never sent so we can safely walk away from it.
-                if (e is not OperationAbandonedException && startScriptCallCount <= 1)
-                {
-                    throw;
-                }
+                // If the call to StartScript is in-flight (being transferred to the Service) or we are retrying the StartScript call
+                // then we do not know if the script has been started or not on Tentacle so need to call CancelScript and CompleteScript
+                var startScriptCallIsBeingRetries = startScriptCallCount > 1;
```
sburmanoctopus, line 97: `startScriptCallIsBeingRetries` or `startScriptCallIsBeingRetried`?

source/Octopus.Tentacle.Client/Scripts/ScriptServiceV3AlphaOrchestrator.cs
```diff
@@ -64,157 +65,125 @@ protected override async Task<ScriptStatusResponseV3Alpha> StartScript(StartScri
...
+            catch (Exception e) when (scriptExecutionCancellationToken.IsCancellationRequested)
             {
-                // If we are not retrying and we managed to cancel execution it means the request was never sent so we can safely walk away from it.
-                if (e is not OperationAbandonedException && startScriptCallCount <= 1)
-                {
-                    throw;
-                }
+                // If the call to StartScript is in-flight (being transferred to the Service) or we are retrying the StartScript call
+                // then we do not know if the script has been started or not on Tentacle so need to call CancelScript and CompleteScript
+                var startScriptCallIsBeingRetries = startScriptCallCount > 1;
```
sburmanoctopus, line 85: Same as before :)

source/Octopus.Tentacle.Tests.Integration/ClientFileTransferRetriesTimeout.cs
```diff
@@ -42,11 +42,11 @@ public async Task WhenRpcRetriesTimeOut_DuringUploadFile_TheRpcCallIsCancelled(T
                     .RecordMethodUsages<IAsyncClientFileTransferService>(out var methodUsages)
                     .HookServiceMethod<IAsyncClientFileTransferService>(nameof(IAsyncClientFileTransferService.UploadFileAsync),
                         async (_,_) =>
-                        {
-                            await tcpConnectionUtilities.RestartTcpConnection();
-
+                        {                            
                             if (methodUsages.For(nameof(IAsyncClientFileTransferService.UploadFileAsync)).LastException is null)
                             {
+                                // Ensure the is an active connection so it can be killed correctly
```
sburmanoctopus, line 48: `the` => `there`?

sburmanoctopus: (and below)

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionRetriesTimeout.cs
```diff
@@ -43,11 +43,11 @@ public async Task WhenRpcRetriesTimeOut_DuringGetCapabilities_TheRpcCallIsCancel
                     .HookServiceMethod<IAsyncClientCapabilitiesServiceV2>(nameof(IAsyncClientCapabilitiesServiceV2.GetCapabilitiesAsync),
                         async (_,_) =>
                         {
-                            await tcpConnectionUtilities.RestartTcpConnection();
-
                             // Kill the first GetCapabilities call to force the rpc call into retries
                             if (capabilitiesMethodUsages.For(nameof(IAsyncClientCapabilitiesServiceV2.GetCapabilitiesAsync)).LastException is null)
                             {
+                                // Ensure the is an active connection so it can be killed correctly
```
sburmanoctopus, line 49: `the` vs `there`?

source/Octopus.Tentacle.Tests.Integration/Support/ExtensionMethods/FluentAssertionExtensionMethods.cs
```diff
@@ -1,15 +1,36 @@
...
+            {
+                return should.Match(x => x is ConnectingRequestCancelledException || IsHalibutWrappedConnectingRequestCancelledException(x as HalibutClientException));
+            }
+
+            return should.Match(x => x is TransferringRequestCancelledException || IsHalibutWrappedTransferringRequestCancelledException(x as HalibutClientException));
+        }
+
+        static bool IsHalibutWrappedConnectingRequestCancelledException(HalibutClientException? halibutClientException)
+        {
+            return halibutClientException != null && halibutClientException.Message.Contains("The Request was cancelled while Connecting");
```
sburmanoctopus, line 23: Like before, if possible, is there any other property that contains maybe the type as a string? Something else to compare against other than the message 🙂

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionCanBeCancelledWhenRetriesAreDisabled.cs
```diff
@@ -185,42 +160,34 @@ public async Task DuringStartScript_ScriptExecutionCanBeCancelled(TentacleConfig
...
-                case ExpectedFlow.AbandonRpcThenCancelScriptThenCompleteScript:
-                    recordedUsages.For(nameof(IAsyncClientScriptServiceV2.CancelScriptAsync)).Started.Should().BeGreaterOrEqualTo(1);
-                    recordedUsages.For(nameof(IAsyncClientScriptServiceV2.CompleteScriptAsync)).Started.Should().Be(1);
-                    break;
-                default:
-                    throw new NotSupportedException();
+                // Assert the CancelScript and CompleteScript flow happened fairly quickly
+                actualException.Should().BeScriptExecutionCancelledException();
+
+                // Assert the CancelScript and CompleteScript flow happened fairly quickly
```
sburmanoctopus, line 184: This feels like a repeat of the comment two lines above.

Actually, the comments in the `Connecting` state are kind of repeated as well (but less so)

sburmanoctopus (approved):
LGTM

source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
nathanwoctopusdeploy: The comment is out of date, I'll try and remove the Task.Run wrappers and see what happens. Hopefully they are now redundant!!

source/Octopus.Tentacle.Client/Scripts/HalibutClientExceptionExtensionMethods.cs
nathanwoctopusdeploy: Yeh this isn't nice code!! I'll revisit. At a minimum, this can be moved over to Halibut and tested to make sure the behaviour doesn't change. I'll see if there is an existing way to surface something more type safe!

nathanwoctopusdeploy: I'll leave this as is and revisit with a better solution in another PR. We have tests that will capture if this isn't working as expected so it shouldn't be the source of bugs.

source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
nathanwoctopusdeploy: Green build with these removed, they are redundant 🥳

source/Octopus.Tentacle.Client/Scripts/ScriptServiceV3AlphaOrchestrator.cs
nathanwoctopusdeploy: Done

source/Octopus.Tentacle.Client/Scripts/ScriptServiceV2Orchestrator.cs
nathanwoctopusdeploy: Done

source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
nathanwoctopusdeploy: Done

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionCanBeCancelledWhenRetriesAreDisabled.cs
nathanwoctopusdeploy: Done

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionRetriesTimeout.cs
nathanwoctopusdeploy: Done

PR #649, opened 2023-11-05
Add new decorators based on DispatchProxyAsync

APErebus:
# Background

When doing a branch to implement `SctiptServiceV3Alpha`, I realised the integration tests were quite heavily couple to specific versions of the services and also required quite a lot of boilerplate code to set up the various decorators for the tests.

This PR introduces proxy-based decorators using `DispatchProxyAsync`.

This doesn't update all the tests, just a selection to validate the new decorators work correctly. A future PR will change the rest of the tests over.

# Results

## Dispatch Proxy decorators
I have introduced 3 new proxy-based based decorators implementing on a `ServiceProxy` abstract based class (which inherits from `DispatchProxyAsync`

- `MethodLoggingProxyDecorator` - Logs when a service method is started and completed and throws an exception
- `MethodTracingProxyDecorator` - "Traces" how many times a method is started, completed and what the last exception was
- `MethodInvocationHookProxyDecorator` - Allows for pre and post invocation hooks to be run before or after the service method execution

All services have the `MethodLogging` decorator attached be default for easier debugging if something is being run unexpectedly.

There are new extension methods on the `TentacleServicesDecoratorBuilder` to use these new extension methods. ~Currently, they support both Async & Sync service test cases by using overloads that take 2 generic params. The exception is the invocation hooks extension method as the hooks themselves are generically typed, so can't be shared between the sync/async services.~ As this is now merging after sync services have been removed, they only support async services now.

If any of these extension methods are used, then it swaps to use the proxy decorators and not use the other existing decorators.

## TcpConnectionUtilities

There are a number of tests where we need/want to do various things to the halibut TCP connection, such as forcing to exist or making sure the polling queue doesn't send work to disconnected tentacles. Previously, this was implemented as extension  method on various services.

To simplify this (and also take the execution of any service method outside of the decorator pipeline), I implemented a new `TcpConnectionUtilities` class, which can be used via the `WithTcpConnectionUtilities` method on the `ClientAndTentacleBuilder`.

## Updated tests

I updated 2 test fixtures (`ScriptServiceV2` and `ClientFileTransferRetriesTimeout.cs`) and one test case from the `ClientScriptExecutionCanBeCancelledWhenRetriesAreDisabled.cs` fixture to test all the different cases (as well as different hooks) and verify they work correctly.

# How to review this PR

I don't love the `MethodTracing` name, but after discussion with @nathanwoctopusdeploy , wanted to avoid the use of "metrics".

Quality :heavy_check_mark:

# Pre-requisites


source/Octopus.Tentacle.Tests.Integration/Util/Builders/Decorators/Proxies/MethodTracingProxyDecorator.cs
```diff
@@ -0,0 +1,66 @@
+﻿using System;
+using System.Reflection;
+using System.Threading.Tasks;
+
+namespace Octopus.Tentacle.Tests.Integration.Util.Builders.Decorators.Proxies
+{
+    public class MethodTracingProxyDecorator : ServiceProxy
```
APErebus, line 7: As mentioned, I don't love this name, but don't have a better one in mind.

source/Octopus.Tentacle.Tests.Integration/Util/Builders/Decorators/Proxies/MethodTracingStats.cs
```diff
@@ -0,0 +1,69 @@
...
+        public  void RecordCallException(MethodInfo targetMethod, Exception exception)
+        {
+            var stats = GetMethodStats(targetMethod);
+            stats.RecordException(exception);
+        }
+
+        TracedMethodStats GetMethodStats(string methodName) => trackedMethods.GetOrAdd(NormalizeMethodName(methodName), _ => new Lazy<TracedMethodStats>(() => new TracedMethodStats())).Value;
+        TracedMethodStats GetMethodStats(MethodInfo targetMethod) => GetMethodStats(targetMethod.Name);
+
+        static string NormalizeMethodName(string methodName)
```
APErebus, line 35: Normalising just for sync/async service method names (as the extension method uses the same object for both). Can be removed when sync services are removed

source/Octopus.Tentacle.Tests.Integration/Util/Builders/Decorators/Proxies/MethodTracingStats.cs
```diff
@@ -0,0 +1,69 @@
...
+        public Exception? LastException { get; set; }
+
+        public void RecordStarted() => Interlocked.Increment(ref started);
+
+        public void RecordCompleted() => Interlocked.Increment(ref completed);
+
+        public void RecordException(Exception exception) => LastException = exception;
+    }
+
+    public interface IRecordedTracedMethodStats
```
APErebus, line 63: The interfaces are so that the contract in the tests are only for the values and can't accidentally record extra things etc..

source/Octopus.Tentacle.Tests.Integration/Util/Builders/Decorators/Proxies/ServiceProxy.cs
```diff
@@ -0,0 +1,121 @@
...
+        public override object Invoke(MethodInfo method, object[] args)
+        {
+            EnsureTargetServiceNotNull();
+            try
+            {
+                OnStartingInvocation(method);
+
+                return method.Invoke(TargetService, args)!;
+            }
+            catch (TargetInvocationException e)
```
APErebus, line 26: If the invocation throws an exception, it gets wrapped in a`TargetInvocationException`, which we then unwrap

source/Octopus.Tentacle.Tests.Integration/Util/Builders/Decorators/Proxies/ServiceProxy.cs
```diff
@@ -0,0 +1,121 @@
...
+            {
+                OnInvocationException(method, e.InnerException!);
+
+                //we need to unwrap the TargetInvocationException
+                ExceptionDispatchInfo.Capture(e.InnerException).Throw();
```
APErebus, line 31: This throws the inner exception with a stack trace that doesn't get restarted from here (as it would be if we did `throw e.InnerException`)

scme0 (commented):
This generally looks good to me, but I'll let the others give the final approval as it's a bit outside of my wheelhouse. 👍

APErebus:
There is a follow up PR (#650) that adds support for tracing against the known version of script service running on the tested tentacle

APErebus:
Once these two are merged, a PR to completely change the tests to use these new decorators. Then the SSv3 implementation can be added to `TentacleClient`

source/Octopus.Tentacle.Tests.Integration/Util/Builders/Decorators/Proxies/ServiceProxy.cs
```diff
@@ -0,0 +1,121 @@
...
+            }
+            catch (Exception e)
+            {
+                OnInvocationException(method, e);
+                throw;
+            }
```
david-staniec-octopus, line 69: Does this need a finally here? This looks like one is intended

source/Octopus.Tentacle.Tests.Integration/Util/Builders/Decorators/Proxies/MethodLoggingProxyDecorator.cs
```diff
@@ -0,0 +1,55 @@
...
+
+            return proxiedService;
+        }
+
+        protected override void OnStartingInvocation(MethodInfo targetMethod)
+        {
+            logger.Information("{ServiceName:l}.{MethodName:l}() started", serviceTypeName, targetMethod.Name);
+        }
+
+        protected override Task OnStartingInvocationAsync(MethodInfo targetMethod)
```
david-staniec-octopus, line 33: Is there a reason this isn't an async method? I noticed you declare the async methods in MethodInvocationHookProxyDecorator using the async keyword.

source/Octopus.Tentacle.Tests.Integration/Util/Builders/Decorators/Proxies/MethodLoggingProxyDecorator.cs
```diff
@@ -0,0 +1,55 @@
...
+        {
+            logger.Information("{ServiceName:l}.{MethodName:l}() started", serviceTypeName, targetMethod.Name);
```
david-staniec-octopus, line 35: Can the async methods in this class call the sync method since the log entries are identical? If we don't want to call sync from async then a private method could be acceptable but may cause bloat?

source/Octopus.Tentacle.Tests.Integration/Util/Builders/Decorators/Proxies/MethodTracingProxyDecorator.cs
```diff
@@ -0,0 +1,66 @@
...
+
+            return proxiedService;
+        }
+
+        protected override void OnStartingInvocation(MethodInfo targetMethod)
+        {
+            tracingStats.RecordCallStart(targetMethod);
+        }
+
+        protected override Task OnStartingInvocationAsync(MethodInfo targetMethod)
```
david-staniec-octopus, line 44: Is there a reason this isn't an async method? I noticed you declare the async methods in MethodInvocationHookProxyDecorator using the async keyword.

david-staniec-octopus (commented):
We're happy with the changes here, and we're fine with the naming too. Left a couple of questions and potentially a bug with a missing `finally` keyword.

source/Octopus.Tentacle.Tests.Integration/Util/Builders/Decorators/Proxies/MethodInvocationHookProxyDecorator.cs
```diff
@@ -0,0 +1,71 @@
...
+            {
+                await postInvocation((TService)TargetService);
+            }
+        }
+
+        protected override void OnInvocationException(MethodInfo targetMethod, Exception exception)
+        {
+        }
+
+        bool IsMethodInteresting(MemberInfo invocationMethod)
```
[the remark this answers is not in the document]
APErebus: Yep, makes sense 👍

source/Octopus.Tentacle.Tests.Integration/Util/Builders/Decorators/Proxies/MethodLoggingProxyDecorator.cs
APErebus: Yeah, I had thought about that... I had separated them as my thought was that when we remove the sync services from Tentacle, that the non-async `OnStartingInvocation`/`OnCompletingInvocation` would go away completely (as there will be no non-async calls to decorated services).

source/Octopus.Tentacle.Tests.Integration/Util/Builders/Decorators/Proxies/ServiceProxy.cs
APErebus: ooops, yep. thanks!

source/Octopus.Tentacle.Tests.Integration/Util/Builders/Decorators/Proxies/MethodLoggingProxyDecorator.cs
APErebus: Yeah, my personal preference is to not decorate a method with the `async` keyword if I'm not using an `await` inside the method. In those cases, I just return a completed task.

This is just a personal preference, so if this is something you'd like changed to `async Task`, then I can do that :)

source/Octopus.Tentacle.Tests.Integration/Util/Builders/Decorators/Proxies/MethodInvocationHookProxyDecorator.cs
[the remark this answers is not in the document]
APErebus: done in dc9c26dc

source/Octopus.Tentacle.Tests.Integration/Util/Builders/Decorators/Proxies/MethodTracingProxyDecorator.cs
acodrington: How about 'Usage'?

`MethodTracingProxyDecorator` -> `MethodUsageStatsProxyDecorator`
`RecordedMethodTracingStats` -> `RecordedMethodUsageStats`
`MethodTracingStats` -> `MethodUsageStats`
...and so on?

acodrington (approved):
Changes look good, removing sync implementations certainly simplifies things!

I left a naming suggestion in the comments, but otherwise no further notes from me.

source/Octopus.Tentacle.Tests.Integration/Util/Builders/Decorators/Proxies/MethodTracingProxyDecorator.cs
APErebus: Yeah, I like it. Thanks @acodrington

APErebus: I chose to drop the `Stats` part, so it's just `RecordedMethodUsages` and `MethodUsages` etc

PR #650, opened 2023-11-06
Test against known latest script service on Tentacle

APErebus:
# Background

As we test more versions of TentacleClient against older versions of Tentacle, we need to adjust the tests to know which script service is running on the target version of tentacle, so we can correctly test that version.

# Results

Adds a new static mapping of Tentacle Version -> Latest script service type supported on that version of tentacle. This is then included in the `TentacleConfigurationTestCase`.

I added an overload to the `RecordMethodUsages` extension method to create the dynamic proxy given a type (rather than a generic parameter).

I updated 2 tests to use this new logic.

Shortcut story: [sc-62862]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionIsolationMutex.cs
+++ b/source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionIsolationMutex.cs
@@ -23,7 +23,7 @@ namespace Octopus.Tentacle.Tests.Integration
         {
             await using var clientTentacle = await tentacleConfigurationTestCase.CreateBuilder()
                 .WithTentacleServiceDecorator(new TentacleServiceDecoratorBuilder()
-                    .TraceService(tentacleConfigurationTestCase.LatestScriptServiceTypes, out var tracingStats)
+                    .TraceService(tentacleConfigurationTestCase.LatestScriptServiceType, out var tracingStats)
                     .Build())
                 .Build(CancellationToken);
 
--- a/source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionIsolationMutex.cs
+++ b/source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionIsolationMutex.cs
@@ -23,7 +23,7 @@ namespace Octopus.Tentacle.Tests.Integration
         {
             await using var clientTentacle = await tentacleConfigurationTestCase.CreateBuilder()
                 .WithTentacleServiceDecorator(new TentacleServiceDecoratorBuilder()
-                    .TraceService(tentacleConfigurationTestCase.LatestScriptServiceType, out var tracingStats)
+                    .RecordMethodUsage(tentacleConfigurationTestCase.LatestScriptServiceType, out var tracingStats)
                     .Build())
                 .Build(CancellationToken);
 
--- a/source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionIsolationMutex.cs
+++ b/source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionIsolationMutex.cs
@@ -6,6 +6,7 @@ using FluentAssertions;
 using NUnit.Framework;
 using Octopus.Tentacle.CommonTestUtils.Builders;
 using Octopus.Tentacle.Contracts;
+using Octopus.Tentacle.Contracts.ClientServices;
 using Octopus.Tentacle.Tests.Integration.Support;
 using Octopus.Tentacle.Tests.Integration.Util;
 using Octopus.Tentacle.Tests.Integration.Util.Builders;
@@ -22,8 +23,7 @@ namespace Octopus.Tentacle.Tests.Integration
         {
             await using var clientTentacle = await tentacleConfigurationTestCase.CreateBuilder()
                 .WithTentacleServiceDecorator(new TentacleServiceDecoratorBuilder()
-                    .CountCallsToScriptServiceV2(out var scriptServiceV2CallCounts)
-                    .CountCallsToScriptService(out var scriptServiceCallCounts)
+                    .TraceService(tentacleConfigurationTestCase.LatestScriptServiceTypes, out var tracingStats)
                     .Build())
                 .Build(CancellationToken);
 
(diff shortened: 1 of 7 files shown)


source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutorObservesScriptObserverBackoffStrategy.cs
```diff
@@ -30,7 +31,11 @@ public async Task TheScriptObserverBackoffShouldBeRespected(TentacleConfiguratio
 
             var (_, logs) = await clientTentacle.TentacleClient.ExecuteScript(startScriptCommand, CancellationToken);
 
-            scriptServiceV2CallCounts.GetStatusCallCountStarted.Should().BeLessThan(3);
+            tracingStats.For(nameof(IAsyncClientScriptServiceV2.GetStatusAsync)).Started
```
APErebus, line 34: Turns out this test wasn't testing the `ScriptServiceV1` as there was no `CountCalls` so it was always returning `0`, which isn't a great test :)

I've updated this test to make sure there GetStatus call's are greater than 0 and it all passes :)

source/Octopus.Tentacle.Tests.Integration/Support/TentacleConfigurationsAttribute.cs
```diff
@@ -18,13 +20,28 @@ public class TentacleConfigurationsAttribute : TentacleTestCaseSourceAttribute
...
         {
         }
     }
 
     static class TentacleConfigurationTestCases
     {
+        static readonly Type[] ScriptServiceV2Types = { typeof(IClientScriptServiceV2), typeof(IAsyncClientScriptServiceV2) };
+        static readonly Type[] ScriptServiceV1Types = { typeof(IClientScriptService), typeof(IAsyncClientScriptService) };
+
+        static readonly Type[] CurrentScriptServiceTypes = ScriptServiceV2Types;
```
APErebus, line 33: Can't have a null dictionary key (as the current version is `null`, so just have a preset value for the current version

source/Octopus.Tentacle.Tests.Integration/Support/TentacleConfigurationsAttribute.cs
```diff
@@ -18,13 +20,28 @@ public class TentacleConfigurationsAttribute : TentacleTestCaseSourceAttribute
...
-                new object[] {testCommonVersions, testCapabilitiesServiceVersions, testNoCapabilitiesServiceVersions, testScriptIsolationLevelVersions, testDefaultTentacleRuntimeOnly, additionalParameterTypes})
+                new object[] { testCommonVersions, testCapabilitiesServiceVersions, testNoCapabilitiesServiceVersions, testScriptIsolationLevelVersions, testDefaultTentacleRuntimeOnly, additionalParameterTypes })
         {
         }
     }
 
     static class TentacleConfigurationTestCases
     {
+        static readonly Type[] ScriptServiceV2Types = { typeof(IClientScriptServiceV2), typeof(IAsyncClientScriptServiceV2) };
+        static readonly Type[] ScriptServiceV1Types = { typeof(IClientScriptService), typeof(IAsyncClientScriptService) };
```
APErebus, line 31: When sync services are removed, this can be changed back to a single `Type`

source/Octopus.Tentacle.Tests.Integration/Util/Builders/Decorators/TentacleServiceDecoratorBuilder.cs
```diff
@@ -9,7 +9,7 @@
 namespace Octopus.Tentacle.Tests.Integration.Util.Builders.Decorators
 {
     public delegate T Decorator<T>(T service);
-    public delegate object WrappedProxyDecoratorFactory(object service);
+    public delegate object ProxyDecoratorFactory(object service);
```
APErebus, line 12: Renamed for clarity

scme0 (approved):
Looks good! Just one minor comment from me. Happy if you want input from server-at-scale otherwise LGTM 🚀

source/Octopus.Tentacle.Tests.Integration/Support/TentacleConfigurationTestCase.cs
```diff
@@ -9,17 +9,23 @@ public class TentacleConfigurationTestCase
         public TentacleType TentacleType { get; }
         public TentacleRuntime TentacleRuntime { get; }
         public Version? Version { get; }
+        /// <summary>
+        /// An array of <see cref="Type">Types</see> of the latest script service available on the tentacle version
```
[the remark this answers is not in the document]
scme0: It says "An array of" but it's not an array :)

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionIsolationMutex.cs
```diff
@@ -22,8 +23,7 @@ public async Task ScriptIsolationMutexFull_EnsuresTwoDifferentScriptsDontRunAtTh
         {
             await using var clientTentacle = await tentacleConfigurationTestCase.CreateBuilder()
                 .WithTentacleServiceDecorator(new TentacleServiceDecoratorBuilder()
-                    .CountCallsToScriptServiceV2(out var scriptServiceV2CallCounts)
-                    .CountCallsToScriptService(out var scriptServiceCallCounts)
+                    .RecordMethodUsage(tentacleConfigurationTestCase.LatestScriptServiceType, out var tracingStats)
```
david-staniec-octopus, line 26: We're not sure why the test case configuration can define the version of the script service to use, but the test still needs to define the version of the command to build. We were getting the impression as we read through the changes that this was heading in the direction that the test would be agnostic of the setup required for different script service / command version combinations. Is that not the case?

source/Octopus.Tentacle.Tests.Integration/Support/TentacleConfigurationTestCase.cs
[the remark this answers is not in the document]
APErebus: 🤦

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionIsolationMutex.cs
APErebus: @david-staniec-octopus The test will always need to use the command from the latest version of the script service. TentacleClient always needs to take the latest version of the `StartScript` command, but internally it may use a different script service version, depending on the capabilities of the client tentacle.

We need the test case to identify what is the latest version of the script service that the tentacle version is running so that we can correctly record the method usage on the correct service (as it'll change depending on the dependent tentacle version).

Happy to clarify on a call if that helps

acodrington (approved):
Looks good, my questions have been discussed in the comments, so good to go. :shipit:

PR #653, opened 2023-11-07
When a scripts tentacle log is corrupt, return all valid logs up to the corrupted log line.

LukeButters:
# Background
[SC-63899]

Fixes a bug where calls to `GetStatus` would fail when a script log line was corrupt e.g. incomplete because tentacle was killed while writing the logs for a running script.

# Results
Fixes https://github.com/OctopusDeploy/OctopusTentacle/issues/655

## Before

Previously the call to `GetStatus` would fail and no further logs could be read.

## After

The call to `GetStatus` will succeed, and we will read up to but not including the corrupt log line.

We do report the fact that the log was corrupt.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Tests/Integration/ScriptLogFixture.cs
+++ b/source/Octopus.Tentacle.Tests/Integration/ScriptLogFixture.cs
@@ -75,13 +75,13 @@ namespace Octopus.Tentacle.Tests.Integration
                 appender.WriteOutput(ProcessOutputSource.StdOut, "World");
             }
 
-            using (var fs = File.Open(logFile, FileMode.Open))
+            using (var logFileStream = File.Open(logFile, FileMode.Open))
             {
-                FileInfo fi = new FileInfo(logFile);
-                fs.SetLength(fi.Length - 10);
+                var logFileInfo = new FileInfo(logFile);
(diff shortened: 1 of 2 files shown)


source/Octopus.Tentacle.Tests/Integration/ScriptLogFixture.cs
```diff
@@ -65,6 +65,27 @@ public void ShouldAppend()
...
+        [Test]
+        public void ShouldHandleIncompleteLine()
+        {
+            using (var appender = sut.CreateWriter())
+            {
+                appender.WriteOutput(ProcessOutputSource.StdOut, "Hello");
+                appender.WriteOutput(ProcessOutputSource.StdOut, "World");
+            }
+
+            using (var fs = File.Open(logFile, FileMode.Open))
```
sburmanoctopus, line 78: 2 nits:
- I still like using the full name. I.e. fs => fileStream
- No `var` for `FileInfo`

sburmanoctopus (approved):
LGTM

LukeButters:
TODO: create a github issue so this has meaningful changes.

PR #669, opened 2023-11-15
Update TentacleClient to use ScriptServiceV3Alpha if available

APErebus:
# Background

Adds support for `ScriptServiceV3Alpha` to TentacleClient. To allow this service to be disabled/ignored, there is a new flag on `TentacleClientOptions.DisableScriptServiceV3Alpha` to explicitly disable this.

At this stage, this contains the same logic as `ScriptServiceV2`, but this will start to change as we add support for Kubernetes Jobs

# Results

Adds a new `ScriptServiceV3AlphaOrchestrator` which is used when the tentacle returns `IScriptServiceV3Alpha` from the GetCapabilities call.

Updates the main entrypoint for executing a script in TentacleClient to take a `StartScriptCommandV3Alpha`. This will require a change to Server to use this.

A lot of the changes in the Tests are just change to use `StartScriptCommandV3AlphaBuilder`

Shortcut story: [sc-62862]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Client/Scripts/ScriptServiceV3AlphaOrchestrator.cs
+++ b/source/Octopus.Tentacle.Client/Scripts/ScriptServiceV3AlphaOrchestrator.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
@@ -10,22 +10,21 @@ using Octopus.Tentacle.Client.Retries;
 using Octopus.Tentacle.Contracts;
 using Octopus.Tentacle.Contracts.ClientServices;
 using Octopus.Tentacle.Contracts.Observability;
-using Octopus.Tentacle.Contracts.ScriptServiceV2;
 using Octopus.Tentacle.Contracts.ScriptServiceV3Alpha;
 using ILog = Octopus.Diagnostics.ILog;
 
 namespace Octopus.Tentacle.Client.Scripts
 {
-    class ScriptServiceV2Orchestrator : ObservingScriptOrchestrator<StartScriptCommandV2, ScriptStatusResponseV2>
+    class ScriptServiceV3AlphaOrchestrator : ObservingScriptOrchestrator<StartScriptCommandV3Alpha, ScriptStatusResponseV3Alpha>
     {
-        readonly IAsyncClientScriptServiceV2 clientScriptServiceV2;
+        readonly IAsyncClientScriptServiceV3Alpha clientScriptServiceV3Alpha;
         readonly RpcCallExecutor rpcCallExecutor;
         readonly ClientOperationMetricsBuilder clientOperationMetricsBuilder;
         readonly TimeSpan onCancellationAbandonCompleteScriptAfter;
         readonly ILog logger;
 
-        public ScriptServiceV2Orchestrator(
-            IAsyncClientScriptServiceV2 clientScriptServiceV2,
+        public ScriptServiceV3AlphaOrchestrator(
+            IAsyncClientScriptServiceV3Alpha clientScriptServiceV3Alpha,
             IScriptObserverBackoffStrategy scriptObserverBackOffStrategy,
             RpcCallExecutor rpcCallExecutor,
             ClientOperationMetricsBuilder clientOperationMetricsBuilder,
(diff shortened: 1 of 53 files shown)


source/Octopus.Tentacle.Tests.Integration/ScriptServiceV3AlphaIntegrationTest.cs
```diff
@@ -0,0 +1,180 @@
...
+using Octopus.Tentacle.Contracts.ClientServices;
+using Octopus.Tentacle.Tests.Integration.Support;
+using Octopus.Tentacle.Tests.Integration.Util;
+using Octopus.Tentacle.Tests.Integration.Util.Builders;
+using Octopus.Tentacle.Tests.Integration.Util.Builders.Decorators;
+
+namespace Octopus.Tentacle.Tests.Integration
+{
+    [IntegrationTestTimeout]
+    public class ScriptServiceV3AlphaIntegrationTest : IntegrationTest
```
APErebus, line 18: This is a direct copy of the `SCriptServiceV2IntegrationTest` above

source/Octopus.Tentacle/Services/Capabilities/CapabilitiesServiceV2.cs
```diff
@@ -13,7 +15,7 @@ public class CapabilitiesServiceV2 : IAsyncCapabilitiesServiceV2
         public async Task<CapabilitiesResponseV2> GetCapabilitiesAsync(CancellationToken cancellationToken)
         {
             await Task.CompletedTask;
-            return new CapabilitiesResponseV2(new List<string>() {nameof(IScriptService), nameof(IFileTransferService), nameof(IScriptServiceV2)});
+            return new CapabilitiesResponseV2(new List<string>() {nameof(IScriptService), nameof(IFileTransferService), nameof(IScriptServiceV2), nameof(IScriptServiceV3Alpha)});
```
APErebus, line 18: Expose `IScriptServiceV3Alpha` via CapabilitiesService

source/Octopus.Tentacle.Client/Scripts/ScriptServiceV3AlphaOrchestrator.cs
```diff
@@ -0,0 +1,226 @@
...
+using Octopus.Tentacle.Client.Retries;
+using Octopus.Tentacle.Contracts;
+using Octopus.Tentacle.Contracts.ClientServices;
+using Octopus.Tentacle.Contracts.Observability;
+using Octopus.Tentacle.Contracts.ScriptServiceV3Alpha;
+using ILog = Octopus.Diagnostics.ILog;
+
+namespace Octopus.Tentacle.Client.Scripts
+{
+    class ScriptServiceV3AlphaOrchestrator : ObservingScriptOrchestrator<StartScriptCommandV3Alpha, ScriptStatusResponseV3Alpha>
```
APErebus, line 18: This is basically a copy of the `ScriptServiceV2Orchestrator`

source/Octopus.Tentacle.Client/TentacleClientOptions.cs
```diff
@@ -7,6 +7,11 @@ public class TentacleClientOptions
     {
         public RpcRetrySettings RpcRetrySettings { get; }
         
+        /// <summary>
+        /// Disables the use of ScriptServiceV3Alpha, even if it's supported on the Tentacle
+        /// </summary>
+        public bool DisableScriptServiceV3Alpha { get; set; }
```
sburmanoctopus, line 13: Nit: I think this would be better with a private set (or read-only) and added via the constructor

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionAdditionalScripts.cs
```diff
@@ -28,7 +28,7 @@ public async Task AdditionalScriptsWork(TentacleConfigurationTestCase tentacleCo
                 .CreateFile(path) // How files are made are different in bash and powershell, doing this ensures the client and tentacle really are using the correct script.
                 .Print("Hello");
 
-            var startScriptCommand = new StartScriptCommandV2Builder()
+            var startScriptCommand = new StartScriptCommandV3AlphaBuilder()
```
sburmanoctopus, line 31: 💭 Just an idea, but since we've had to replace 52 instances of `StartScriptCommandV2Builder` with `StartScriptCommandV3AlphaBuilder`, did we want to instead make a new builder called something like `StartScriptCommandLatestBuilder` or `LatestStartScriptCommandBuilder`, and use that instead everywhere?

That way, when we do V4, it won't be as invasive.

`LatestStartScriptCommandBuilder` could extend `StartScriptCommandV3AlphaBuilder`, so we keep `StartScriptCommandV3AlphaBuilder` for when it is needed.

Anyway, just a thought 🙂

source/Octopus.Tentacle.Tests.Integration/Support/TentacleConfigurationsAttribute.cs
```diff
@@ -59,7 +61,7 @@ static class TentacleConfigurationTestCases
                     TentacleVersions.Current,
                     TentacleVersions.v5_0_15_LastOfVersion5,
                     TentacleVersions.v6_3_417_LastWithScriptServiceV1Only,
-                    TentacleVersions.v7_0_1_ScriptServiceV2Added
+                    TentacleVersions.v8_0_73_LastWithoutScriptServiceV3Alpha
```
sburmanoctopus, line 64: @nathanwoctopusdeploy , do you think it's worth keeping `v7_0_1_ScriptServiceV2Added` as well as this new one?

source/Octopus.Tentacle.Tests.Integration/Support/TentacleVersions.cs
```diff
@@ -32,12 +36,17 @@ public static class TentacleVersions
             v5_0_15_LastOfVersion5,
             v6_3_417_LastWithScriptServiceV1Only,
             v6_3_451_NoCapabilitiesService,
-            v7_0_1_ScriptServiceV2Added
+            v8_0_73_LastWithoutScriptServiceV3Alpha
```
sburmanoctopus, line 39: Like before, should we keep 7_0_1 as well?

sburmanoctopus (approved):
A couple of small things, but generally LGTM.

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionAdditionalScripts.cs
APErebus: Yeah, that's not a bad idea. I'll make that change 👍

source/Octopus.Tentacle.Client/TentacleClientOptions.cs
APErebus: The reason I didn't make it part of the constructor is that it's not "required" for the options to be "valid". 

We do want it to be off by default on only disabled in certain circumstances (in tests + in Server while we work on the k8s agent)

source/Octopus.Tentacle.Tests.Integration/Support/TentacleConfigurationsAttribute.cs
sburmanoctopus: Ah, sorry. Re-reading my question it didn't capture what I meant...

What I was asking was do we think we should also add `v7_0_1_ScriptServiceV2Added` as well as `v8_0_73_LastWithoutScriptServiceV3Alpha` in this list? I.e. increase the number of test cases.

As it stands, we are replacing v7 with v8. Which could be fine, as v8 has both.

APErebus: ah yep, fair call. I'll let @nathanwoctopusdeploy answer that one :)

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionAdditionalScripts.cs
APErebus: Addressed in [cf37208](https://github.com/OctopusDeploy/OctopusTentacle/pull/669/commits/cf3720897e4b34f2f1399e93609abad6fff26098)

source/Octopus.Tentacle.Tests.Integration/Support/TentacleConfigurationsAttribute.cs
nathanwoctopusdeploy: Yeh lets keep v7 and v8 as there are enough significant changes between v6 -> v7 and the v7 -> v8 to warrant testing both.

I would opt to go for the latest version of v7 e.g. `7.0.189` as well in place of 7.0.1

APErebus: no probs!

APErebus: Readded testing of v7 in 160be6c6

APErebus: turns the version was `7.1.189`... Also, `8.0.73` won't be the latest before this PR is merged, so updated to `8.0.81` in 720742f7

PR #674, opened 2023-11-16
Update Kubernetes Tentacle to use ConfigMap for configuration storage

scme0:
# Background

Part of #project-k8s-agent

To make sure that the Kubernetes Tentacle is able to handle restarts (including when the nfs pod is restarted) we need to update Tentacle to be able to store its configuration into a ConfigMap.

# Results

I've add a new `IKeyValueStore` and `IWritableKeyValueStore` implementation which stores data in a Kubernetes ConfigMap. It requires a namespace to make sure that each Kubernetes Tentacle in the cluster has a unique ConfigMap.

I've also added an `IsRegistered` value to the store so that we can skip registration for Kubernetes Tentacles. We also needed to update the configuration process to remove the "remove trust" step which breaks the configuration process if we're already registered. I thought about updating that command to not clear if we're already registered but that seems to defeat the purpose of the command.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/docker/kubernetes-tentacle/scripts/configure-tentacle.sh
+++ b/docker/kubernetes-tentacle/scripts/configure-tentacle.sh
@@ -119,6 +119,25 @@ function configureTentacle() {
     tentacle new-certificate --instance "$instanceName" --if-blank
 }
 
+function setupVariablesForRegistrationCheck() {
+    local namespace=$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace)
+    local config_map_name="tentacle-config"
+    SERVICE_URL="https://$KUBERNETES_SERVICE_HOST:$KUBERNETES_SERVICE_PORT/api/v1/namespaces/$namespace/configmaps/$config_map_name";
+    SERVICE_ACCOUNT_TOKEN_PATH="/var/run/secrets/kubernetes.io/serviceaccount/token"
+}
+
+function getStatusOfRegistration() {
+    echo "Checking registration status..."
+
+    IS_REGISTERED=$(curl -s --request GET \
+    --url "$SERVICE_URL" \
+    --header "Authorization: Bearer $(cat $SERVICE_ACCOUNT_TOKEN_PATH)" \
+    --header 'Accept: application/json' \
+    --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt \
+    | grep -o '"Tentacle.Services.IsRegistered": "[^"]*"' \
+    | cut -d'"' -f4)
+}
+
 function registerTentacle() {
     echo "Registering with server ..."
 
@@ -203,6 +222,18 @@ function registerTentacle() {
 
     tentacle "${ARGS[@]}"
 }
+
+function verifyTentacleIsNotRegistered() {
+  setupVariablesForRegistrationCheck
+  getStatusOfRegistration
+
+  if [ "$IS_REGISTERED" == "true" ]; then
+      echo "Tentacle is already configured and registered with server."
+      exit 0
+  fi
+}
+
+verifyTentacleIsNotRegistered
 
 echo "==============================================="
 echo "Configuring Octopus Deploy Kubernetes Tentacle"
--- a/docker/kubernetes-tentacle/scripts/configure-tentacle.sh
+++ b/docker/kubernetes-tentacle/scripts/configure-tentacle.sh
@@ -115,9 +115,6 @@ function configureTentacle() {
         tentacle configure --instance "$instanceName" --port $internalListeningPort --noListen "False"
     fi
 
-    echo "Updating trust ..."
-    tentacle configure --instance "$instanceName" --reset-trust
-
     echo "Creating certificate ..."
     tentacle new-certificate --instance "$instanceName" --if-blank
 }
--- a/docker/kubernetes-tentacle/scripts/configure-tentacle.sh
+++ b/docker/kubernetes-tentacle/scripts/configure-tentacle.sh
@@ -7,7 +7,11 @@ if [[ "$ACCEPT_EULA" != "Y" ]]; then
 fi
 
 # Tentacle Docker images only support once instance per container. Running multiple instances can be achieved by running multiple containers.
-instanceName=Tentacle
+if [[ -n "$InstanceName" ]]; then
+    instanceName="$InstanceName"
+else
+    instanceName=Tentacle
+fi
 internalListeningPort=10933
 
 #If TentacleHome environment variable exists, use that
--- a/docker/kubernetes-tentacle/scripts/configure-tentacle.sh
+++ b/docker/kubernetes-tentacle/scripts/configure-tentacle.sh
@@ -7,11 +7,7 @@ if [[ "$ACCEPT_EULA" != "Y" ]]; then
 fi
 
 # Tentacle Docker images only support once instance per container. Running multiple instances can be achieved by running multiple containers.
-if [[ -n "$InstanceName" ]]; then
-    instanceName="$InstanceName"
-else
-    instanceName=Tentacle
-fi
+instanceName=Tentacle
 internalListeningPort=10933
 
 #If TentacleHome environment variable exists, use that
--- a/source/Octopus.Tentacle.Tests.Integration/TentacleCommandLineTests.cs
+++ b/source/Octopus.Tentacle.Tests.Integration/TentacleCommandLineTests.cs
@@ -596,9 +596,6 @@ Or one of the common options:
         {{
           ""Thumbprint"": ""{clientAndTentacle.Server.Thumbprint}"",
           ""CommunicationStyle"": 2,
-          ""KubernetesTentacleCommunicationMode"": {{
-            ""Value"": ""Polling""
-          }},
           ""Address"": ""https://localhost:{clientAndTentacle.Server.ServerListeningPort}"",
           ""Squid"": null,
           ""SubscriptionId"": ""{clientAndTentacle.RunningTentacle.ServiceUri}""
@@ -635,9 +632,6 @@ Or one of the common options:
         {{
           ""Thumbprint"": ""{clientAndTentacle.Server.Thumbprint}"",
           ""CommunicationStyle"": 1,
-          ""KubernetesTentacleCommunicationMode"": {{
-            ""Value"": ""Polling""
-          }},
           ""Address"": null,
           ""Squid"": null,
           ""SubscriptionId"": null
--- a/source/Octopus.Tentacle/Commands/ShowConfigurationCommand.cs
+++ b/source/Octopus.Tentacle/Commands/ShowConfigurationCommand.cs
@@ -20,8 +20,8 @@ namespace Octopus.Tentacle.Commands
 {
     public class ShowConfigurationCommand : AbstractStandardCommand
     {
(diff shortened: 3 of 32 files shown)


scme0:
[sc-60480]

docker/kubernetes-tentacle/scripts/configure-tentacle.sh
```diff
@@ -115,9 +115,6 @@ function configureTentacle() {
         tentacle configure --instance "$instanceName" --port $internalListeningPort --noListen "False"
     fi
 
-    echo "Updating trust ..."
-    tentacle configure --instance "$instanceName" --reset-trust
```
scme0, line 119: This clears out the `TrustedOctopusServers` which is set when registering. If we skip registering when we've already registered, we effectively clear out the trusted servers and then never replace them meaning a restart of the container fails.

I've remove this because we set the trusted server once and then never clear it. Is this going to cause us issues?

source/Octopus.Tentacle.Tests.Integration/TentacleCommandLineTests.cs
```diff
@@ -596,9 +596,6 @@ public async Task ShowConfigurationCommandLooksSensibleToHumans(TentacleConfigur
         {{
           ""Thumbprint"": ""{clientAndTentacle.Server.Thumbprint}"",
           ""CommunicationStyle"": 2,
-          ""KubernetesTentacleCommunicationMode"": {{
-            ""Value"": ""Polling""
-          }},
```
scme0, line 601: This property only shows up on serialisation if the CommunicationStyle is `KubernetesTentacle` now.

scme0: I did this because the example below is using `CommunicationStyle` of `Passive` but the `KubernetesTentacleCommunicationMode` says `Polling` which is confusing. I've just made it so that it's just not included at all to make it clearer.

source/Octopus.Tentacle/Commands/ShowConfigurationCommand.cs
```diff
@@ -86,26 +87,14 @@ async Task StartAsync()
 
         internal async Task CollectConfigurationSettings(DictionaryKeyValueStore outputStore)
```
scme0, line 88: This method was a pain in the ass - The `HomeConfiguration` and `TentacleConfiguration` classes would take a Configuration store (`IKeyValueStore`) but also an instance selector which is what you're meant to use to get the `IKeyValueStore`. But it was used to collect the properties to output them in JSON format. It's just very unclear where values are sourced from  and where values are set. I think I've made it clearer by using the `WriteTo` method so we know that the `IKeyValueStore` from the selector is used to source the data and it's written to the outputStore.

source/Octopus.Tentacle/Commands/ShowConfigurationCommand.cs
```diff
@@ -86,26 +87,14 @@ async Task StartAsync()
...
         {
-            var configStore = instanceSelector.Current.Configuration;
+            homeConfiguration.Value.WriteTo(outputStore);
 
-            var oldHomeConfiguration = new HomeConfiguration(ApplicationName.Tentacle, configStore!, instanceSelector);
-            var homeConfiguration = new WritableHomeConfiguration(ApplicationName.Tentacle, outputStore, instanceSelector);
-            homeConfiguration.SetHomeDirectory(oldHomeConfiguration.HomeDirectory);
+            tentacleConfiguration.Value.WriteTo(outputStore,
+                TentacleConfiguration.CertificateSettingName,
```
scme0, line 93: These are excluded values - it's not super clear in github though. Might change this.

scme0: Fixed

docker/kubernetes-tentacle/scripts/configure-tentacle.sh
APErebus: Rather than removing it, could we skip some of this configuration if we know it's configured? Or do we need to set some information before we start the tentacle app?

source/Octopus.Tentacle/Configuration/Instances/ConfigMapKeyValueStore.cs
```diff
@@ -0,0 +1,91 @@
...
+            {
+                config = client.CoreV1.ReadNamespacedConfigMap(Name, @namespace);
+            }
+            catch
+            {
+                log.Verbose($"ConfigMap for Tentacle Configuration not found for namespace {@namespace}, creating new ConfigMap.");
+                config = null;
+            }
+
+            configMap = config ?? client.CoreV1.CreateNamespacedConfigMap(new V1ConfigMap { Metadata = new V1ObjectMeta { Name = Name, NamespaceProperty = @namespace }}, @namespace);
```
APErebus, line 37: To keep with our existing structure, this should be in a `KubernetesConfigMapService` and injected into this value store.

source/Octopus.Tentacle/Kubernetes/KubernetesConfig.cs
```diff
@@ -1,9 +1,11 @@
 ﻿using System;
+using Octopus.Tentacle.Util;
 
 namespace Octopus.Tentacle.Kubernetes
 {
     public static class KubernetesConfig
     {
+        public static bool IsRunningInKubernetesCluster => !Environment.GetEnvironmentVariable("KUBERNETES_SERVICE_HOST").IsNullOrEmpty();
```
APErebus, line 8: We have something like this already in the `PlatformDetection`

source/Octopus.Tentacle/Configuration/Instances/ApplicationInstanceSelector.cs
```diff
@@ -64,30 +66,50 @@ ApplicationInstanceConfiguration LoadCurrentInstance()
...
+#endif
+            if (!ConfigurationFileExists(appInstance.configurationpath))
+            {
+                ThrowConfigurationMissingException(appInstance.instanceName, appInstance.configurationpath);
+            }
+
+            log.Verbose($"Loading configuration from {appInstance.configurationpath}");
+            var writable = new XmlFileKeyValueStore(fileSystem, appInstance.configurationpath!);
+            return (ContributeAdditionalConfiguration(writable), writable);
         }
```
APErebus, line 94: This i'm not so sure about, especially as we'll start to need DI injection 🤔 
Maybe we can use the factory pattern to get both DI & non-di here

APErebus (changes requested):
Broadly I like the approach, there is just a few things that need cleaning up/addressing

docker/kubernetes-tentacle/scripts/configure-tentacle.sh
scme0: If we just skip it when "isRegistered" is true, the only time it would be executed is when "isRegistered" is false, which is when there would be no trusted servers configured anyway. So I thought it didn't really make sense to have it when running in kubernetes...

source/Octopus.Tentacle/Kubernetes/KubernetesConfig.cs
scme0: Ah I didn't realise, thanks 👍

source/Octopus.Tentacle/Configuration/Instances/ConfigMapKeyValueStore.cs
scme0: Ah yep ok 👍

source/Octopus.Tentacle/Configuration/Instances/ApplicationInstanceSelector.cs
```diff
@@ -64,30 +69,48 @@ ApplicationInstanceConfiguration LoadCurrentInstance()
...
-                    ? $"The configuration file for instance {instanceName} could not be located at the specified location {configurationPath}. " +
-                    "The file might have been manually removed without properly removing the instance and as such it is still listed as present." +
-                    "The instance must be created again before you can interact with it."
-                    : $"The configuration file at {configurationPath} could not be located at the specified location.";
-                    
-                throw new ControlledFailureException(message);
+                log.Verbose($"Loading configuration from ConfigMap for namespace {@namespace}");
+                var configMapWritableStore = configMapStoreFactory.Value;
+                return (ContributeAdditionalConfiguration(configMapWritableStore), configMapWritableStore);
             }
```
APErebus, line 86: This check seems weird (checking the namespace). Should we be using the `PlatformDetection` here?

source/Octopus.Tentacle/Kubernetes/KubernetesConfigMapService.cs
```diff
@@ -0,0 +1,49 @@
...
+namespace Octopus.Tentacle.Kubernetes
+{
+    public interface IKubernetesV1ConfigMapService
+    {
+        Task<V1ConfigMap?> TryGet(string name, CancellationToken cancellationToken);
+
+        Task<V1ConfigMap> Patch(string name, IDictionary<string, string> data, CancellationToken cancellationToken);
+    }
+
+    public class KubernetesV1ConfigMapService : KubernetesService, IKubernetesV1ConfigMapService
```
APErebus, line 18: You can drop the `V1` here. I doubt we'll need to worry about `V2` K8s apis/types for a while

APErebus (approved):
Looks good! Couple of minor things that aren't blocking. The only other piece missing is the setting encryption when `ProectionLevel.MachineKey` using a value in a secret

source/Octopus.Tentacle/Configuration/Instances/ApplicationInstanceSelector.cs
scme0: Ah yep, good call!

source/Octopus.Tentacle/Kubernetes/KubernetesConfigMapService.cs
scme0: Removed 👍

source/Octopus.Tentacle/Configuration/Instances/ApplicationInstanceSelector.cs
scme0: Fixed 👍

source/Octopus.Tentacle/Configuration/Instances/ApplicationInstanceSelector.cs
```diff
@@ -64,30 +69,48 @@ ApplicationInstanceConfiguration LoadCurrentInstance()
...
+                ThrowConfigurationMissingException(appInstance.instanceName, appInstance.configurationpath);
+            }
+
+            log.Verbose($"Loading configuration from {appInstance.configurationpath}");
+            var writable = new XmlFileKeyValueStore(fileSystem, appInstance.configurationpath!);
+            return (ContributeAdditionalConfiguration(writable), writable);
         }
 
-        AggregatedKeyValueStore ContributeAdditionalConfiguration(XmlFileKeyValueStore writableConfig)
+        bool ConfigurationFileExists([NotNullWhen(true)] string? configurationPath)
```
LukeButters, line 97: nit: Did these methods need to be pulled out? It may be best to leave what we can alone to avoid any chance of introducing inadvertent changes.

ultra nit: Also since this just throws we could have pulled out the entire `if` into a single method.

PR #676, opened 2023-11-17
Ensure test coverage of ScriptServiceV2 now ScriptServiceV3Alpha has been added

nathanwoctopusdeploy:
# Background

If a test was using the `[TentacleConfigurations]` without explicitly including other versions of Tentacle, it would only run against the latest Tentacle. 

The latest tentacle has ScriptServiceV3Alpha, this resulted in a number of tests such as `ClientScriptExecutionRetriesTimeout` now only testing ScriptServiceV3Apha and not ScriptServiceV2 leaving significant code paths in TentacleClient untested.

# Results

This PR changes the default `[TentacleConfigurations]` to run against

- Latest Tentacle (build from source code)
- v7.1.189 which is the latest Tentacle with Sync Halibut and without ScriptServiceV3Alpha
- v8.0.81 which is the latest Tentacle with Async Halibut and without ScriptServiceV3Alpha

This will result in an increase in Tests run in the Halibut Build Chain

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Tests.Integration/ScriptServiceV3AlphaIntegrationTest.cs
+++ b/source/Octopus.Tentacle.Tests.Integration/ScriptServiceV3AlphaIntegrationTest.cs
@@ -18,7 +18,7 @@ namespace Octopus.Tentacle.Tests.Integration
     public class ScriptServiceV3AlphaIntegrationTest : IntegrationTest
     {
         [Test]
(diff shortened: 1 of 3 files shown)


APErebus (approved):
Thanks for picking that up and those names make a lot more sense for the version bumps 👍

PR #690, opened 2023-11-27
Add ability to execute scripts via Kubernetes Jobs

APErebus:
# Background

This is the big one 😱 

This adds support for executing scripts supplied to `ScriptServiceV3Alpha` by Kubernetes Jobs.

# Results

Adds a new `KubernetesJobScriptExecutor` that is used when the environment variable `OCTOPUS__TENTACLE__K8SUSEJOBS` is set to `True`

At a high level, the new `RunningKubernetesJobScript` creates a new k8s Job and executes the `Bootstrap` script sent from server in the `octopuslabs/k8-workertools` container image. It then starts two background processes, one to monitor the status of the job itself (so it knows when it's finished) and another to read the streamed log files (`stdout.log` and `stderr.log`) and write them into the `Output.log`. For more info on this process, see below

The tag of the worker tools image used is linked to the version of k8s the tentacle is running in, which we query via the API.

It's implied that there is also a volume called `tentacle-home` that has access to the tentacle's work directory so it can execute the script. This is supplied as raw yaml in the `OCTOPUS__TENTACLE__K8SJOBVOLUMEYAML` environment variable.

### Script Logs

As Tentacle needs to write the `Output.log` in a thread-safe manner (so logs don't get munged), we execute the bootstrap script using a `bootstrapRunner.sh` script which redirects the `stdout` and `stderr` streams from the `Bootstrap` script into two files in the work directory, `stdout.log` and `stderr.log`. Then there is a process in the `KubernetesJobOutputStreamWriter` class that periodically (every 250ms) tail observes those files for new log entries. When there are, they are then written to the tentacle's `Output.log` under the same locking mechanism as the rest of the code.

Once the job is finished, we perform one last read of those files just in case there was final log items flushed.

### Testing

Currently there is no automated tests for this code. I'm still working on refining it so manual testing will suffice for now until the codebase stabilises.

I was able to successfully run a health check for the the `Kubernetes Tentacle` in my local server.

Shortcut story: [sc-59580]

## After

image

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Octopus.Tentacle/Scripts/IScriptExecutor.cs
```diff
@@ -5,6 +5,7 @@ namespace Octopus.Tentacle.Scripts
 {
     public interface IScriptExecutor
     {
-        IRunningScript ExecuteOnBackgroundThread(StartScriptCommandV3Alpha command, IScriptWorkspace workspace, ScriptStateStore? scriptStateStore, CancellationToken cancellationToken);
+        bool ValidateExecutionContext(IScriptExecutionContext executionContext);
```
zentron, line 8: Naming suggestion, we use a semantic like `CanExecute` in other places. It makes it a little clearer what the check is doing than just validate, which sounds like it could be verifying the contents itself rather than just checking if its valid to be passed in at all.

source/Octopus.Tentacle/Scripts/Kubernetes/RunningKubernetesJobScript.cs
```diff
@@ -0,0 +1,323 @@
...
+        });
+
+        readonly IScriptWorkspace workspace;
+        readonly ScriptTicket scriptTicket;
+        readonly string taskId;
+        readonly ILog log;
+        readonly IScriptStateStore stateStore;
+        readonly IKubernetesJobService jobService;
+        readonly IKubernetesClusterService kubernetesClusterService;
+        readonly KubernetesJobScriptExecutionContext executionContext;
```
zentron, line 37: Doesn't look like this is used anywhere

source/Octopus.Tentacle/Scripts/Kubernetes/KubernetesJobOutputStreamWriter.cs
```diff
@@ -0,0 +1,141 @@
...
+            {
+                //ignore all task cancelled exceptions as they may be thrown by the job finishing (and thus signally)
+            }
+        }
+
+        async Task<StreamReader?> SafelyOpenLogStreamReader(string filename, CancellationToken cancellationToken)
+        {
+            while (true)
+            {
+                if (cancellationToken.IsCancellationRequested)
```
zentron, line 85: I was wondering about the maximum timeout here till I realised we were going to rely on the cancellationToken timeout which makes total sense. Maybe just throw a comment above the `while(true)` about how it can exit since that can raise eyebrows.

source/Octopus.Tentacle/Scripts/Kubernetes/KubernetesJobOutputStreamWriter.cs
```diff
@@ -0,0 +1,141 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Octopus.Tentacle.Contracts;
+using Octopus.Tentacle.Util;
+
+namespace Octopus.Tentacle.Scripts.Kubernetes
+{
+    public class KubernetesJobOutputStreamWriter
```
zentron, line 13: To confirm what this is trying to bring together:

1. The jobs themselves now have a bash wrapper that pipes the logs into one of two files based on stderr and stdout. To prevent interleaved log entries being created at this stage we keep the files separate on the initial write.
2. The  `RunningKubernetesJobScript` is constantly running,  and waits until effectively the task has been completed.
3. The `KubernetesJobOutputStreamWriter` meanwhile is sitting in a loop, reading the 2 generated files from step 1, and combining them into a single `IScriptLogWriter` log.
4. The various ScriptService implimentations meanwhile, are periodically called with `GetResponse` (in addition to from other calls) which read from single `IScriptLogWriter`, offsetting at each stage by whatever the `lastLogSequence `read was. Since we want to return a single stream over to server, this requirement is effectively what creates the need for the intermediate step, rather than leaving it as two separate `std*` files and avoiding having to combine them.

It would be nice if we didnt need the double read/write but I think I can see that we dont have too many other options based on how it works.

source/Octopus.Tentacle/Scripts/Kubernetes/RunningKubernetesJobScript.cs
```diff
@@ -0,0 +1,323 @@
...
+            var exitCode = -1;
+
+            var linkedCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(scriptCancellationToken, taskCancellationToken);
+            var cancellationToken = linkedCancellationTokenSource.Token;
+            try
+            {
+                using var writer = ScriptLog.CreateWriter();
+                try
+                {
+                    using (ScriptIsolationMutex.Acquire(workspace.IsolationLevel,
```
zentron, line 85: This presumably works because when running one instance, the mutex is accessable by that one Tentacle process.
If/when we allow multiple instances to run within a cluster we may need to answer if this isolationlevel check is still relevant.

source/Octopus.Tentacle/Scripts/Kubernetes/RunningKubernetesJobScript.cs
```diff
@@ -0,0 +1,323 @@
...
+                                        new(EnvironmentVariables.TentacleHome, $"/data/tentacle-home/{instanceName}"),
+                                        new(EnvironmentVariables.TentacleVersion, Environment.GetEnvironmentVariable(EnvironmentVariables.TentacleVersion)),
+                                        new(EnvironmentVariables.TentacleCertificateSignatureAlgorithm, Environment.GetEnvironmentVariable(EnvironmentVariables.TentacleCertificateSignatureAlgorithm)),
+                                        new("OCTOPUS_RUNNING_IN_CONTAINER", "Y")
+                                    }
+                                }
+                            },
+                            ServiceAccountName = KubernetesConfig.JobServiceAccountName,
+                            RestartPolicy = "Never",
+                            Volumes = volumes
```
zentron, line 245: Nice to see that worked :)

source/Octopus.Tentacle/Scripts/Kubernetes/RunningKubernetesJobScript.cs
```diff
@@ -0,0 +1,323 @@
...
+                                        {
+                                            $"'{arg}'"
+                                        })).ToList(),
+                                    VolumeMounts = new List<V1VolumeMount>
+                                    {
+                                        new("/data/tentacle-home", "tentacle-home"),
+                                    },
+                                    Env = new List<V1EnvVar>
+                                    {
+                                        new(EnvironmentVariables.TentacleHome, $"/data/tentacle-home/{instanceName}"),
```
zentron, line 236: Some of those other environment variables sound like they could be relevant for Calamari. `TentacleJournal` or `TentacleInstanceName`?

source/Octopus.Tentacle/Scripts/Kubernetes/RunningKubernetesJobScript.cs
```diff
@@ -0,0 +1,323 @@
...
+        }
+
+        static readonly List<Version> KnownLatestContainerTags = new List<Version>
+        {
+            new(1, 26, 3),
+            new(1, 27, 3),
+            new(1, 28, 2),
+        };
+
+        async Task<string> GetDefaultContainer()
```
zentron, line 265: Minor, consider pulling some parts of this class such as this worker tool container resolution into its own class that can be tested seperately. ALso makes this class easier to grok.

source/Octopus.Tentacle/Scripts/ScriptExecutorFactory.cs
```diff
@@ -1,19 +1,27 @@
...
+        public ScriptExecutorFactory(Lazy<LocalShellScriptExecutor> shellScriptExecutor, Lazy<KubernetesJobScriptExecutor> kubernetesJobScriptExecutor)
         {
             this.shellScriptExecutor = shellScriptExecutor;
+            this.kubernetesJobScriptExecutor = kubernetesJobScriptExecutor;
         }
 
         public IScriptExecutor GetExecutor()
         {
-            return shellScriptExecutor.Value;
+            return KubernetesConfig.UseJobs switch
```
zentron, line 20: Would it be simpler to push as much of the "if k8s do this" into being container registration concern?

zentron (changes requested):
Nice work, It took a bunch of time work to get to this point but it looks fairly clean. And nice that it was done in conjunction with the H&E group.

Just a couple questions here and there, I dont  think anything particularly structural.

source/Octopus.Tentacle/ExternalInit.cs
```diff
@@ -0,0 +1,4 @@
+namespace System.Runtime.CompilerServices
```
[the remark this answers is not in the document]
APErebus: Oh yeah, I meant to make a note of this in the PR. It's needed for .NET 4.8 and `record` types. I'll add a comment in code

source/Octopus.Tentacle/Kubernetes/KubernetesJobService.cs
```diff
@@ -0,0 +1,89 @@
...
+                return await Client.ReadNamespacedJobStatusAsync(jobName, KubernetesConfig.Namespace, cancellationToken: cancellationToken);
+            }
+            catch (HttpOperationException opException)
+            {
+                if (opException.Response.StatusCode == HttpStatusCode.NotFound)
+                    return null;
+
+                //if there is some other error, just throw the exception
+                throw;
+            }
```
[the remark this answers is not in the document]
APErebus: Good call, I always forget about exception filters

source/Octopus.Tentacle/Kubernetes/KubernetesJobService.cs
```diff
@@ -0,0 +1,89 @@
...
+        public async Task Watch(ScriptTicket scriptTicket, Func<V1Job, bool> onChange, Action<Exception> onError, CancellationToken cancellationToken)
+        {
+            var jobName = BuildJobName(scriptTicket);
+
+            using var response = Client.BatchV1.ListNamespacedJobWithHttpMessagesAsync(
+                KubernetesConfig.Namespace,
+                //only list this job
+                fieldSelector: $"metadata.name=={jobName}",
+                watch: true,
+                timeoutSeconds: 1800, //same as the TTL of the job itself
```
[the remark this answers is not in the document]
APErebus: yep, good call.

source/Octopus.Tentacle/Kubernetes/bootstrapRunner.sh
```diff
@@ -0,0 +1,46 @@
...
+#This is the args for the Bootstrap script
+shift
+
+exec > >(logStdOut)
+exec 2> >(logStdErr >&2)
+
+/bin/bash $BOOTSTRAP_SCRIPT "$@"
+
+# This ungodly hack is to stop the pod from being killed before the last log has been flushed
+sleep 0.250 #250ms
```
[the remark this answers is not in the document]
APErebus: Yeah, that's a great idea

source/Octopus.Tentacle/Octopus.Tentacle.csproj
```diff
@@ -40,6 +40,12 @@
 	<PropertyGroup Condition=" '$(TargetFramework)' == 'net6.0' ">
 		<DefineConstants>$(DefineConstants);HTTP_CLIENT_SUPPORTS_SSL_OPTIONS;REQUIRES_EXPLICIT_LOG_CONFIG;REQUIRES_CODE_PAGE_PROVIDER;USER_INTERACTIVE_DOES_NOT_WORK;DEFAULT_PROXY_IS_NOT_AVAILABLE;HAS_NULLABLE_REF_TYPES</DefineConstants>
 	</PropertyGroup>
+	<ItemGroup Condition=" '$(TargetFramework)' == 'net48' ">
+		<PackageReference Include="KubernetesClient.Classic" Version="10.1.19" />
```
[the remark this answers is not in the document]
APErebus: I think it's more for completeness and then I don't need to put pre-processor symbols everywhere to not compile the code for .NET 4.8.

source/Octopus.Tentacle/Scripts/Kubernetes/RunningKubernetesJobScript.cs
APErebus: ah yep, isn't anymore.

source/Octopus.Tentacle/Scripts/Kubernetes/KubernetesJobOutputStreamWriter.cs
APErebus: this is correct. I'd love to be able to write directly to the `Output.log` but I was concerned about multiple writes on different threads causing issues (it was raised as a concern by @LukeButters)

source/Octopus.Tentacle/Scripts/Kubernetes/RunningKubernetesJobScript.cs
APErebus: Correct. When/if we get to HA, then we'll need a different approach to how to do the script isolation mutex or, as you say, if it's still relevant

source/Octopus.Tentacle/Scripts/Kubernetes/RunningKubernetesJobScript.cs
APErebus: hrm, good call. I'll investigate. I added these ones because they were needed for the health-check

source/Octopus.Tentacle/Scripts/ScriptExecutorFactory.cs
APErebus: yeah, fair call. I'll move this into the module and only register the appropriate executor based on the env var

source/Octopus.Tentacle/Scripts/Kubernetes/RunningKubernetesJobScript.cs
APErebus: Yep, I missed `TentacleJournal` and `TentacleInstanceName`. Added

APErebus:
@zentron I've updated the PR with the changes suggested. I did end up removing the `IScriptExecutorFactory`  and moving it into the module.

I also pulled the container stuff into a new class, `IKubernetesJobContainerResolver`

source/Octopus.Tentacle/Scripts/Kubernetes/RunningKubernetesJobScript.cs
APErebus: Done. Created a new `KubernetesJobContainerResolver` in [a2a48b7](https://github.com/OctopusDeploy/OctopusTentacle/pull/690/commits/a2a48b76b42da74a82b8d99115f1ce3d09c49010)

source/Octopus.Tentacle/Scripts/ScriptExecutorFactory.cs
APErebus: Done in [a2a48b7](https://github.com/OctopusDeploy/OctopusTentacle/pull/690/commits/a2a48b76b42da74a82b8d99115f1ce3d09c49010)

There is no longer a factory and the script executors are now conditionaly registered in the Autofac container

source/Octopus.Tentacle/Kubernetes/bootstrapRunner.sh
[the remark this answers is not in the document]
APErebus: Done in [a2a48b7](https://github.com/OctopusDeploy/OctopusTentacle/pull/690/commits/a2a48b76b42da74a82b8d99115f1ce3d09c49010)

source/Octopus.Tentacle/Scripts/IScriptExecutor.cs
APErebus: Changed in [a2a48b7](https://github.com/OctopusDeploy/OctopusTentacle/pull/690/commits/a2a48b76b42da74a82b8d99115f1ce3d09c49010)

Now takes the entire command, rather than just the execution context

source/Octopus.Tentacle/Kubernetes/KubernetesConfig.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+
+namespace Octopus.Tentacle.Kubernetes
+{
+    public static class KubernetesConfig
+    {
+        public static string Namespace => Environment.GetEnvironmentVariable("OCTOPUS__K8STENTACLE__NAMESPACE")
```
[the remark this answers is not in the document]
APErebus: I chose to pull out a common method in [a2a48b7](https://github.com/OctopusDeploy/OctopusTentacle/pull/690/commits/a2a48b76b42da74a82b8d99115f1ce3d09c49010)

zentron (approved):
Nice one, this is a huge piece of the puzzle 👍

PR #703, opened 2023-11-29
Attempt to download tentacles multiple times

LukeButters:
# Background

We are seeing time outs when downloading older tentacle versions, where the timeout is set to 150s.

This PR attempts to help with that by:
* Reducing the timoeut to 20s, since the timeout appears to apply to each read/write rather than the operation as a whole.
* Attempt 5 times (with a simple for loop since the problem we have is a timout of 20s to read any bytes so we probably don't have a reason to get fancy).

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Tests.Integration/Support/OctopusPackageDownloader.cs
+++ b/source/Octopus.Tentacle.Tests.Integration/Support/OctopusPackageDownloader.cs
@@ -86,7 +86,7 @@ namespace Octopus.Tentacle.Tests.Integration
                             
                             totalTime.Stop();
 
-                            logger.Information("Download Finished in {totalTime}", totalTime.ElapsedMilliseconds);
+                            logger.Information("Download Finished in {totalTime}ms", totalTime.ElapsedMilliseconds);
                         }
                     }
                 }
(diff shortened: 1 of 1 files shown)


source/Octopus.Tentacle.Tests.Integration/Support/OctopusPackageDownloader.cs
```diff
@@ -17,56 +18,99 @@ public class OctopusPackageDownloader
...
+                            logger.Information("Download Finished in {totalTime}", totalTime.ElapsedMilliseconds);
                         }
-
-                        logger.Information("Download Finished");
                     }
                 }
             }
+            catch (Exception e)
+            {
+                throw new Exception($"Failure to download: {downloadUrl}. After {totalTime.Elapsed.TotalSeconds} seconds we only downloaded, {totalRead}", e);
```
LukeButters, line 96: These logs don't make it to the build server :(, so instead make the exception contain information already in the logs.

source/Octopus.Tentacle.Tests.Integration/Support/OctopusPackageDownloader.cs
```diff
@@ -17,56 +18,99 @@ public class OctopusPackageDownloader
     {
         public static async Task DownloadPackage(string downloadUrl, string filePath, ILogger logger)
         {
+            var exceptions = new List<Exception>();
+            for (int i = 0; i < 5; i++)
```
LukeButters, line 22: Simple for loop is simple

nathanwoctopusdeploy (approved):

PR #705, opened 2023-11-29
Support overriding configuration directories in K8s Container

APErebus:
# Background

The Kubernetes Tentacle needs to be able to configure the directory path for the home + application directories so they can be specified to be on a NFS/volume mount 

# Results

The container now takes the `TentacleHome` and `TentacleApplications` environment variables and uses them in `configure-tentacle.sh` if they are set to override the default values.

We also now explicitly set the `home` directory in the `create-instance` command (for clarity).

Also some minor bash cleanups: `! -z` is the same as `-n`

Shortcut story: [sc-66137]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/build-k8s-docker-local.ps1
+++ b/build-k8s-docker-local.ps1
@@ -1,0 +1,24 @@
+param (
+    [Parameter(Mandatory=$True)]
+    [string]
+    $BuildNumber,
+    
+    [Parameter()]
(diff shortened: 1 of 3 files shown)


MissedTheMark (approved):
LGTM 👍

PR #707, opened 2023-11-30
Generate machine key in home directory in K8s

APErebus:
# Background

The machine key is currently generated for linux in `/etc/octopus/machinekey`. This then is _regenerated_ when the pod restarts, so the encrypted values (which are stored in the home directory, on a NFS share) can't be decrypted

# Results

When running in K8S only, we change the location of the file to generate the machine key to be in the configured environment variable `TentacleHome` (which currently is the NFS share `/octopus`). This results in the key being persisted across pod restarts and the cert can be decrypted

Shortcut story: [sc-66137]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle/Util/PlatformDetection.cs
+++ b/source/Octopus.Tentacle/Util/PlatformDetection.cs
@@ -14,8 +14,7 @@ namespace Octopus.Tentacle.Util
             /// <summary>
             /// Indicates if the Tentacle is running inside a Kubernetes cluster.
             /// </summary>
-            public static bool IsRunningInKubernetes => !string.IsNullOrWhiteSpace(Environment.GetEnvironmentVariable("KUBERNETES_SERVICE_HOST")) ||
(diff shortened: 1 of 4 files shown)


source/Octopus.Tentacle/Util/PlatformDetection.cs
```diff
@@ -14,8 +14,7 @@ public static class Kubernetes
             /// <summary>
             /// Indicates if the Tentacle is running inside a Kubernetes cluster.
             /// </summary>
-            public static bool IsRunningInKubernetes => !string.IsNullOrWhiteSpace(Environment.GetEnvironmentVariable("KUBERNETES_SERVICE_HOST")) ||
-                (bool.TryParse(Environment.GetEnvironmentVariable("OCTOPUS__TENTACLE__FORCEK8S"), out var b) && b);
+            public static bool IsRunningInKubernetes => bool.TryParse(Environment.GetEnvironmentVariable("OCTOPUS__K8STENTACLE__FORCE"), out var b) && b;
```
APErebus, line 17: Changed this to be explicit and this env var _will_ be set in the helm/pod configuration

zentron (approved):

PR #716, opened 2023-12-02
Bump Halibut so HalibutRuntime is Disposed correctly on Tentacle Shutdown

nathanwoctopusdeploy:
# Background

Fix a bug where Tentacle does not dispose of the HalibutRuntime on shutdown.

This can result in the listening port for a listening tentacle not being freed up quickly and stop the tentacle service from being stopped and started quickly.

The HalibutRuntime only had a DisposeAsync method, Dispose has also been added so it works correctly with the version of Autofac used in Tentacle.

Fixes https://github.com/OctopusDeploy/OctopusTentacle/issues/725 

[sc-66309]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Client/TentacleClient.cs
+++ b/source/Octopus.Tentacle.Client/TentacleClient.cs
@@ -37,7 +37,7 @@ namespace Octopus.Tentacle.Client
             var innerHandler = halibutRuntime.OverrideErrorResponseMessageCaching;
             halibutRuntime.OverrideErrorResponseMessageCaching = response =>
             {
-                if (BackwardsCompatibleCapabilitiesV2Helper.ExceptionTypeLooksLikeTheServiceWasNotFound(response.Error.HalibutErrorType) ||
+                if (BackwardsCompatibleCapabilitiesV2Helper.ExceptionTypeLooksLikeTheServiceWasNotFound(response.Error!.HalibutErrorType!) ||
                     BackwardsCompatibleCapabilitiesV2Helper.ExceptionMessageLooksLikeTheServiceWasNotFound(response.Error.Message))
                 {
                     return true;
--- a/source/Octopus.Tentacle.Contracts/Octopus.Tentacle.Contracts.csproj
+++ b/source/Octopus.Tentacle.Contracts/Octopus.Tentacle.Contracts.csproj
@@ -28,7 +28,7 @@
     </Otherwise>
   </Choose>
   <ItemGroup>
-    <PackageReference Include="Halibut" Version="7.0.285" />
+    <PackageReference Include="Halibut" Version="7.0.359" />
     <PackageReference Include="Octopus.Diagnostics" Version="2.1.0" />
   </ItemGroup>
   <ItemGroup>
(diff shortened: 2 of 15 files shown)


source/Octopus.Tentacle/Communications/HalibutInitializer.cs
```diff
@@ -133,6 +134,17 @@ IPEndPoint GetEndPointToListenOn()
 
         public void Stop()
         {
+            halibut.DisposeAsync().GetAwaiter().GetResult();
+        }
+
+        public async ValueTask DisposeAsync()
```
nathanwoctopusdeploy, line 140: DisposeAsync is not called when the container is disposed of, but this has been added for completeness.

source/Octopus.Tentacle/Communications/HalibutInitializer.cs
```diff
@@ -134,5 +135,19 @@ IPEndPoint GetEndPointToListenOn()
...
         }
+        
+        public void Dispose()
+        {
+            halibut.DisposeAsync().GetAwaiter().GetResult();
+
+            log.Dispose();
+        }
+
+        public async ValueTask DisposeAsync()
```
nathanwoctopusdeploy, line 146: DisposeAsync is not called but has been added for completeness

source/Octopus.Tentacle/Communications/HalibutInitializer.cs
```diff
@@ -134,5 +135,19 @@ IPEndPoint GetEndPointToListenOn()
         public void Stop()
         {
         }
+        
+        public void Dispose()
```
sburmanoctopus, line 139: 🤔 This struck me as odd that we have to do this, as Autofac naturally disposes of dependencies given to a class.

I tested this, and this is indeed true. So why did we need to do this for Halibut?

Turns out it's the difference between `IDisposable` and `IAsyncDisposable`. The former will get disposed by Autofac, and the latter will not 😑.

Since we recently removed the `IDisposable` interface from Halibut, this is why it has stopped being disposed of by Autofac.

Knowing all this, it made me wonder:
- What we did here will work. But would it be better or worse to put back the `HalibutRuntime.Dispose` method as an alternative to solving this problem?
- Also, we end up double disposing `log`, as Autofac also disposes it (I tested it to make sure).

source/Octopus.Tentacle/Tentacle.exe.nlog
```diff
@@ -34,7 +34,7 @@
     <logger name="LogFileOnlyLogger" writeTo="octopus-log-file" final="true" />
     <logger name="*" minlevel="Info" maxLevel="Warn" writeTo="stdout" />
     <logger name="*" minlevel="Error" writeTo="stderr" />
-    <logger name="*" minlevel="Info" writeTo="octopus-log-file" />
+    <logger name="*" minlevel="Trace" writeTo="octopus-log-file" />
```
sburmanoctopus, line 37: Do we want to keep this at Trace level?

sburmanoctopus (approved):
Nice test for checking works correctly.

I'll leave it to you to decide what to do about:
- The double dispose of log
- Whether we want to solve this with reintroducing `Dispose` to `HalibutRuntime` instead.

source/Octopus.Tentacle/Tentacle.exe.nlog
nathanwoctopusdeploy: Ahh, no. I'm having issues with the service not starting correctly sometimes hence this hack!!

PR #730, opened 2023-12-05
Co-operatively cancel Tentacle Client Operations

nathanwoctopusdeploy:
This is the reverted PR https://github.com/OctopusDeploy/OctopusTentacle/pull/647 rebased

# Background

[sc-58429]

Related to: https://github.com/OctopusDeploy/Halibut/pull/525

Before Async Halibut was introduced, an RPC the was transferring (reading or writing over a TCP Connection) could not be cancelled in Halibut, only an RPC that was connecting could be cancelled.

Async Halibut, along with https://github.com/OctopusDeploy/Halibut/pull/525 has added the ability to cancel RPC whether connecting or transferring. This allows the caller of Halibut to cancel an RPC with a cancellation token co-operatively.

Before Async Halibut, abandoning a Transferring RPC was the only way to "cancel" it. That is, run it as a different Task and walk away from the running Task if a cancellation was performed. While this method would allow the caller to regain control flow, it left the Task still executing and left IO operations on TCP Connection from Tentacle Client to Tentacle active.

# Results

## Before

Tentacle Client only supported the cancellation of connecting RPC. If an RPC was transferred, cancellation could not be performed, and the caller would have to wait for the RPC to complete.

Tentacle Client used the abandon technique for RPC Retry Timeouts, as it was the only reliable way to timeout the Retries and return control flow to the caller. 

## After

Tentacle Client will co-operatively cancel RPC to the Socket, stopping active reads and writes over TCP Connections.

Tentacle Client will co-operatively cancel RPC to the Socket when RPC Retries Timeout.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Octopus.Tentacle.Contracts/Octopus.Tentacle.Contracts.csproj
```diff
@@ -28,7 +28,7 @@
     </Otherwise>
   </Choose>
   <ItemGroup>
-    <PackageReference Include="Halibut" Version="7.0.359" />
+    <PackageReference Include="Halibut" Version="7.0.374-pull-576" />
```
nathanwoctopusdeploy, line 31: TODO - Bump when Halibut PR is merged

source/Octopus.Tentacle.Client/Scripts/HalibutClientExceptionExtensionMethods.cs
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        public static bool IsHalibutWrappedConnectingRequestCancelledException(this HalibutClientException? ex)
+        {
+            if (ex is null)
+            {
+                return false;
+            }
+
+            // TODO: Make this type safe
```
nathanwoctopusdeploy, line 23: We have tests that will catch if this stops working but it needs to be improved. This will be done in a future PR

source/Octopus.Tentacle.Client/Scripts/HalibutClientExceptionExtensionMethods.cs
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        public static bool IsHalibutWrappedTransferringRequestCancelledException(this HalibutClientException? ex)
+        {
+            if (ex is null)
+            {
+                return false;
+            }
+
+            // TODO: Make this type safe
```
nathanwoctopusdeploy, line 44: We have tesats that will catch if this stops working but it needs to be improved. This will be done in a future PR

source/Octopus.Tentacle.Contracts/Octopus.Tentacle.Contracts.csproj
```diff
@@ -28,7 +28,7 @@
     </Otherwise>
   </Choose>
   <ItemGroup>
-    <PackageReference Include="Halibut" Version="7.0.359" />
+    <PackageReference Include="Halibut" Version="7.0.386-pull-576" />
```
nathanwoctopusdeploy, line 31: TODO bump once https://github.com/OctopusDeploy/Halibut/pull/576 is merged

sburmanoctopus (approved):
LGTM. Some small changes to the original PR, but they look legit 😁

PR #736, opened 2023-12-11
Make list instances command less fragile to instances being deleted while the command is running

nathanwoctopusdeploy:
# Background

Try and fix https://build.octopushq.com/buildConfiguration/TeamFireAndMotion_OctopusTentacle_IntegrationTestNet60onWindows2016/10515949?expandBuildDeploymentsSection=false&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildProblemsSection=true&expandBuildTestsSection=true

image

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle/Configuration/Instances/ApplicationInstanceStore.cs
+++ b/source/Octopus.Tentacle/Configuration/Instances/ApplicationInstanceStore.cs
@@ -92,16 +92,38 @@ namespace Octopus.Tentacle.Configuration.Instances
             var listFromRegistry = registryApplicationInstanceStore.GetListFromRegistry();
             var listFromFileSystem = new List<ApplicationInstanceRecord>();
             if (fileSystem.DirectoryExists(InstancesFolder))
-                listFromFileSystem = fileSystem.EnumerateFiles(InstancesFolder)
-                    .Select(LoadInstanceConfiguration)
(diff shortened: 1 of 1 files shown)


source/Octopus.Tentacle/Configuration/Instances/ApplicationInstanceStore.cs
```diff
@@ -92,16 +92,38 @@ public IList<ApplicationInstanceRecord> ListInstances()
             var listFromRegistry = registryApplicationInstanceStore.GetListFromRegistry();
             var listFromFileSystem = new List<ApplicationInstanceRecord>();
             if (fileSystem.DirectoryExists(InstancesFolder))
```
LukeButters, line 94: i think we ought to have a env var which lets us control the instance folder so each test can be isolated.

LukeButters (approved):

source/Octopus.Tentacle/Configuration/Instances/ApplicationInstanceStore.cs
LukeButters: I would prefer we do this over swallowing exeptions, but approved if u want to proceed as is

nathanwoctopusdeploy: Mmm, that would be better 👍

PR #737, opened 2023-12-12
Fix flakey tentacle command line tests

sburmanoctopus:
[sc-65398]

# Background

The `TentacleCommandLineTests.ShouldLogStartupDiagnosticsToInstanceLogFileOnly` test was [flaky](https://build.octopushq.com/buildConfiguration/TeamFireAndMotion_OctopusTentacleNet48_SensibleDefaultsChainFullChain/10493993?buildTab=tests&status=muted&expandedTest=build%3A%28id%3A10493988%29%2Cid%3A2000000029). This needed to be fixed.

# Results

## Before

The test would create an instance of Tentacle by running `Tentacle.exe` (setting everything up). Then, while this instance was running, it would run the `show-thumbprint` command against the same executable, creating a second instance.

What we believe was causing the failure was that the first instance was writing to the log file while the second instance was starting up. This would, lock the log file and prevent the second instance from writing logs.

This test does its assertion by looking at the log file. Since the second instance logs were not being logged, it would therefore fail.

One would expect the second instance to throw an error when file locking occurs. However, we have configured NLog with `throwExceptions="false"`, so if any errors are thrown while logging, it will simply **not** log and continue as if nothing has happened. This is great for customers, but not for finding test issues.

## After

To solve this, we stop the first instance of the running tentacle before we attempt to run the command. This will prevent the file lock, and will therefore allows the logs to be populated as expected.

To make the other tests in this file safer, we also stopped the initial Tentacle instance for each test that also ran a second instance over the first.

To help diagnose future issues, we have also set `throwExceptions="true"` for tests only. We proved this does indeed work with [this test run](https://build.octopushq.com/buildConfiguration/TeamFireAndMotion_OctopusTentacleNet48_SensibleDefaultsChainFullChain/10540109?buildTab=tests&name=ShouldLogStartupDiagnosticsToInstanceLogFileOnly&expandedTest=build%3A%28id%3A10540085%29%2Cid%3A2000000029) (where we purposely locked the file)

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Octopus.Tentacle.Tests.Integration/TentacleCommandLineTests.cs
```diff
@@ -404,10 +408,12 @@ public async Task ListInstancesCommandJson(TentacleConfigurationTestCase tc)
         public async Task ShouldLogStartupDiagnosticsToInstanceLogFileOnly(TentacleConfigurationTestCase tc)
```
sburmanoctopus, line 408: This was the test that was flaky

LukeButters (approved):

sburmanoctopus:
FYI, turns out this was not the primary reason for the flaky test.

The main reason was that when Tentacle starts up, it checks to see if it `CanLoadCurrentInstance` before configuring logging.

If that fails, it will fall back to logging into the default log file (that all tentacles log into). This is what happened when the logs were not showing up.

`CanLoadCurrentInstance` was claiming to not be able to load an instance because it loads all instances before it tries to find the instance we are looking for. So when another test is manipulating that directory (e.g. deleting instances), we may encounter an exception from that one unrelated instance.

[This PR](https://github.com/OctopusDeploy/OctopusTentacle/pull/738) will solve that issue for tests.

nathanwoctopusdeploy (approved):
LGTM

sburmanoctopus:
We decided to remove `throwExceptions="true"` adjustment for testing, as it was causing a test to fail because NLog was trying to create the "Octopus Tentacle" event source, and it already existed. This is not an issue normally, as `throwExceptions="false"` normally.

It was considered not worth investing time to fix this. So we deleted it.

PR #738, opened 2023-12-13
Allow the MachineConfigurationHomeDirectory to be changed and use a unique directory per test to isolate cross test issues

nathanwoctopusdeploy:
# Background

Allow the MachineConfigurationHomeDirectory to be changed and use a unique directory per test to isolate cross test issues

[sc-66742]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

- [ ] I have read [How we use GitHub Issues](https://github.com/OctopusDeploy/Issues/blob/master/docs/CONTRIBUTING.internal.md) for help deciding when and where it's appropriate to make an issue.
- [ ] I have considered informing or consulting the right people, according to the [ownership map](https://whimsical.com/ownership-map-NzbiD4HJyvhC9jNJNfS6TG).
- [ ] I have considered appropriate testing for my change.

PR #741, opened 2023-12-14
Fix Tests that start Tentacle as a Service

nathanwoctopusdeploy:
# Background

Set Environment variables for the Windows and Linux Service so it behaves like running Tentacle via the console in tests.

The tests were already flakey and then broken with https://github.com/OctopusDeploy/OctopusTentacle/pull/738 

No muted tests with the last run of this PR so looking like it has solved the main issues

image

I can still get failures on Ubuntu locally if I run to failure, so it's not 100% reliable but it shifts the issue on to the restart `--stop --start` performed during the test, rather than the failure occurring when trying to setup and start the service initally.
# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Tests.Integration/Support/TentacleBuilder.cs
+++ b/source/Octopus.Tentacle.Tests.Integration/Support/TentacleBuilder.cs
@@ -157,9 +157,10 @@ namespace Octopus.Tentacle.Tests.Integration.Support
                         var serviceInstalled = false;
                         var serviceStarted = false;
 
-                        await RunTentacleCommandOutOfProcess(
+                        await RunCommandOutOfProcess(
                             tentacleExe, 
                             new[] {"service", "--install", $"--instance={instanceName}"}, 
+                            "Tentacle",
                             tempDirectory,
                             s =>
                             {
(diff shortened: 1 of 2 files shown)


source/Octopus.Tentacle.Tests.Integration/Support/TentacleBuilder.cs
```diff
@@ -206,52 +211,15 @@ public ITentacleBuilder InstallAsAService()
                         }
                         else
                         {
-                            logger.Warning("The Tentacle failed to start correctly. Trying Again. Last Log File Content");
```
nathanwoctopusdeploy, line 209: Removing the retry logic as this was just hiding the underlying testing issue

source/Octopus.Tentacle/Configuration/Instances/ApplicationInstanceStore.cs
```diff
@@ -37,9 +37,11 @@ class ApplicationInstanceStore : IApplicationInstanceStore
             else
             {
                 machineConfigurationHomeDirectory = PlatformDetection.IsRunningOnWindows ? 
-                    Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData), "Octopus") : 
-                    "/etc/octopus";
+                    Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData), "Octopus") :
+                "/etc/octopus";
             }
+
+            log.Verbose("Machine configuration home directory is " + machineConfigurationHomeDirectory);
```
nathanwoctopusdeploy, line 44: Useful for test debugging, comes out as DEBUG level into the log file

nathanwoctopusdeploy:
[sc-66745]

source/Octopus.Tentacle.Tests.Integration/Support/TentacleBuilder.cs
```diff
@@ -297,6 +265,40 @@ public ITentacleBuilder InstallAsAService()
             return (runningTentacle, serviceUri);
         }
 
+        async Task SetEnvironmentVariablesForService(string instanceName, TemporaryDirectory tempDirectory, ILogger logger, CancellationToken cancellationToken)
```
sburmanoctopus, line 268: I like that this sets it for the service only (so it won't affect other tests).

It would be nice to not have to poke the registry. But at least it should clean up after itself when we uninstall the service.

An alternative could be to provide a command line argument to Tentacle to allow it to set variables for itself (maybe?). But a bridge too far for fixing tests 🙂

sburmanoctopus (approved):
LGTM

PR #742, opened 2023-12-14
Cancel Kubernetes Job when CancelScript is called

APErebus:
# Background

Currently, when a script being executed via Kubernetes Job is cancelled (via the CancelScript RPC call), we don't cancel the Job.

# Results

In the `RunningKubernetesJob`, we now attach a cancellation callback to the `scriptCancellationToken`, so we are notified when the cancellation token is cancelled.

This callback spawns a background task which calls the `DeleteJob` method in the `KubernetesJobService`. This marks the job for termination, which triggers a `SIGTERM` signal, allowing for graceful shutdown before the termination command is sent 30s later (`SIGKILL`).

I have adjusted the code so that the `scriptCancellationToken` only cancels the job, not the monitoring of the job (which will now stop watching when the job is deleted)

image

Shortcut story: [sc-67104]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Octopus.Tentacle/Kubernetes/Scripts/RunningKubernetesJob.cs
```diff
@@ -165,7 +179,13 @@ async Task<int> CheckIfJobHasCompleted(CancellationToken cancellationToken, Canc
             {
                 log.Error(ex);
                 resultStatusCode = 0;
-            }, cancellationToken);
+            }, CancellationToken.None);
+
+            //if the job was killed by cancellation, then we need to change the exit code
```
MissedTheMark, line 184: Good pick-up, would have been easy to forget this 🥇

MissedTheMark (approved):
Looks good!

source/Octopus.Tentacle/Kubernetes/Scripts/RunningKubernetesJob.cs
```diff
@@ -59,41 +63,68 @@ public class RunningKubernetesJob : IRunningScript
...
+                    writer.WriteOutput(ProcessOutputSource.StdOut, "Script execution canceled.");
+                    writer.WriteVerbose($"Cancelling Kubernetes Job '{jobName}'.");
+
+                    //we spawn the pod cancellation on a background thread (as this callback runs synchronously)
+                    Task.Run(async () =>
+                    {
+                        try
+                        {
+                            //to cancel the job we just kill the underlying pod(s)
+                            await podService.DeletePodsForJob(jobName, CancellationToken.None);
```
APErebus, line 99: To cancel the job, we need to delete the pod

source/Octopus.Tentacle/Kubernetes/Scripts/KubernetesJobScriptExecutor.cs
```diff
@@ -29,10 +29,7 @@ public IRunningScript ExecuteOnBackgroundThread(StartScriptCommandV3Alpha comman
         {
             var runningScript = new RunningKubernetesJob(workspace, workspace.CreateLog(), command.ScriptTicket, command.TaskId, cancellationToken, log, scriptStateStore, jobService, containerResolver, appInstanceSelector);
```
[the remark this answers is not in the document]
APErebus: Yeah, good call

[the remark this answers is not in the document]
APErebus: fixed in [e1fbeed](https://github.com/OctopusDeploy/OctopusTentacle/pull/742/commits/e1fbeed6016e5a24ea6ee20d1d1cf1edc5ad9041)

source/Octopus.Tentacle/Kubernetes/Scripts/RunningKubernetesJob.cs
APErebus: Changing to suspending the job, then deleting it

APErebus:
After advice/discussion with @cailyoung, I have changed it again so that the job is suspended (which tells the control plane to terminate the pods), then we delete the job

source/Octopus.Tentacle/Kubernetes/Scripts/RunningKubernetesJob.cs
```diff
@@ -65,35 +66,60 @@ public class RunningKubernetesJob : IRunningScript
...
+                    //we spawn the job cancellation on a background thread (as this callback runs synchronously)
+                    Task.Run(async () =>
+                    {
+                        try
+                        {
+                            writer.WriteVerbose($"Cancelling Kubernetes Job '{jobName}'.");
+                            //first we suspend the job, which terminates the underlying pods
+                            await jobService.SuspendJob(scriptTicket, CancellationToken.None);
+                            //then we delete the job (because we no longer need it)
+                            await jobService.Delete(scriptTicket, CancellationToken.None);
```
MissedTheMark, line 95: +1 for helpful comments like this 👍

APErebus, line 95: new logic is that we suspend the job, then delete it
> When you suspend a Job, any running Pods that don't have a status of Completed will be [terminated](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination). with a SIGTERM signal.

MissedTheMark (approved):
New change looks to be a lot more graceful 👍

PR #743, opened 2023-12-17
Removing the need to unwrap operation cancelled exceptions

sburmanoctopus:
[sc-66743]

# Background

Following on from [this PR](https://github.com/OctopusDeploy/Halibut/pull/582), we now want to use those changes to simplify detecting when Halibut thrown an `OperationCanceledException` using the actual type, instead of unwrapping the exception message.

# Results

## Before

We would have to 'unpack' the exception to see if it was an `OperationCanceledException`

## After

We can now simply check the type of the exception.

This has allowed us to dramatically simplify a lot of what we were doing to see if it was an operation cancellation.

For example, several decorators were created to do this 'unpacking'. These can now be deleted.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


nathanwoctopusdeploy (approved):
LGTM

source/Octopus.Tentacle.Tests.Integration/Support/ExtensionMethods/FluentAssertionExtensionMethods.cs
```diff
@@ -12,22 +11,12 @@ public static AndConstraint<ObjectAssertions> BeRequestCancelledException(this O
         {
             if (rpcCallStage == RpcCallStage.Connecting)
             {
-                return should.Match(x => x is ConnectingRequestCancelledException || IsHalibutWrappedConnectingRequestCancelledException(x as HalibutClientException));
+                return should.Match(x => x is ConnectingRequestCancelledException);
```
sburmanoctopus, line 14: FYI @nathanwoctopusdeploy , I found this (after you approved).

PR #756, opened 2023-12-20
Cancel ScriptExecution without calling CancelScript when the StartScript request has never been transferred to Tentacle

nathanwoctopusdeploy:
# Background

Related to: https://github.com/OctopusDeploy/Issues/issues/8535

Given ScriptExecution has started for a Polling Tentacle that has recently gone offline
And the GetCapabilities response is cached
And RPC retries have started for the StartScript Call 
When the ScriptExecution is cancelled
Then TentacleClient will try and cancel script execution on Tentacle by calling CancelScript

In this scenario, we can guarantee that StartScript has never been transferred to Tentacle so the script is definitely not running.
Tentacle Client should be able to cancel any queued request and immediately return.

## Before

Cancelling ScriptExecution would start trying to cancel the script on Tentacle by calling CancelScript

## After

Cancelling ScriptExecution does not try and call CancelScript on Tentacle and instead cancels the queued request and stops ScriptExecution immediately.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Tests.Integration/Support/TentacleConfigurationsAttribute.cs
+++ b/source/Octopus.Tentacle.Tests.Integration/Support/TentacleConfigurationsAttribute.cs
@@ -79,6 +79,11 @@ namespace Octopus.Tentacle.Tests.Integration.Support
                 tentacleTypes.Add(TentacleType.Listening);
             }
 
+            if (!tentacleTypes.Any())
+            {
+                throw new ArgumentException("At least one tentacle type to test must be specified");
+            }
+
             List<Version?> versions = new();
 
             if (testCommonVersions)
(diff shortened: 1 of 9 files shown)


nathanwoctopusdeploy:
[sc-67724]

source/Octopus.Tentacle.Tests.Integration/Support/TentacleConfigurationsAttribute.cs
```diff
@@ -53,9 +63,22 @@ static class TentacleConfigurationTestCases
...
         {
-            var tentacleTypes = new[] { TentacleType.Listening, TentacleType.Polling };
+            var tentacleTypes = new List<TentacleType>();
+
+            if (testPolling)
+            {
+                tentacleTypes.Add(TentacleType.Polling);
+            }
+
+            if (testListening)
```
sburmanoctopus, line 77: Nit: Do you think it's worth ensuring we've added at least one of `testPolling` or `testListening`?

sburmanoctopus (approved):
LGTM

source/Octopus.Tentacle.Tests.Integration/Support/TentacleConfigurationsAttribute.cs
nathanwoctopusdeploy: 100% - Update pushed

PR #761, opened 2024-01-15
Remove unecessary references to NuGet fork in Octopus.Manager.Tentacle

zentron:
Looking at bringing back to life the work to upgrade\deprecate our NuGet fork. Looking around I found that we are referencing the assemblies in `Octopus.Manager.Tentacle` which I can't imagine why it would be needed.


zentron:
Issuing PR to run all the tests. if the tests pass Id say its good to go

PR #763, opened 2024-01-16
Move Octopus.Configuration interfaces and enum into tentacle and update references

scme0:
# Background

As part of #project-k8s-agent, I was adding the ability for Tentacle to save and source configuration from a configmap. During this work, I found that the `Octopus.Configuration` project has been merged back into `Octopus.Server` and the repo in github has been archived.

# Results

 This PR just removes the now defunct nuget package and moves the two interfaces and one enum into Tentacle. It's a relatively simple change, but it touches quite a few files to remove the old namespace.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle/Configuration/TentacleConfiguration.cs
+++ b/source/Octopus.Tentacle/Configuration/TentacleConfiguration.cs
@@ -1,3 +1,4 @@
+#nullable enable
 using System;
 using System.Collections.Generic;
 using System.IO;
--- a/source/Octopus.Tentacle/Configuration/TentacleConfiguration.cs
+++ b/source/Octopus.Tentacle/Configuration/TentacleConfiguration.cs
@@ -1,4 +1,3 @@
-#nullable enable
 using System;
 using System.Collections.Generic;
 using System.IO;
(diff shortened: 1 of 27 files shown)


source/Octopus.Tentacle/Configuration/TentacleConfiguration.cs
```diff
@@ -1,4 +1,3 @@
-#nullable enable
```
LukeButters, line 1: Was this intentional?

LukeButters (approved):
LGTM,

Just double check if that removal of nullable was intentional, if so a small note on why would be worthwhile.

source/Octopus.Tentacle/Configuration/TentacleConfiguration.cs
scme0: Ah yeah I think that was rider deeming it unnecessary but I'll undo that change to keep this PR focused.

LukeButters: Either way is fine :D

PR #663, opened 2023-11-14
Migrate all integration tests to use new decorators

APErebus:
# Background

Migrates all the tests from the old decorators to the new proxy-based decorators and per-tentacle versioned  tests.

# Results

I also removed all the old decorators and TCP workaround extension methods

Shortcut story: [sc-62862]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

- [ ] I have read [How we use GitHub Issues](https://github.com/OctopusDeploy/Issues/blob/master/docs/CONTRIBUTING.internal.md) for help deciding when and where it's appropriate to make an issue.
- [x] I have considered informing or consulting the right people, according to the [ownership map](https://whimsical.com/ownership-map-NzbiD4HJyvhC9jNJNfS6TG).
- [x] I have considered appropriate testing for my change.

PR #768, opened 2024-01-17
Add back non generic test decorators

LukeButters:
# Background

https://github.com/OctopusDeploy/OctopusTentacle/pull/663 replaced all decorators with a generic proxy decorator. That works great for things like logging ALL calls or counting ALL calls. However it is very difficult to use when the test needs to do something special with one call because:
* The proxy decorator has no autocompletion making it hard to tab your way through.
* The proxy decorator has no type safety.
* The proxy decorator assumes each RPC has exactly one parameter.

The change here is to add back some of the original test decorators, specifically it adds back the decorators that allow of doing something before/after a specific method. The result is:
* Decorations that want to apply to all calls is automatically done through the proxy decorator.
* Decorations that want to apply to just one call is easier to do in, since autocompletion and type safety is enforced. Users (ie test writers) can tab complete their way to glory.

This PR doesn't stop the proxy decorator from being used as something that overrides a single call, a future PR will need to be done to do that. For now we need this PR to unblock https://github.com/OctopusDeploy/OctopusTentacle/pull/769

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Client/ITentacleServiceDecoratorFactory.cs
+++ b/source/Octopus.Tentacle.Client/ITentacleServiceDecoratorFactory.cs
@@ -2,7 +2,7 @@ using Octopus.Tentacle.Contracts.ClientServices;
 
 namespace Octopus.Tentacle.Client
 {
-    public interface ITentacleServiceDecoratorFactory
+    interface ITentacleServiceDecoratorFactory
     {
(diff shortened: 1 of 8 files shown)


evolutionise (approved):
LGTM - haven't dived too deeply into the specifics based on the understanding that we're restoring existing, tried-and-testing decorators

PR #767, opened 2024-01-17
increased the timeout for completion phase

benPearce1:
[sc-68281]

# Background

The completion phase was recently [changed](https://github.com/OctopusDeploy/OctopusTentacle/commit/be322a02d8b267c42fbedfcc79b96191f6cf0943#diff-cc31b0e7a7349a8280a9d04aa2f689706d3543aad6d0ecf6e5775ef492a75983) to be async with a cancellation token.
The default time on the cancellation token is 1 minute, which in some cases was being hit when the size of the workspace was very large or had many files.

[Customer reported](https://octopus.zendesk.com/agent/tickets/167317) that this problem only started occurring after upgrading to Tentacle version 8.1.558. Previous versions did not have this issue.

# Results

## Before

The timestamps in the below task log clearly show a 1 minute gap between the end of the task execution and the timeout of the cancellation token.
```
15:53:30   Verbose  |       Process C:\Windows\system32\WindowsPowershell\v1.0\PowerShell.exe in C:\Octopus\OctopusCloud-Tentacle\Work\DGCCBsHs606XsqM5b9FLA exited with code 0
15:54:30   Warning  |       Failed to cleanup the script working directory on Tentacle
15:54:30   Verbose  |       The task was canceled
                    |       The Request was cancelled while Transferring.
                    |       Halibut.Exceptions.TransferringRequestCancelledException: The Request was cancelled while Transferring.
                    |       ---> System.Threading.Tasks.TaskCanceledException: A task was canceled.
                    |       at Halibut.ServiceModel.PendingRequestQueueAsync.PendingRequest.WaitForResponseToBeSet(Nullable`1 timeout, Boolean cancelTheRequestWhenTransferHasBegun, CancellationToken cancellationToken)
                    |       --- End of inner exception stack trace ---
                    |       at Halibut.ServiceModel.PendingRequestQueueAsync.PendingRequest.WaitForResponseToBeSet(Nullable`1 timeout, Boolean cancelTheRequestWhenTransferHasBegun, CancellationToken cancellationToken)
                    |       at Halibut.ServiceModel.PendingRequestQueueAsync.PendingRequest.WaitUntilComplete(CancellationToken cancellationToken)
                    |       at Halibut.ServiceModel.PendingRequestQueueAsync.QueueAndWaitAsync(RequestMessage request, CancellationToken cancellationToken)
[...]

PR #769, opened 2024-01-18
Client should wait for script cleanup call

LukeButters:
# Background

[SC-68685]

The tentacle client had a bug in which it would not wait longer than 1 min to wait for the response to `CompleteScriptAsync`. This meant that although Tentacle would (assuming no errors occure on tentacle itself) complete the clean up. The tentacle client would report that the cleanup was failed to be cleaned up.

Fixes: https://github.com/OctopusDeploy/OctopusTentacle/issues/770

Relates to: https://github.com/OctopusDeploy/OctopusTentacle/pull/767
# Results

We now don't set a 1 min timeout when waiting for the `CompleteScriptAsync` to complete, instead the normal Halibut timeouts will apply.

Now when a call to `CompleteScriptAsync` takes a couple of minutes, we will wait for the RPC and we will NOT report it failed (unless it actually did).

This fixes both v2 and v3Alpha Script orchestrators in the Tentacle Client.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Client/Scripts/ScriptServiceV2Orchestrator.cs
+++ b/source/Octopus.Tentacle.Client/Scripts/ScriptServiceV2Orchestrator.cs
@@ -200,7 +200,10 @@ namespace Octopus.Tentacle.Client.Scripts
             {
                 // Finish performs a best effort cleanup of the Workspace on Tentacle
                 // If we are cancelling script execution we abandon a call to complete script after a period of time
-                using var completeScriptCancellationTokenSource = new CancellationTokenSource(onCancellationAbandonCompleteScriptAfter);
+                
+                using var completeScriptCancellationTokenSource = new CancellationTokenSource();
+
+                await using var _ = scriptExecutionCancellationToken.Register(() => completeScriptCancellationTokenSource.CancelAfter(onCancellationAbandonCompleteScriptAfter));
 
                 await rpcCallExecutor.ExecuteWithNoRetries(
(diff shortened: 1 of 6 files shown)


source/Octopus.Tentacle.Client/Scripts/ScriptServiceV2Orchestrator.cs
```diff
@@ -200,7 +200,10 @@ protected override async Task<ScriptStatusResponseV2> Finish(ScriptStatusRespons
             {
                 // Finish performs a best effort cleanup of the Workspace on Tentacle
                 // If we are cancelling script execution we abandon a call to complete script after a period of time
-                using var completeScriptCancellationTokenSource = new CancellationTokenSource(onCancellationAbandonCompleteScriptAfter);
+                
+                using var completeScriptCancellationTokenSource = new CancellationTokenSource();
```
evolutionise, line 204: If only someone had given an in-depth talk on the subtle differences here recently

LukeButters: yeah, I couldn't see a way of not using register that was also nice :(

It should be ok since in the registration we are not doing a extremely long lived operation and we don't pass in something that is going to hold lots of references to things we didn't expect to hold references to.

Also this PR is moved to https://github.com/OctopusDeploy/OctopusTentacle/pull/775

as I made a mistake

PR #772, opened 2024-01-19
Add encryption to configmap store

scme0:
# Background

Part of #project-k8s-agent

In the recent PR #674 I updated tentacle to allow it to store it's configuraiton in a kubernetes configmap instead of a file. This bypassed any machine key encryption which is what this PR adds.

# Results

We need to encrypt certain items that are added to the configuration store using a machine key which is generated when the tentacle is first configured.

We're adding the `KubernetesSecretService` 🕵️ to provide access to kubernetes secrets so we can store the machine key there.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle/Configuration/Crypto/KubernetesMachineKeyEncryptor.cs
+++ b/source/Octopus.Tentacle/Configuration/Crypto/KubernetesMachineKeyEncryptor.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Security.Cryptography;
 using System.Text;
+using System.Threading.Tasks;
 using k8s;
 using k8s.Models;
 using Octopus.Diagnostics;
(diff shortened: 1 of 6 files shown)


source/Octopus.Tentacle/Configuration/Crypto/KubernetesMachineKeyEncryptor.cs
```diff
@@ -0,0 +1,70 @@
...
+                var (key, iv) = GenerateMachineKey(log);
+                var data = new Dictionary<string, byte[]> { { MachineKeyName, machineKey = key }, { MachineIvName, machineIv = iv } };
+
+                kubernetesSecretService.Patch(SecretName, data, CancellationToken.None).GetAwaiter().GetResult();
+            }
+
+            if (machineKey == null || machineIv == null)
+            {
+                throw new InvalidOperationException("Unable to retrieve or create a machine key for encryption.");
+            }
```
APErebus, line 40: I probably would caution against putting service/api calls like this in a constructor.

I'd probably go with something in `Encrypt`/`Decrypt` that makes sure the machine key & IV are correctly loaded before encryption/decryption.

APErebus (changes requested):

source/Octopus.Tentacle/Configuration/Crypto/KubernetesMachineKeyEncryptor.cs
scme0: Good callout, I realised that I did this in the ConfigMapKeyValueStore as well, so I've updated the other PR so that the configMap is loaded Lazily.

scme0:
[sc-69174]

APErebus (approved):
Nice work 👍

APErebus:
@scme0 When this PR merges and we have a `main` build number for tentacle with this change, can you also update the helm chart so it uses this build? 🙇

PR #780, opened 2024-01-29
Add script for running the Fedora integration tests in a container

LukeButters:
# Background

Targeting `fedora:39` this adds a script which can be run by team city in a Fedora container, to run the integration tests.

The change is to add a new directory containing scripts needed to setup the fedora container to run the tests as well as a script to run the test. This means the config in teamcity is reduced to just running a single step, while the details of the step are version controlled.

Ideally the container image and parameters passed to docker would be in git as well, however I could not find a way to make that work.

Merging this PR does nothing to the build, we will need to add the [Integration Test: Containerised Fedora (not sudo)](https://build.octopushq.com/admin/editBuild.html?id=buildType:TeamFireAndMotion_OctopusTentacle_VLatest_LukeSIntegrationTestFedoraNotSudo) to a dependency of the integration gate first. Which must happen after the PR is merged. (If only everything was in git)

# How to review this PR

This appears to work when manually running the build and reduces the integration test time from 27m to 6m.
# Pre-requisites


build/container-test-scripts/integration/fedora/dotnet-install.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -o xtrace
+set -e
+
+# See https://github.com/OctopusDeploy/BuildAgentAutomation/blob/main/Common/shared/dotnet-install.sh
+curl --silent --fail -L -O https://dot.net/v1/dotnet-install.sh || exit 1
+chmod +x dotnet-install.sh || exit 1
+
+./dotnet-install.sh --channel 6.0
```
evolutionise, line 9: Wondering if this would be easier to update/find when trying to update if it were sourced from a build environment variable? Do we do something similar elsewhere?

evolutionise: It could even go in our script to set dotnet env vars - that feels discoverable

evolutionise (approved):
Minor suggestion, otherwise LGTM

build/container-test-scripts/integration/fedora/dotnet-install.sh
LukeButters: We currently set the version of tentacle in its csproj file, I would opt for keeping as much as we can in git rather than spread across git and teamcity.

LukeButters: I don't know if setting the channel in the dotnet env vars file would be more discoverable.

LukeButters: Then again on the other hand are we already setting 6.0 as a env var that would need to be updated anyway...

LukeButters: had a look and could not see one that we would hook into. We set agent parameters requirements  e.g. netcore must contain 6.0. But we don't actually get given net6

PR #781, opened 2024-01-30
Multi-Arch k8s tentacle build

scme0:
# Background

Part of #project-k8s-agent

Developers on Apple Silicon have trouble with the `amd64` version of kubernetes tentacle because kubernetes won't automatically pull an `amd64` platform image for an `arm64` platform (like apple silicon).

[sc-66060]

# Results

I've updated the build to use `docker buildx build` which allows you to specify many different platforms and produces a multi-platform image and I've moved the setup for this into a nuke build target instead of a docker-compose file plus magic environment variables set in the team city build configuration command.

I've added a new [build configuration](https://build.octopushq.com/buildConfiguration/TeamFireAndMotion_OctopusTentacle_VLatest_BuildKubernetesTentacleMultiArchDockerImage?branch=pull/781&buildTypeTab=overview&mode=builds), which builds and pushes the multi-arch image using the new nuke target.

**Note:** I can't test that the powershell local build script still works as expected but I think it should generally be ok.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


docker/kubernetes-tentacle/Dockerfile
```diff
@@ -2,19 +2,21 @@ FROM mcr.microsoft.com/dotnet/runtime-deps:6.0
 
 ARG BUILD_NUMBER
 ARG BUILD_DATE
-ARG BUILD_ARCH
+ARG TARGETARCH
+ARG TARGETOS
+ARG TARGETVARIANT
```
scme0, line 7: These are automagically set by docker buildx when building for each platform.

docker-compose.build.yml
```diff
@@ -18,7 +18,9 @@ services:
             args:
                 BUILD_DATE: ${BUILD_DATE:?err}
                 BUILD_NUMBER: ${BUILD_NUMBER:?err}
-                BUILD_ARCH: ${BUILD_ARCH:?err}
+                TARGETARCH: ${BUILD_ARCH:?err}
+                TARGETOS: linux
+                TARGETVARIANT: ${BUILD_VARIANT}
```
scme0, line 23: I've updated the docker-compose definition so it should still work.

scme0: `BUILD_VARIANT` is optional as it's only used for armhf (the arch would be `arm` and the variant would be `v7` for instance).

build/Build.Pack.cs
```diff
@@ -454,4 +476,37 @@ void PackTarballs(string runtimeId)
             ArtifactsDirectory / "zip",
             $"tentacle-{FullSemVer}-{NetCore}-{runtimeId}.tar.gz");
     }
+
+    void BuildAndPushOrLoadKubernetesTentacleContainerImage(bool push, bool load)
+    {
+        DockerTasks.DockerBuildxBuild(settings =>
+            settings.AddBuildArg($"BUILD_NUMBER={FullSemVer}", $"BUILD_DATE={DateTime.UtcNow:O}")
+                .SetPlatform("linux/arm64,linux/amd64")
```
scme0, line 484: I've removed linux/arm/v7 for now because I was having some trouble with docker failing to run commands when building the container. I'll take a look later but it's not really too important.

build-k8s-docker-local.ps1
```diff
@@ -1,6 +1,9 @@
 param (    
     [string]
     $BuildArch = "amd64",
+
+    [string]
+    $BuildArchVariant = "",
```
scme0, line 6: This will probably never be used, but for completeness 🤷

scme0:
Note that the build will show `Build: Kubernetes Tentacle Docker image` failing, which is the old build configuration and doesn't work anymore. I haven't disabled it as it is still used for the main branch, but I've set it to not fail the Full Chain build if it fails. Once I merge this PR I'll remove that build configuration completely.

The new build configuration `Build: Kubernetes Tentacle Multi-Arch Docker image` is ✅

APErebus (approved):
awesome stuff 🚀

PR #782, opened 2024-01-31
Adds a TentacleBackwardsCompatibility test category

LukeButters:
# Background

Integration tests which test with an older version of tentacle are now categorised with `TentacleBackwardsCompatibility`.

Doing so will make it possible to not test this category on some teamcity OS VM agents.

[SC-65861]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

- [ ] I have read [How we use GitHub Issues](https://github.com/OctopusDeploy/Issues/blob/main/docs/CONTRIBUTING.internal.md) for help deciding when and where it's appropriate to make an issue.
- [ ] I have considered informing or consulting the right people, according to the [ownership map](https://whimsical.com/ownership-map-NzbiD4HJyvhC9jNJNfS6TG).
- [ ] I have considered appropriate testing for my change.

PR #783, opened 2024-01-31
Add scripts for running integration tests on Team City VM agents.

LukeButters:
# Background

[SC-65861]

This just adds the scripts required for running integration tests on TC VM agents.

The intention is that the TC step calls a single script, and the script contains ALL of the complexity. That way if we need to change something we change it in a PR rather than on TC (skipping a PR).

The scripts:
* gets the latest dotnet 6, since sometimes the agents have older versions which causes problems.
* set env vars to use the dotnet version we just installed.
* runs the integration tests, without sudo since VMs don't support sudo. This is what we are already doing.

Once these scripts are in it should be possible to change the script to stop testing the [TentacleBackwardsCompatibility](https://github.com/OctopusDeploy/OctopusTentacle/pull/782) category in a future PR.

Also the TC config will need to be changed to run this script after this is merged.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


build/test-scripts/integration/common/run-integration-tests-on-vm-agents.sh
```diff
@@ -0,0 +1,17 @@
...
+# Runs integration tests on VM backed agents.
+# Currently VM agents don't supply sudo, so tests that need that are not run.
+
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
+
+$SCRIPT_DIR/setup-vm-agent.sh
+
+source $SCRIPT_DIR/set-dotnet-envvars.sh
```
LukeButters, line 14: We must do this here since this sets the envvars for the dotnet call below.

samdanaei (approved):

PR #790, opened 2024-02-05
Add new timeout for job watch api call

APErebus:
# Background

We identified an issue where if the job took longer to run than the job TTL (default 60s), then the `jobService.Watch` call would timeout, which would "complete" the job and no longer record any logs.

This would happen most obviously the first time you acquire an execution image and the job would complete in the octopus UI before the job would execute.

# Results

Introduces a new environment variable `OCTOPUS__K8STENTACLE__JOBMONITORTIMEOUT` for adjusting the timeout of this watch call. Default's to 30min if not specified.

The watch still finishes when signalled or when the job has been deleted.

Shortcut story: [sc-70357]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle/Kubernetes/KubernetesConfig.cs
+++ b/source/Octopus.Tentacle/Kubernetes/KubernetesConfig.cs
@@ -11,6 +11,7 @@ namespace Octopus.Tentacle.Kubernetes
         public static string JobVolumeYaml => GetRequiredEnvVar($"{EnvVarPrefix}__JOBVOLUMEYAML", "Unable to determine Kubernetes Job volume yaml.");
         public static bool UseJobs => bool.TryParse(Environment.GetEnvironmentVariable($"{EnvVarPrefix}__USEJOBS"), out var useJobs) && useJobs;
         public static int JobTtlSeconds => int.TryParse(Environment.GetEnvironmentVariable($"{EnvVarPrefix}__JOBTTL"), out var jobTtl) ? jobTtl : 60; //Default 1min
+        public static int JobMonitorTimeoutSeconds => int.TryParse(Environment.GetEnvironmentVariable($"{EnvVarPrefix}__JOBMONITORTIMEOUT"), out var jobMonitorTimeout) ? jobMonitorTimeout : 1800; //30min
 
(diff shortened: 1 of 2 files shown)


source/Octopus.Tentacle/Kubernetes/KubernetesConfig.cs
```diff
@@ -11,6 +11,7 @@ public static class KubernetesConfig
         public static string JobVolumeYaml => GetRequiredEnvVar($"{EnvVarPrefix}__JOBVOLUMEYAML", "Unable to determine Kubernetes Job volume yaml.");
         public static bool UseJobs => bool.TryParse(Environment.GetEnvironmentVariable($"{EnvVarPrefix}__USEJOBS"), out var useJobs) && useJobs;
         public static int JobTtlSeconds => int.TryParse(Environment.GetEnvironmentVariable($"{EnvVarPrefix}__JOBTTL"), out var jobTtl) ? jobTtl : 60; //Default 1min
+        public static int JobMonitorTimeoutSeconds => int.TryParse(Environment.GetEnvironmentVariable($"{EnvVarPrefix}__JOBMONITORTIMEOUT"), out var jobMonitorTimeout) ? jobMonitorTimeout : 1800; //30min
```
flin-8, line 14: This has me thinking ... we have deployment steps that run for hours (copying Azure files), we might need a longer timeout

APErebus: yeah. It's a fair concern I think when we are running as a worker. My gut feeling is let's unblock the k8s issue now and look at it more comprehensively when we get to this running as a worker.

flin-8 (approved):
Merge this for now, but let's ensure we track the fact that we have this known issue here.

source/Octopus.Tentacle/Kubernetes/KubernetesConfig.cs
flin-8: When we deploy a container (like Server), would it be sad if the upgrade migration took > 30min?

APErebus: It would depend if we had something like "is atomic" on, where the script waits until the pod is fully running

Issue #527, opened 2023-06-21
Add Support for Polling 443 Tentacles to Windows Installer & Linux configure-tentacle.sh  


FinnianDempsey:
### Team

- [X] I've assigned a team label to this issue

### What happened?

Currently the Tentacle installers don't support the new feature Polling Tentacles over Standard Ports: [https://octopus.com/docs/infrastructure/deployment-targets/tentacle/polling-tentacles-over-port-443](https://octopus.com/docs/infrastructure/deployment-targets/tentacle/polling-tentacles-over-port-443)

The windows installer and linux configuration.sh both create scripts that use `--server-comms-port "10943"` without allowing for `--server-comms-address` to be specified.

### Reproduction

Use the Windows Installer or Linux configure-tentacle.sh to configure a Polling Tentacle and see `--server-comms-port` used in the generated scripts 

### Error and Stacktrace

```shell
N/A
```

### More Information

[Slack - Internal](https://octopusdeploy.slack.com/archives/C27FNL3QW/p1687147913057339)

### Workaround

Copy the generated installer scripts and modify the `register-with` (or `register-worker`) command to use `--server-comms-address` instead of `--server-comms-port`

E.g. [https://octopus.com/docs/infrastructure/deployment-targets/tentacle/polling-tentacles-over-port-443#registering-a-new-tentacle](https://octopus.com/docs/infrastructure/deployment-targets/tentacle/polling-tentacles-over-port-443#registering-a-new-tentacle)
```
.\Tentacle register-with --instance MyInstance --server "https://<yoururl>.octopus.app" --server-comms-address "https://polling.<yoururl>.octopus.app" --comms-style TentacleActive --apiKey "API-YOURKEY" --environment "Test" --role "Web"
```

Feel free to reach out to [support@octopus.com](support@octopus.com) with any questions!

PR #792, opened 2024-02-06
The linux configure-tentacle.sh script now supports setting comms ports other than 10943.

LukeButters:
# Background

[SC-68270]

Fixes: https://github.com/OctopusDeploy/OctopusTentacle/issues/793

Related to: https://github.com/OctopusDeploy/OctopusTentacle/issues/527

# Results

## Before

The script would always assume `10943` as the comms port.
## After
It now asks if the comms port is at the same or different address:

```
Is the comms port on the same domain as the Octopus Server : 1) Yes or 2) No (default 1): 2
What is the Octopus Server comms address including port e.g. 'https://polling.<yoururl>.octopus.app:443' ?: https://127.0.0.1:11943
```

which results in:
```
sudo /opt/octopus/tentacle/Tentacle register-with --instance "Tentacle" --server "http://127.0.0.1:8066" --name "aero2" --comms-style "TentacleActive" --server-comms-address "https://127.0.0.1:11943" --apiKey "API-XXXXXXXXXXXXXXXXXXXXXXXXXX" --space "Default" --environment "bready"  --role "roll" 
```

If the address is the same we also ask what port:

```
Is the comms port on the same domain as the Octopus Server : 1) Yes or 2) No (default 1): 
What port is the Octopus Server comms port? (10943): 11943
```
and the resulting command:
```
sudo /opt/octopus/tentacle/Tentacle register-with --instance "aa-luke-2" --server "http://127.0.0.1:8066" --name "aero4" --comms-style "TentacleActive" --server-comms-port "11943" --apiKey "API-XXXXXXXXXXXXXXXXXXXXXXXXXX" --space "Default" --environment "Dev"  --role "bread" 
```

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/linux-packages/content/configure-tentacle.sh
+++ b/linux-packages/content/configure-tentacle.sh
@@ -150,8 +150,24 @@ function setupPollingTentacle {
     read -p "What name would you like to register this Tentacle with? ($displayname): " displaynameinput
     displayname=$(assignNonEmptyValue "$displaynameinput" $displayname)
 
-    read -p "What port is the Octopus Server comms port? ($servercommsport): " servercommsportinput
-    servercommsport=$(assignNonEmptyValue "$servercommsportinput" $servercommsport)
+    $doesconnecttosamedomain=1
+    read -p 'Is the comms port on the same domain as the Octopus Server : 1) Yes or 2) No (default 1): ' doesconnecttosamedomain
+
+    
+    
+    case $doesconnecttosamedomain in
+        2)
+            read -p "What is the Octopus Server comms address including port e.g. 'https://polling.<yoururl>.octopus.app:443' ?: " commsAddress
+            commsAddressOrPortArgs="--server-comms-address \"$commsAddress\""
+            ;;
+            
+        *)
+            read -p "What port is the Octopus Server comms port? ($servercommsport): " servercommsportinput
+            servercommsport=$(assignNonEmptyValue "$servercommsportinput" $servercommsport)
+
+            commsAddressOrPortArgs="--server-comms-port \"$servercommsport\""
+            ;;
+    esac
 
     case $machinetype in
         2 | worker)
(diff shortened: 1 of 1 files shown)


linux-packages/content/configure-tentacle.sh
```diff
@@ -100,6 +100,7 @@ function setupPollingTentacle {
     rolesstring=""
     workerpoolsstring=""
     machinetype=1
+    servercommsport=10943
```
raynhamdev, line 103: Out of curiosity, do we know why this was strictly only set to `10943` before?

LukeButters: Probably because it is the default.

linux-packages/content/configure-tentacle.sh
```diff
@@ -149,6 +150,25 @@ function setupPollingTentacle {
...
     displayname=$(assignNonEmptyValue "$displaynameinput" $displayname)
 
+    $doesconnecttosamedomain=1
+    read -p 'Is the comms port on the same domain as the Octopus Server : 1) Yes or 2) No (default 1): ' doesconnecttosamedomain
+
+    
+    
+    case $doesconnecttosamedomain in
+        2)
+            read -p "What is the Octopus Server comms address including port e.g. 'https://polling.<yoururl>.octopus.app:443' ?: " commsAddress
```
raynhamdev, line 160: [The public doco](https://octopus.com/docs/infrastructure/deployment-targets/tentacle/polling-tentacles-over-port-443) doesn't mention that the port number needs to be specified. Would it work if we do?

LukeButters: I requested it since it makes the bash easier, otherwise we then need to ask for port if a port is not set.

Also yes this does work.

raynhamdev (approved):
LGTM 👍

PR #795, opened 2024-02-06
Refactor InstallTab.cs

raynhamdev:
# Background
We want to add some automated tests to Tentacle Manager as part of the [Tentacle Manager on .NET](https://app.shortcut.com/octopusdeploy/epic/68113/tentacle-manager-on-windows-net-core) project.

We first need to make the Tentacle Manager project testable by automated tests (e.g. NUnit) by separating the UI logic from the business logic. This will allow the automated tests to exercise most of the code (i.e. everything _not_ in a UI component) without the pain of dealing with UI threads and Dispatchers etc.

This PR refactors the InstallTab component in TentacleManager to separate its business logic from the view.

[sc-70325]

# Results

There should be *zero visible changes* from an end-user's perspective. The intent is that the functionality is completely unchanged, but the code base is in a better shape for automated tests.

Summary of the main changes within:
- Renamed `InstallTab` to `ReviewAndRunScriptTabView`: this view purpose isn't only for "Installation" but it is indeed a re-usable view to generate/run scripts. The tab view are currently used for three things: installing tentacle, deleting tentacle, and configuring tentacle proxy.
- Added `ReviewAndRunScriptTabViewModel`: This approach allows us to have a ViewModel that contains the business logic of `ReviewAndRunScriptTabView`.
- Having `ReviewAndRunScriptTabViewModel`  taking a dependency of `IScriptableViewModel`: Because the `ReviewAndRunScriptTabView` is used for multiple purposes, I decided to have it take an `IScriptableViewModel wizardModel` which is the original ViewModel of the task it is supposed to do (eg. install script, delete script, etc.).

# How to review this PR

Quality :heavy_check_mark:

As there are no automated tests, the only way to verify this PR is to manually test the Tentacle Manager. There's not much functionality in here, so testing the following should cover everything:
* Install new instance
* Reinstall existing instance
* Restart service
* Start service
* Stop service
* Update proxy settings
* Delete instance

# Pre-requisites

### DIFF
--- a/source/Octopus.Manager.Tentacle/TentacleConfiguration/SetupWizard/ReviewAndRunScriptTabViewModel.cs
+++ b/source/Octopus.Manager.Tentacle/TentacleConfiguration/SetupWizard/ReviewAndRunScriptTabViewModel.cs
@@ -1,0 +1,94 @@
+﻿using System;
+using System.Collections.Generic;
+using Autofac;
+using FluentValidation;
+using Octopus.Diagnostics;
+using Octopus.Manager.Tentacle.Infrastructure;
+using Octopus.Manager.Tentacle.Util;
+using Octopus.Tentacle.Util;
+
+namespace Octopus.Manager.Tentacle.TentacleConfiguration.SetupWizard
+{
+    public class ReviewAndRunScriptTabViewModel : ViewModel, IScriptableViewModel
+    {
+        readonly ICommandLineRunner commandLineRunner;
+        readonly IScriptableViewModel wizardModel;
+        TextBoxLogger logger;
+
+        public ReviewAndRunScriptTabViewModel(IScriptableViewModel wizardModel, IComponentContext container)
+        {
+            this.wizardModel = wizardModel;
+            InstanceName = wizardModel.InstanceName;
+            Executable = CommandLine.PathToTentacleExe();
+            Validator = CreateValidator();
+            commandLineRunner = container.Resolve<ICommandLineRunner>();
+        }
+        
+        public string InstanceName { get; }
+        
+        public string Executable { get; set; }
+        
+        public bool GenerateAndExecuteScript()
+        {
+            var success = false;
+            try
+            {
+                var script = GenerateScript();
+                ContributeSensitiveValues(logger);
+                success = commandLineRunner.Execute(script, logger);
+            }
+            catch (Exception ex)
+            {
+                logger.Error(ex);
+            }
+            finally
+            {
+                if (!success)
+                {
+                    Rollback();
+                }
+            }
+            
+            return success;
+        }
+            
+        void Rollback()
+        {
+            try
+            {
+                var script = GenerateRollbackScript();
+                commandLineRunner.Execute(script, logger);
+            }
+            catch (Exception ex)
+            {
+                logger.Error(ex);
+            }
+        }
+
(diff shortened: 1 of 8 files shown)


source/Octopus.Manager.Tentacle/TentacleConfiguration/SetupWizard/ReviewAndRunScriptTabViewModel.cs
```diff
@@ -0,0 +1,94 @@
...
+        readonly IScriptableViewModel wizardModel;
+        TextBoxLogger logger;
+
+        public ReviewAndRunScriptTabViewModel(IScriptableViewModel wizardModel, IComponentContext container)
+        {
+            this.wizardModel = wizardModel;
+            InstanceName = wizardModel.InstanceName;
+            Executable = CommandLine.PathToTentacleExe();
+            Validator = CreateValidator();
+            commandLineRunner = container.Resolve<ICommandLineRunner>();
```
acodrington, line 24: As this is the only usage of `IComponentContext` in this class (nice!), please depend on `ICommandLineRunner` directly and remove the usage of the container.

source/Octopus.Manager.Tentacle/TentacleConfiguration/SetupWizard/ReviewAndRunScriptTabViewModel.cs
```diff
@@ -0,0 +1,94 @@
...
+        }
+        
+        public string InstanceName { get; }
+        
+        public string Executable { get; set; }
```
acodrington, line 29: The setter can be removed here, as the property is only set in the constructor.

source/Octopus.Manager.Tentacle/TentacleConfiguration/SetupWizard/Views/ReviewAndRunScriptTabView.cs
```diff
@@ -0,0 +1,89 @@
...
+{
+    public partial class ReviewAndRunScriptTabView
+    {
+        public static readonly DependencyProperty SuccessMessageProperty = DependencyProperty.Register("SuccessMessage", typeof(string), typeof(ReviewAndRunScriptTabView), new PropertyMetadata(null));
+        public static readonly DependencyProperty ReadyMessageProperty = DependencyProperty.Register("ReadyMessage", typeof(string), typeof(ReviewAndRunScriptTabView), new PropertyMetadata(null));
+        public static readonly DependencyProperty TitleProperty = DependencyProperty.Register("Title", typeof(string), typeof(ReviewAndRunScriptTabView), new PropertyMetadata("Install"));
+        public static readonly DependencyProperty ExecuteButtonTextProperty = DependencyProperty.Register("ExecuteButtonText", typeof(string), typeof(ReviewAndRunScriptTabView), new PropertyMetadata("INSTALL"));
+        readonly ReviewAndRunScriptTabViewModel viewModel;
+
+        public ReviewAndRunScriptTabView(ReviewAndRunScriptTabViewModel viewModel, ContentControl additionalContent = null)
```
acodrington, line 19: I don't think the optional `additionalContent` param is ever used. Please confirm and remove if possible.

acodrington (changes requested):
Nice work! Just a couple of small change requests, but otherwise all good.

acodrington (approved):
Nice one!

PR #799, opened 2024-02-09
Allowing callers of tentacle client to drive orchestration themselves

sburmanoctopus:
# Background

# Results

## Before

Before

## After

After

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Octopus.Tentacle.Client/Scripts/IScriptOrchestratorFactory.cs
```diff
@@ -1,11 +0,0 @@
-﻿using System;
-using System.Threading;
-using System.Threading.Tasks;
-
-namespace Octopus.Tentacle.Client.Scripts
```
sburmanoctopus, line 5: I didn't see the use of this.

source/Octopus.Tentacle.Contracts/ScriptServiceV3Alpha/GetStatusCommandV3Alpha.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+
+namespace Octopus.Tentacle.Contracts.ScriptServiceV3Alpha
+{
+    public class GetStatusCommandV3Alpha
```
LukeButters, line 5: Do these classes need to be tied to what is going over the wire, maybe it could just be `GetStatusCommand` no v3.

source/Octopus.Tentacle.Client/TentacleClient.cs
```diff
@@ -191,11 +193,11 @@ async Task<DataStream> DownloadFileAction(CancellationToken ct)
                     clientOptions,
                     logger);
 
-                var orchestrator = await factory.CreateOrchestrator(scriptExecutionCancellationToken);
+                var orchestrator = factory.CreateOrchestrator(scriptServiceVersion);
```
LukeButters, line 196: @sburmanoctopus Could we make it so that the "in memory" version of execute a scripts calls the below methods. That way the client itself gives an example of how to call the methods below and we always test those methods.

source/Octopus.Tentacle.Client/TentacleClient.cs
```diff
@@ -209,6 +211,136 @@ async Task<DataStream> DownloadFileAction(CancellationToken ct)
...
+
+        public async Task<ScriptStatusResponseV3Alpha> GetStatus(
+            GetStatusCommandV3Alpha getStatusCommand,
+            ILog logger,
+            CancellationToken scriptExecutionCancellationToken)
+        {
+            //TODO: sast - think about metrics. This isn't right, so we have ignored it for now. But we need to ask what we want to report on in an async event based world.
+            var operationMetricsBuilder = ClientOperationMetricsBuilder.Start();
+
+            var scriptServiceVersion = await DetermineScriptServiceVersionToUse(operationMetricsBuilder, logger, scriptExecutionCancellationToken);
```
LukeButters, line 249: Could this be cached within that second object that could be returned by start script. Otherwise we are very much relying on halibut caching to work. If we don't cache this we will need to ensure we don't log too much since `DetermineScriptServiceVersionToUse` might be logging to task log.

source/Octopus.Tentacle.Client/TentacleClient.cs
```diff
@@ -209,6 +211,136 @@ async Task<DataStream> DownloadFileAction(CancellationToken ct)
             }
         }
 
+
+        public async Task<ScriptStatusResponseV3Alpha> StartScript(
```
LukeButters, line 215: Would it be possible to return something that is not the type that is received over the wire.

Perhaps two things distinct objects should be returned here
 - The interesting data returned e.g. logs.
 - An object which can be passed to `GetStatus` containing everything required to do the next `GetStatus` call. This object could additionally have some extension methods for serialization and deserialization, which is all taken care of in the client (and tested in the client).

source/Octopus.Tentacle.Client/TentacleClient.cs
```diff
@@ -209,6 +211,136 @@ async Task<DataStream> DownloadFileAction(CancellationToken ct)
...
+                OnCancellationAbandonCompleteScriptAfter,
+                clientOptions,
+                logger);
+
+            var executor = factory.CreateExecutor(scriptServiceVersion);
+
+            var result = await executor.GetStatus(getStatusCommand, scriptExecutionCancellationToken);
+
+            return result;
+        }
```
LukeButters, line 266: is the caller of this method now responsible for delays between calls?

source/Octopus.Tentacle.Contracts/ScriptServiceV3Alpha/GetStatusCommandV3Alpha.cs
sburmanoctopus: They don't need to be. I wasn't sure what convention we were following.

I guess it depends whether we want the versioning of all commands to increment together (with `StartScriptCommandV3Alpha` etc.), or whether they can progress at their own cadence. 

I leant towards keeping them in sync, as we tend to make orchestrators (and now executors) for a specific version.

But if this is wrong, then I am happy to change it.

LukeButters: We probably need more guidance.

But in general the tentacle client should be working as hard as it can be for you. Which I think should me trying to hide away details best it can, including new different wire command/responses.

source/Octopus.Tentacle.Client/TentacleClient.cs
sburmanoctopus: I don't think I follow this one sorry.

The below methods are the publicly facing versions, which will have their own issues with metrics gathering etc. So I would not be inclined to call them from other methods.

They do call the same executors, so end up calling the same code.

It would be nice to have something show you how they should be called. But that's all done within the orchestrator.

Maybe this one is worth a chat?

LukeButters:
I think in general this is rather good.

source/Octopus.Tentacle.Client/TentacleClient.cs
sburmanoctopus: Yeah, good points 🤔 

The complication is a multi-node environment.

Returning the version in start script would be the most efficient, but it does expose internal concerns to the public 😞 

Like you say, caching does help, but it does mean every server node that does a task (e.g. GetStatus) now has to determine the version again (and log it to the task log 😑 ).

Which then brings us to the idea of reducing what we log when doing this.

@jimmyp , do you have any thoughts on this one?

source/Octopus.Tentacle.Client/TentacleClient.cs
sburmanoctopus: Then we could use `CommandContext` as the parameter to the other commands, to really highlight this is what you need to use these.

source/Octopus.Tentacle.Client/TentacleClient.cs
sburmanoctopus: Yes 🙂

source/Octopus.Tentacle.Client/TentacleClient.cs
LukeButters: nice, in the future it would be nice if the client can tell us if the script is hung (which is after the script has started and it has not logged output for more than some amount of time). Could we stash the information needed to do that in the command context?

sburmanoctopus: Yeah, probably. 

This sounds like a nice separate piece of work to come afterwards. I'll create a card for it.

sburmanoctopus: Done.

Also, [story created](https://app.shortcut.com/octopusdeploy/story/71068/detect-if-tentacle-client-has-hung)

gb-8:
Summary of discussion of PR:

What the PR achieves:
 - It separates orchestration and execution logic inside tentacle client.
 - It adds new methods to allow callers to manage orchestration, leaving client responsible for execution.
 
What the PR does not address:
 - It does not allow the determination of what features the tentacle supports, so does not allow us to detect when we are dealing with a V1 tentacle which does not support idempotent script execution, and generates a new script ticket on the tentacle.

Benefits of getting this PR merged:
 - It does separate orchestration and execution concerns in `TentacleClient` and moves us towards being able to fully orchestrate tentacle interactions (with the exclusion of handling V1 tentacle limitations).
 - It would allow us to stop using `FudgedTentacleClient` and reduce filesystem usage that that is involves.
 
Cost of getting this PR merged:
 - Still requires further review before we can consider merging (It's had some attention from Luke).
 - It does not move us towards our cycle goal, and could distract from that.

Risks management:
   - The PR would need to be re-worked if we want to roll it out via a feature toggle.
   - The changes in the PR should not change behaviour for existing usages of `TentacleClient`, and this behaviour is well covered by automated tests.
   - New methods (for external orchestration) are not covered by automated tests to the same level.
   - Even if we wanted to address all required changes to tentacle client as a contiguous piece of work (including the missing tentacle version management concerns), we would likely want to split it up into smaller PRs, and this PR would be a reasonable first step.

pawelpabich:
A few thoughts:

1. `ITentacleClient` is used by the classic execution pipeline and Dynamic Workers.  Would it make sense to create a different interface for the resilient pipeline so you can freely experiment with it in isolation?
2. The Modern Deployment Team will prepare a mini RFC re how to expose the new K8s feature in the Tentacle Client. I don't think this should affect this PR and it is more of FYI.

source/Octopus.Tentacle.Client/ITentacleClient.cs
```diff
@@ -5,13 +5,17 @@
...
 using Octopus.Tentacle.Client.Scripts;
 using Octopus.Tentacle.Contracts;
-using Octopus.Tentacle.Contracts.ScriptServiceV2;
 using Octopus.Tentacle.Contracts.ScriptServiceV3Alpha;
 
 namespace Octopus.Tentacle.Client
 {
     public interface ITentacleClient : IDisposable
     {
+        Task<CommandResponseV3Alpha> StartScript(StartScriptCommandV3Alpha startScriptCommand, ILog logger, CancellationToken scriptExecutionCancellationToken);
```
APErebus, line 14: Further to @LukeButters point around service types leaking out of tentacle client, The `StartScriptCommandV3Alpha` also really needs to get changed to a non-service type.

In my mini-RFC (as mentioned by Pawel), part of the proposal is to remove service types leaking into the Server codebase.

I'm happy for you to leave it for now, but just be aware that this is something I'm intending to address as part of the proposed solution in the RFC

source/Octopus.Tentacle.Client/TentacleClient.cs
jimmyp: > very much relying on halibut caching to work
I dont understand halibut enough to understand this

> ensure we don't log too much since DetermineScriptServiceVersionToUse might be logging to task log.
I'm wondering if this is something that might need to change about tentacle client. The way I've been thinking about task logging in the new pipeline is that its not something most components should know about. The pipeline will raise events and task logging component somewhere will build up the task log as a read model.

> Returning the version in start script would be the most efficient, but it does expose internal concerns to the public 😞
What exactly is the internal concern being exposed here sorry? In general our multi-node caching mechanism is going to be the `DeploymentExecution` so my first thought is to reach for that if we do need caching. But I'm keen to here more about why we need caching here and what dependancies this would create

jimmyp: Is there anything written down anywhere of the goals and concerns tentacle client is supposed to deal with?

source/Octopus.Tentacle.Client/ITentacleClient.cs
sburmanoctopus: Sounds good. When you're back, maybe we should get together to sync our two streams of work so we get a good path forward 🙂

PR #808, opened 2024-02-15
Added some wizard model tests for Tentacle Manager

raynhamdev:
# Background

This PR adds three basic tests for `DeleteWizardModel` and `ProxyWizardModel` for Tentacle Manager.

More tests will be added later through subsequent PRs.

[sc-69623]

# Pre-requisites


source/Octopus.Manager.Tentacle/Proxy/ProxyWizardModel.cs
```diff
@@ -153,48 +153,51 @@ public IEnumerable<CommandLineInvocation> GenerateScript()
...
+                        .Argument("proxyEnable", true)
+                        .Argument("proxyUsername", ProxyUsername)
+                        .Argument("proxyPassword", ProxyPassword)
+                        .Argument("proxyHost", host)
+                        .Argument("proxyPort", ProxyServerPort)
+                        .Build();
+                    break;
+                }
+                default:
+                    throw new ArgumentOutOfRangeException();
```
acodrington, line 200: This is a change of behaviour.

Previously, if the `ProxyConfigType` was an unsupported value, the enumeration would not return a value at this point. The new beavhiour will throw an exception in the same situation.

This change in behaviour would not be an issue right now, as all values of the enum are supported. However, if a new value is added to the enum, it could throw here instead of silently skipping over it.

raynhamdev: I think whoever adds a new `ProxyConfigType` needs to be aware that they need to implement changes to this method too. So this change of behaviour would ensure they do.

acodrington (approved):
Tests look good, it's nice to finally have them!

PR #811, opened 2024-02-19
Add .NET 6 support to Tentacle Manager projects

raynhamdev:
# Background
This PR enables the projects `Octopus.Tentacle.Manager` & `Octopus.Tentacle.Manager.Tests` to build on .NET 6, as part of the [Tentacle Manager on .NET](https://app.shortcut.com/octopusdeploy/epic/68113/tentacle-manager-on-windows-net-core) project.

This PR also update the Tentacle build project to ensure it builds the aforementioned projects in .NET 6.

Shortcuts: [sc-69659] [sc-69655]
# Results

Changes in this PRs includes:
### Get the projects `Octopus.Tentacle.Manager` & `Octopus.Tentacle.Manager.Tests` to build on .NET6
- Added a new `TargetFramework` of `net6.0-windows` and make `RuntimeIdentifiers` to be `win-x86;win-x64` when the new target framework is select during builds.
- Changes to dependencies/references to ensure they are compatible with .NET6
- Suppress some warnings that are newly introduced in .NET6

### Changes to build configurations
- Added two new build configuration `Release-net6.0-windows-win-x64` and `Release-net6.0-windows-win-x86`. These are added because Nuke build treats `net6.0` and `net6.0-windows` as two different target frameworks.
- The new build configurations only primarily builds  `Octopus.Tentacle.Manager` & `Octopus.Tentacle.Manager.Tests` (including its dependencies: `Octopus.Tentacle` & `Octopus.Tentacle.Contracts`)

### Changes to build project.
- Changes to `BuildWindows` step to ensure the new framework `net6.0-windows` is taken into account.
- Changes to `PackWindowsZips` step to ensure the outputs built for `Octopus.Tentacle.Manager` on `net6.0-windows` is packed.
- Changes to `PackWindowsInstallers` step to ensure msi file is produced for  `Octopus.Tentacle.Manager` built on `net6.0-windows`

# How to review this PR
Quality :heavy_check_mark:
Build the projects `Octopus.Tentacle.Manager` & `Octopus.Tentacle.Manager.Tests` with the new build configurations locally.
[Optional] Locally run the build steps that have been changed.

# Pre-requisites


source/Octopus.Manager.Tentacle/Octopus.Manager.Tentacle.csproj
```diff
@@ -4,25 +4,34 @@
...
     <OutputType>WinExe</OutputType>
     <PublishDir>../../_build/$(AssemblyName)/$(TargetFramework)/$(RuntimeIdentifier)</PublishDir>
-    <RuntimeIdentifiers>win</RuntimeIdentifiers>
-    <TargetFramework>net48</TargetFramework>
     <UseWindowsForms>true</UseWindowsForms>
     <UseWPF>true</UseWPF>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <EnableWindowsTargeting>true</EnableWindowsTargeting>   
+    <TargetFrameworks>net48;net6.0-windows</TargetFrameworks>
+    <SelfContained>false</SelfContained>
```
acodrington, line 16: What happens if `SelfContained == true`? Using false here might work fine as Tentacle .NET 6 is self-contained, but maybe we would be communicating our intent better by setting this to `true`.

raynhamdev: `Self-Contained` was never specified during Tentacle built previously. So it was defaulted to `false`.
When we target `net6.0-windows`, `Self-Contained` is set to `true` by default. This leads to an error saying that self-contained build can't have a dependency on non-self-contained project (Tentacle). 

If we'd like to set this to `true` for Tentacle Manager, Octopus.Tentacle will also need to be self-contained. Are we sure we want this?

raynhamdev: Changing `Octopus.Tentacle` to be self-contained will lead to:
- Its dependencies must also be self-contained. eg. `Octopus.Tentacle.Contracts`
- An increase in tentacle package size.

acodrington: I was under the impression that Tentacle on .NET 6 was already self-contained, as that was a requirement for Philips when the work was initially done.

I might be getting my wires crossed here. Let's discuss further on slack.

build/Build.cs
```diff
@@ -292,7 +312,7 @@ ModifiableFileWithRestoreContentsOnDispose UpdateMsiProductVersion()
         return new ModifiableFileWithRestoreContentsOnDispose(productWxsFilePath);
     }
 
-    void RunBuildFor(string framework, string runtimeId)
+    void RunBuildFor(string framework, string? runtimeId)
```
rhysparry, line 315: Why can `runtimeId` be `null` now? How does that impact the `configuration` string below?

source/Octopus.Manager.Tentacle.Tests/ApplicationStartupSanityCheckFixture.cs
```diff
@@ -47,13 +48,15 @@ public void ApplicationCanStartWithoutCrashing()
                 }
             });
 
+#pragma warning disable CA1416
```
rhysparry, line 51: If I understand [CA1416](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1416) correctly, I'm not sure why we are ignoring this warning. What platform combination are we safeguarding against and if it does build, what do we expect the code to do when the underlying API is unavailable?

(Also applies to the other inclusions below)

source/Octopus.Manager.Tentacle.Tests/Octopus.Manager.Tentacle.Tests.csproj
```diff
@@ -10,9 +10,23 @@
...
+
+  <Choose>
+    <When Condition="'$(TargetFramework)' == 'net48'">
+      <PropertyGroup>
+        <RuntimeIdentifiers>win</RuntimeIdentifiers><!-- This is AnyCPU -->
+      </PropertyGroup>
+    </When>
+    <When Condition="'$(TargetFramework)' == 'net6.0-windows'">
+      <PropertyGroup>
+        <RuntimeIdentifiers>win-x86;win-x64</RuntimeIdentifiers><!-- This is AnyCPU -->
```
rhysparry, line 25: Probably not AnyCPU any more?

source/Octopus.Manager.Tentacle/Infrastructure/ElevationHelper.cs
```diff
@@ -19,12 +18,20 @@ public static void Elevate(IEnumerable<string> args)
             {
                 // We removed this class in favour of using the application manifest for elevation... but then!
                 // See https://github.com/OctopusDeploy/Issues/issues/3875
-                var info = new ProcessStartInfo(Assembly.GetEntryAssembly().Location, String.Join(" ", args));
-                info.Verb = "runas";
+                //var fileName = Assembly.GetEntryAssembly().Location;
```
rhysparry, line 21: Nit: can you clean up this line please?

build/Build.cs
raynhamdev: Nice pickup. This was changed for an experimental approach but I forgot to revert it back.

build/Build.Pack.cs
```diff
@@ -193,22 +193,38 @@ void CreateLinuxPackages(string runtimeId)
...
+                    default:
+                        PackWindowsZip(NetCore, runtimeId);
+                        break;
+                }
+            }
+        });
 
-                var workingDirectory = BuildDirectory / "zip" / framework / runtimeId;
-                var workingTentacleDirectory = workingDirectory / "tentacle";
+    void PackWindowsZip(string framework, string runtimeId)
```
LukeButters, line 213: Is this specific to windows? could it just be called packzip? @acodrington

build/Build.Pack.cs
```diff
@@ -193,22 +193,38 @@ void CreateLinuxPackages(string runtimeId)
...
+                    case "win":
+                        PackWindowsZip(NetFramework, runtimeId);
+                        break;
+                    case "win-x86":
+                    case "win-x64":
+                        PackWindowsZip(NetCoreWindows, runtimeId);
+                        PackWindowsZip(NetCore, runtimeId);
+                        break;
+                    default:
+                        PackWindowsZip(NetCore, runtimeId);
```
LukeButters, line 207: when would we go down this path? @acodrington

build/Build.cs
```diff
@@ -152,8 +153,25 @@ public Build()
                 using var versionInfoFile = ModifyTemplatedVersionAndProductFilesWithValues();
                 using var productWxsFile = UpdateMsiProductVersion();
 
-                RuntimeIds.Where(x => x.StartsWith("win"))
-                    .ForEach(runtimeId => RunBuildFor(runtimeId.Equals("win") ? NetFramework : NetCore, runtimeId));
+                var runtimeIds = RuntimeIds.Where(x => x.StartsWith("win"));
+                
+                foreach (var runtimeId in runtimeIds)
```
LukeButters, line 158: this looks like duplicate of the other switch the looks the same, should they be in sync? if so maybe a refactor this out?

source/Octopus.Tentacle/Octopus.Tentacle.csproj
```diff
@@ -10,7 +10,7 @@
 		<OutputType>Exe</OutputType>
 		<PublishDir>../../_build/$(AssemblyName)/$(TargetFramework)/$(RuntimeIdentifier)</PublishDir>
 		<RootNamespace>Octopus.Tentacle</RootNamespace>
-		<TargetFrameworks>net48;net6.0</TargetFrameworks>
+		<TargetFrameworks>net48;net6.0;net6.0-windows</TargetFrameworks>
```
LukeButters, line 13: we did we need to do this? tentacle is x-platform but doesn't need it

LukeButters: we may need to come back and simplify the build

LukeButters: @acodrington

build/Build.Pack.cs
rhysparry: win-arm64?

build/Build.Pack.cs
```diff
@@ -193,22 +193,38 @@ void CreateLinuxPackages(string runtimeId)
...
-                var framework = runtimeId.Equals("win") ? NetFramework : NetCore;
+                switch (runtimeId)
+                {
+                    case "win":
+                        PackWindowsZip(NetFramework, runtimeId);
+                        break;
+                    case "win-x86":
+                    case "win-x64":
+                        PackWindowsZip(NetCoreWindows, runtimeId);
+                        PackWindowsZip(NetCore, runtimeId);
```
LukeButters, line 204: On the surface it is a little confusing, do we end up packing the net core for windows stuff twice? But we don't below for arm?

I wonder why we do that?

PR #818, opened 2024-02-21
Monitor all jobs status via long-running monitor

APErebus:
Note: There are changes here that have been abstracted out into 2 other PRs #821  and #822. I'll update the PR when they are approved and merged

# Background

In preparation for making Tentacle when running in Kubernetes a much more stateless application, we need to change how we monitor the status of jobs.

Currently, when a script is started, it spawns a `RunningKubernetesJob` which runs it's `Execute` in a background thread. In this `Execute`, it creates the job and starts monitoring the logs & job status. This means that if the Tentacle is torn-down or evicted etc, then we lose all that state.

This PR starts this improvement by introducing a long-running `KubernetesJobMonitorTask` which is used to watch changes to _all_ jobs in the Tentacle namespace. This is then used in the `RunningKubenetesJob` to monitor the status. Eventually, this would be done directly in the `GetScriptStatus` RPC call, but that will come in later PR's.

# Results

Introduces a the new `KubernetesJobMonitorTask` and `KubernetesJobMonitor` class which is started in the `RunAgentCommand` (much like the `WorkspaceCleanerTask`).

The `KubernetesJobMonitor` first loads the current state of all the jobs and stores them in a local dictionary. It then starts a watch from the `resourceVersion` returned from the initial list call. This is considered best practice and is recommended in the Kubernetes documentation under [Efficient detection of changes](https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes).

We then access the job monitor (via a different interface) in the `RunningKubernetesJob` and use the resolved status to determine if it's finished.

I have done testing and verified that this correctly monitors successful and failed jobs and deployments are completed successfully in Octopus Server.

This also increases the default Job TTL to 30min

**Note:** This will require an update to the Helm Chart as the Tentacle service account will need read access to `pods` to retrieve the exit code in case of failure

Shortcut story: [sc-71688]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle/Kubernetes/KubernetesJobMonitor.cs
+++ b/source/Octopus.Tentacle/Kubernetes/KubernetesJobMonitor.cs
@@ -37,36 +37,43 @@ namespace Octopus.Tentacle.Kubernetes
 
         async Task IKubernetesJobMonitor.StartAsync(CancellationToken cancellationToken)
         {
-            //initially load all the jobs and their status's
-            await InitialLoadAsync(cancellationToken);
-
             await jobService.WatchAllJobsAsync(async (type, job) =>
                 {
-                    await Task.CompletedTask;
-
-                    var scriptTicket = job.GetScriptTicket();
-
-                    switch (type)
+                    try
                     {
-                        case WatchEventType.Added or WatchEventType.Modified:
+                        log.Verbose($"Received {type} event for job {job.Name()}");
+
+                        var scriptTicket = job.GetScriptTicket();
+
+                        switch (type)
                         {
-                            if (!jobStatusLookup.TryGetValue(scriptTicket, out var status))
+                            case WatchEventType.Added or WatchEventType.Modified:
                             {
-                                status = new JobStatus(job.GetScriptTicket());
-                                jobStatusLookup[scriptTicket] = status;
+                                if (!jobStatusLookup.TryGetValue(scriptTicket, out var status))
+                                {
+                                    status = new JobStatus(job.GetScriptTicket());
+                                    jobStatusLookup[scriptTicket] = status;
+                                }
+
+                                await status.UpdateAsync(job, podService, cancellationToken);
+                                log.Verbose($"Updated job {job.Name()} status. {status}");
+
+                                break;
                             }
+                            case WatchEventType.Deleted:
+                                log.Verbose($"Removed {type} job {job.Name()} status");
 
-                            await status.UpdateAsync(job, podService, cancellationToken);
-
-                            break;
+                                //if the job is deleted, remove it
+                                jobStatusLookup.Remove(scriptTicket);
+                                break;
+                            default:
+                                log.Warn($"Received watch event type {type} for job {job.Name()}. Ignoring as we don't need it");
+                                break;
                         }
-                        case WatchEventType.Deleted:
-                            //if the job is deleted, remove it
-                            jobStatusLookup.Remove(scriptTicket);
-                            break;
-                        default:
-                            log.Warn($"Received watch event type {type} for job {job.Name()}. Ignoring");
-                            break;
+                    }
+                    catch (Exception e)
+                    {
+                        log.Error(e, $"Failed to process event {type} for job {job.Name()}.");
                     }
                 }, ex =>
                 {
@@ -77,18 +84,6 @@ namespace Octopus.Tentacle.Kubernetes
 
         JobStatus? IKubernetesJobStatusProvider.TryGetJobStatus(ScriptTicket scriptTicket)
             => jobStatusLookup.TryGetValue(scriptTicket, out var status) ? status : null;
-
-        async Task InitialLoadAsync(CancellationToken cancellationToken)
-        {
-            var allJobs = await jobService.ListAllJobsAsync(cancellationToken);
-            foreach (var job in allJobs.Items)
-            {
-                var status = new JobStatus(job.GetScriptTicket());
-                await status.UpdateAsync(job, podService, cancellationToken);
-
-                jobStatusLookup[status.ScriptTicket] = status;
-            }
-        }
     }
 
     public class JobStatus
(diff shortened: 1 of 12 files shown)


source/Octopus.Tentacle/Kubernetes/KubernetesJobMonitor.cs
```diff
@@ -0,0 +1,165 @@
...
+
+        async Task IKubernetesJobMonitor.StartAsync(CancellationToken cancellationToken)
+        {
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                //initially load all the jobs and their status's
+                var initialResourceVersion = await InitialLoadAsync(cancellationToken);
+
+                //we start the watch from the resource version we initially loaded. This means we receive w
+                await jobService.WatchAllJobsAsync(initialResourceVersion,async (type, job) =>
```
APErebus, line 46: As mentioned, this is how to safely retrieve the status. You do an initial load and get the `resourceVersion` for when the list was loaded, you then start the `watch` from that point, so if there are any changes in-between the list being loaded and the watch starting, they will be immediately retrieved in the watch.

source/Octopus.Tentacle/Kubernetes/KubernetesJobMonitor.cs
```diff
@@ -0,0 +1,165 @@
...
+        public KubernetesJobMonitor(IKubernetesJobService jobService, IKubernetesPodService podService, ISystemLog log)
+        {
+            this.jobService = jobService;
+            this.podService = podService;
+            this.log = log;
+        }
+
+        async Task IKubernetesJobMonitor.StartAsync(CancellationToken cancellationToken)
+        {
+            while (!cancellationToken.IsCancellationRequested)
```
APErebus, line 40: If the watch has an error (which can happen), this restarts the watch from scatch

source/Octopus.Tentacle/Kubernetes/KubernetesJobService.cs
```diff
@@ -63,6 +66,39 @@ await foreach (var (type, job) in response.WatchAsync<V1Job, V1JobList>(onError,
...
+            var watchErrorCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+
+            Action<Exception> internalOnError = ex =>
+            {
+                //We cancel the watch explicitly (so it can be restarted)
+                watchErrorCancellationTokenSource.Cancel();
+
+                //notify there was an error
+                onError(ex);
+            };
```
APErebus, line 94: My read of the source code for `WatchAsync` is that it won't kill the watch onError, so this allows us to cancel the `WatchAsync` if there is an error, so we can then restart it.

source/Octopus.Tentacle/Kubernetes/OctopusLabels.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Octopus.Tentacle.Kubernetes
+{
+    public static class OctopusLabels
+    {
+        public const string ScriptTicketId = "octopus.com/scriptTicketId";
```
APErebus, line 5: This is how we only look for jobs we are have created.

source/Octopus.Tentacle/Kubernetes/Scripts/RunningKubernetesJob.cs
```diff
@@ -75,6 +78,8 @@ public class RunningKubernetesJob : IRunningScript
 
             outputStreamWriter = new KubernetesJobOutputStreamWriter(workspace);
 
+            State = ProcessState.Pending;
```
APErebus, line 81: We weren't setting the initial state to be `Pending`, so it isn't immediately completed... which was wrong

source/Octopus.Tentacle/Kubernetes/Scripts/RunningKubernetesJob.cs
```diff
@@ -182,26 +187,24 @@ async Task<int> MonitorJobAndLogs(IScriptLogWriter writer)
         async Task<int> CheckIfJobHasCompleted(CancellationTokenSource jobCompletionCancellationTokenSource)
         {
             var resultStatusCode = 0;
-            await jobService.Watch(scriptTicket, job =>
+            JobStatus? status = null;
+            while (!scriptCancellationToken.IsCancellationRequested)
             {
-                var firstCondition = job.Status?.Conditions?.FirstOrDefault();
-                switch (firstCondition)
+                status = jobStatusProvider.TryGetJobStatus(scriptTicket);
```
APErebus, line 193: We now get the status from the status provider (the monitor)

source/Octopus.Tentacle/Kubernetes/KubernetesJobMonitor.cs
```diff
@@ -0,0 +1,165 @@
...
+    {
+        Task StartAsync(CancellationToken token);
+    }
+
+    public interface IKubernetesJobStatusProvider
+    {
+        JobStatus? TryGetJobStatus(ScriptTicket scriptTicket);
+    }
+
+    public class KubernetesJobMonitor : IKubernetesJobMonitor, IKubernetesJobStatusProvider
```
APErebus, line 24: I have 2 separate interfaces so that one is for starting the monitor, and the other is for retrieving the status. It means you   are getting an interface that represents the action you want, not the underlying class

source/Octopus.Tentacle/Kubernetes/KubernetesJobMonitorTask.cs
```diff
@@ -0,0 +1,77 @@
...
+
+namespace Octopus.Tentacle.Kubernetes
+{
+    public interface IKubernetesJobMonitorTask
+    {
+        void Start();
+        void Stop();
+    }
+
+    public class KubernetesJobMonitorTask: IKubernetesJobMonitorTask, IDisposable
```
APErebus, line 14: This structure/pattern is almost a copy of the `WorkspaceCleanerTask` in how it spawns a background thread and manages the lifecycle and cancellation.

APErebus:
no longer required as we decided to merge #823 and redo this PR afterwards

PR #815, opened 2024-02-20
Fix pod logs (and dynamically updating the log level in general)

scme0:
# Background

My previous attempt at add pod logs doesn't seem to work properly. I'm not sure how that slipped by but anyways, this seems to fix it.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

- [ ] I have read [How we use GitHub Issues](https://github.com/OctopusDeploy/Issues/blob/main/docs/CONTRIBUTING.internal.md) for help deciding when and where it's appropriate to make an issue.
- [ ] I have considered informing or consulting the right people, according to the [ownership map](https://whimsical.com/ownership-map-NzbiD4HJyvhC9jNJNfS6TG).
- [ ] I have considered appropriate testing for my change.

PR #821, opened 2024-02-22
Format kubernetes console logs with log level and timestamp

APErebus:
# Background

We are correctly logging system logs to the Kubernetes pod logs, however they don't have some useful information like timestamp or log level, making it hard to correlate with other logs for timings etc.

# Results

This PR changes the target layout for the `stdout` and `stderr` targets, but only when running in Kubernetes.

## After

image

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle/Startup/OctopusProgram.cs
+++ b/source/Octopus.Tentacle/Startup/OctopusProgram.cs
@@ -11,6 +11,7 @@ using Autofac;
 using Autofac.Core;
 using NLog;
 using NLog.Config;
+using NLog.Layouts;
 using NLog.Targets;
 using Octopus.Diagnostics;
 using Octopus.Tentacle.Configuration;
@@ -311,6 +312,14 @@ namespace Octopus.Tentacle.Startup
             LogManager.ThrowConfigExceptions = true;
             LogManager.Configuration = new XmlLoggingConfiguration(nLogFile);
 #endif
+            //if we are running in kubernetes, make the console log more helpful
+            if (PlatformDetection.Kubernetes.IsRunningInKubernetes)
+            {
+                var consoleLayout = new SimpleLayout("${longdate}  ${uppercase:${level}:padding=5}  ${message}${onexception:${newline}${exception:format=ToString}}");
+                LogManager.Configuration.FindTargetByName<ColoredConsoleTarget>("stdout").Layout = consoleLayout;
+                LogManager.Configuration.FindTargetByName<ColoredConsoleTarget>("stderr").Layout = consoleLayout;
+            }
+
             SystemLog.Appenders.Add(new NLogAppender());
 
             if (Environment.GetEnvironmentVariable("OCTOPUS__TENTACLE__LOGLEVEL") is { } logLevel &&
@@ -384,7 +393,7 @@ namespace Octopus.Tentacle.Startup
             }
             if (!string.IsNullOrWhiteSpace(configFile))
                 return new StartUpConfigFileInstanceRequest(configFile);
-            
+
             return new StartUpDynamicInstanceRequest();
         }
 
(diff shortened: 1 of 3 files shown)


source/Octopus.Tentacle/Startup/OctopusProgram.cs
```diff
@@ -311,6 +312,14 @@ void InitializeLogging()
             LogManager.ThrowConfigExceptions = true;
             LogManager.Configuration = new XmlLoggingConfiguration(nLogFile);
 #endif
+            //if we are running in kubernetes, make the console log more helpful
+            if (PlatformDetection.Kubernetes.IsRunningInKubernetes)
+            {
+                var consoleLayout = new SimpleLayout("${longdate}  ${uppercase:${level}:padding=5}  ${message}${onexception:${newline}${exception:format=ToString}}");
```
flin-8, line 318: Do we need the date if Lens can add the timestamp for us anyway?

APErebus: Not necessarily, however the lens timestamps are slightly lower resolution (no milliseconds).

Also, not everyone uses Lens for viewing logs (in this team they do, but I know Cail uses K9s for his k8s stuff)

flin-8: Actually ... is this similar to @scme0 's change? https://github.com/OctopusDeploy/OctopusTentacle/pull/815/files/3f6991192d857afe7d0526aa5fe6ae09e986387a

APErebus: well he reverted it in his PR and I put it back it 😆 

I would argue there is some benefit in having it written in the log message and not requiring a 3rd party tool to add it

scme0: Yep it is 😅 but happy with it either way. This code updates it only for Kubernetes Tentacle. Not sure if it's worth just updating the nlog config so it's the same for all Tentacles? Might be a bit cleaner.

APErebus: @scme0 I didn't want to change logging for non-k8s tentacles because I don't know how people use (or not use) those logs. At least this is segregated to just the k8s

flin-8 (approved):
I think this change might be better done in the nlog file itself, but you can make the call :)

source/Octopus.Tentacle/Startup/OctopusProgram.cs
scme0: I would suggest attempting to use a variables like I did with log level as it's a bit clearer what's happening maybe?

APErebus: Taking with @flin-8, I'm going to have a different nlog file completely for k8s

APErebus: @scme0 @flin-8 I've changed this to use a custom nlog file for Kubernetes with the different log layout in 63aa6e0a

source/Octopus.Tentacle/Tentacle.exe.k8s.nlog
```diff
@@ -0,0 +1,41 @@
...
+  throwExceptions="false"
+  internalLogLevel="Fatal"
+  internalLogToConsole="true">
+
+  <extensions>
+    <add assembly="Tentacle"/>
+  </extensions>
+
+  <variable name="appName" value="Octopus Tentacle" />
+  <variable name="messageLayout" value="${longdate} | ${uppercase:${level}:padding=5} | ${message}${onexception:${newline}${exception:format=ToString}}"/>
```
APErebus, line 15: :pipe:

scme0 (approved):

APErebus:
Updated the pic and verified that it works with the custom file

PR #822, opened 2024-02-22
Fix building debug Kubernetes image

APErebus:
# Background

Debugging Tentacle running in a Kubernetes cluster is challenging as you can't just hook up the debugger and go. 

To do this via Rider, we have a custom docker image that includes the Rider Remote Debugging Tools. This PR fixes the build since it was moved to `buildx` and the multi-arch images.

# Results

One key point is that this can only build a single architecture image as the rider tools are only single arch. To do this, I added a new Nuke parameter called `DockerPlatform` which allows for changing the default multi-arch platforms. Also, I added a `DockerBuilder` parameter so you can specify a custom docker build if you haven't swapped to buildx for all builds.

There is a new Nuke target to build and load the debug image. It's essentially the same, just sources a different docker file

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/build/Build.Pack.cs
+++ b/build/Build.Pack.cs
@@ -23,8 +23,13 @@ partial class Build
     //We don't sign linux packages when building locally
     readonly bool SignLinuxPackages = !IsLocalBuild;
 
-    [Parameter(Name = "DockerBuilder")] string? DockerBuildxBuilder;
-    [Parameter(Name = "DockerPlatform")] string DockerPlatform = "linux/arm64,linux/amd64";
+    [Parameter("Used to set a custom docker builder when executing DockerBuildxBuild tasks",
+        Name = "DockerBuilder")]
+    string? DockerBuildxBuilder;
+
+    [Parameter("Specifies the platforms to build the docker images in. Multiple platforms must be comma-separated. Defaults to 'linux/arm64,linux/amd64'.",
+        Name = "DockerPlatform")]
+    string DockerPlatform = "linux/arm64,linux/amd64";
 
     [PublicAPI]
     Target PackOsxTarballs => _ => _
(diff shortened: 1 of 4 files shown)


docker/kubernetes-tentacle/dev/Dockerfile
```diff
@@ -1,14 +1,81 @@
-ARG IMAGE_TAG
-FROM docker.packages.octopushq.com/octopusdeploy/kubernetes-tentacle:$IMAGE_TAG
```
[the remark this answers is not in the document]
scme0: Damn that's a shame

build/Build.Pack.cs
```diff
@@ -23,6 +23,9 @@ partial class Build
     //We don't sign linux packages when building locally
     readonly bool SignLinuxPackages = !IsLocalBuild;
 
+    [Parameter(Name = "DockerBuilder")] string? DockerBuildxBuilder;
+    [Parameter(Name = "DockerPlatform")] string DockerPlatform = "linux/arm64,linux/amd64";
```
scme0, line 27: Is it worth adding some descriptions for these? If you do `./build.[sh/ps1] --help` you get a list of parameters and the description could help someone with what value they need without looking at the code. (I only found this out yesterday 😅 )

Screenshot 2024-02-22 at 13 50 11

scme0 (approved):
I've pulled your branch and it builds for me! 🚀

build/Build.Pack.cs
APErebus: ok yeah, I didn't know that was a thing. Will do

APErebus: Updated in ba48c6b9

scme0 (approved):
Nice!

PR #824, opened 2024-02-23
Monitor pod status via long-running watch task

APErebus:
# Background

Superceeds #818 

In preparation for making Tentacle when running in Kubernetes a much more stateless application, we need to change how we monitor the status of jobs.

Currently, when a script is started, it spawns a `RunningKubernetesPod` which runs it's `Execute` in a background thread. In this `Execute`, it creates the pod and starts monitoring the logs & pod status. This means that if the Tentacle is torn-down or evicted etc, then we lose all that state.

This PR starts this improvement by introducing a long-running `KubernetesPodMonitorTask` which is used to watch changes to _all_ pods in the Tentacle namespace. This is then used in the `RunningKubenetesPod` to get the status. Eventually, this would be done directly in the `GetScriptStatus` RPC call, but that will come in later PR's.

# Results

Introduces a the new `KubernetesPodMonitorTask` and `KubernetesPodMonitor` class which is started in the `RunAgentCommand` (much like the `WorkspaceCleanerTask`).

The `KubernetesPodMonitor` first loads the current state of all the pods and stores them in a local dictionary. It then starts a watch from the `resourceVersion` returned from the initial list call. This is considered best practice and is recommended in the Kubernetes documentation under [Efficient detection of changes](https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes).

We then access the pod monitor (via a different interface) in the `RunningKubernetesPod` and use the resolved status to determine if it's finished.

I have done testing and verified that this correctly monitors successful and failed pods and deployments are completed successfully in Octopus Server.

Shortcut story: [sc-71688]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle/Kubernetes/KubernetesPodMonitor.cs
+++ b/source/Octopus.Tentacle/Kubernetes/KubernetesPodMonitor.cs
@@ -41,47 +41,7 @@ namespace Octopus.Tentacle.Kubernetes
 
                 // We start the watch from the resource version we initially loaded.
                 // This means we only receive events that occur after the resource version
-                await podService.WatchAllPods(initialResourceVersion, async (type, pod) =>
-                    {
-                        await Task.CompletedTask;
-
-                        try
-                        {
-                            log.Verbose($"Received {type} event for pod {pod.Name()}");
-
-                            var scriptTicket = pod.GetScriptTicket();
-
-                            switch (type)
-                            {
-                                case WatchEventType.Added or WatchEventType.Modified:
-                                {
-                                    if (!podStatusLookup.TryGetValue(scriptTicket, out var status))
-                                    {
-                                        status = new PodStatus(pod.GetScriptTicket());
-                                        podStatusLookup[scriptTicket] = status;
-                                    }
-
-                                    status.Update(pod);
-                                    log.Verbose($"Updated pod {pod.Name()} status. {status}");
-
-                                    break;
-                                }
-                                case WatchEventType.Deleted:
-                                    log.Verbose($"Removed {type} pod {pod.Name()} status");
-
-                                    //if the pod is deleted, remove it
-                                    podStatusLookup.Remove(scriptTicket);
-                                    break;
-                                default:
-                                    log.Warn($"Received watch event type {type} for pod {pod.Name()}. Ignoring as we don't need it");
-                                    break;
-                            }
-                        }
-                        catch (Exception e)
-                        {
-                            log.Error(e, $"Failed to process event {type} for pod {pod.Name()}.");
-                        }
-                    }, ex =>
+                await podService.WatchAllPods(initialResourceVersion, OnNewEvent, ex =>
                     {
                         log.Error(ex, "An unhandled error occured in monitoring the pods");
                     }, cancellationToken
@@ -89,7 +49,7 @@ namespace Octopus.Tentacle.Kubernetes
             }
         }
 
-        async Task<string> InitialLoadAsync(CancellationToken cancellationToken)
+        internal async Task<string> InitialLoadAsync(CancellationToken cancellationToken)
         {
             log.Verbose("Preloading pod statuses");
             //clear the status'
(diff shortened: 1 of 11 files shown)


source/Octopus.Tentacle/Kubernetes/KubernetesPodMonitorTask.cs
```diff
@@ -0,0 +1,77 @@
...
+        void Stop();
+    }
+
+    public class KubernetesPodMonitorTask: IKubernetesPodMonitorTask, IDisposable
+    {
+        readonly IKubernetesPodMonitor podMonitor;
+        readonly ISystemLog log;
+        readonly CancellationTokenSource cancellationTokenSource = new ();
+
+        static readonly object LockObj = new();
```
scme0, line 20: nit: this probably shouldn't be static, it's there to protect instance properties so it should also be an instance property. If you had two instances of `KubernetesPodMonitorTask` they would share the same lock. I don't think this is really an issue though as it's registered as a singleton, thus a "nit" 😅.

scme0 (approved):
I've left a bunch of comments but they are mainly nits and questions so I'll pre-approve. It's looking really good! Nice work!

source/Octopus.Tentacle/Kubernetes/KubernetesPodMonitorTask.cs
APErebus: fair call, I'll change it :)

source/Octopus.Tentacle/Kubernetes/KubernetesPodMonitorTask.cs
```diff
@@ -0,0 +1,77 @@
...
+        public void Stop()
+        {
+            lock (LockObj)
+            {
+                if (monitorTask is null) return;
+
+                try
+                {
+                    log.Info("Stopping Kubernetes Pod Monitor");
+                    cancellationTokenSource.Cancel();
```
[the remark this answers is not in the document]
APErebus: The cancellation token is more for signalling graceful shutdown. The shutdown is only called when the Tentacle is shutting down itself, so it'll be killed

source/Octopus.Tentacle/Kubernetes/KubernetesPodMonitorTask.cs
```diff
@@ -0,0 +1,77 @@
...
+
+                    monitorTask.Wait();
```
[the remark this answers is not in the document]
APErebus: 🤔 Probably not a bad idea 👍

source/Octopus.Tentacle/Kubernetes/KubernetesPodMonitor.cs
```diff
@@ -0,0 +1,165 @@
...
+            this.podService = podService;
+            this.log = log;
+        }
+
+        async Task IKubernetesPodMonitor.StartAsync(CancellationToken cancellationToken)
+        {
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                //initially load all the pods and their status's
+                var initialResourceVersion = await InitialLoadAsync(cancellationToken);
```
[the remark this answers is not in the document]
APErebus: From the previous PR #818 

> As mentioned, this is how to safely retrieve the status. You do an initial load and get the resourceVersion for when the list was loaded, you then start the watch from that point, so if there are any changes in-between the list being loaded and the watch starting, they will be immediately retrieved in the watch.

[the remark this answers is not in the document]
scme0: Answered in engineering sync: The "initialResourceVersion" from this call is passed to the watch call which means all updates from that version are given to the watch call guaranteeing that nothing is missed.

source/Octopus.Tentacle/Kubernetes/KubernetesPodMonitorTask.cs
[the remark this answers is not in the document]
APErebus: Added a 30s timeout to allow for graceful shutdown

source/Octopus.Tentacle/Kubernetes/KubernetesPodMonitor.cs
```diff
@@ -0,0 +1,166 @@
...
+                    {
+                        await Task.CompletedTask;
+
+                        try
+                        {
+                            log.Verbose($"Received {type} event for pod {pod.Name()}");
+
+                            var scriptTicket = pod.GetScriptTicket();
+
+                            switch (type)
```
flin-8, line 54: Can we extract this logic and unit test?

flin-8 (approved):
Code looks nice! Let's add some unit tests around the status updates so we add confidence and reduce regresssions in the future

source/Octopus.Tentacle/Kubernetes/KubernetesPodMonitor.cs
APErebus: yep

APErebus: @flin-8 Added an initial set of tests in 03baaa4

flin-8: Good start 👍 

I usually prefer to extract out classes instead of making methods internal/public just for tests (we could potentially separate out a class for the state transition from the watching) - but don't let that stop you from merging

APErebus: ok, I'll keep that in mind. This area is getting a bit of rework with the logs monitoring, so I'll try and clean it up there 👍

PR #829, opened 2024-03-05
Can we default to pwsh?

matt-richardson:
# Background

`PowerShell.exe` is very slow. 

# Results

Can we speed it up by using `pwsh` instead?

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle/Scripts/PowerShellCore.cs
+++ b/source/Octopus.Tentacle/Scripts/PowerShellCore.cs
@@ -1,0 +1,80 @@
+using System;
+using System.IO;
+using System.Text;
+using Octopus.Tentacle.Util;
+
+namespace Octopus.Tentacle.Scripts
+{
+    public class PowerShellCore : IShell
+    {
+        const string EnvPowerShellPath = "pwsh";
+        static string? powerShellPath;
+
+        public string Name => nameof(PowerShell);
+
+        public string GetFullPath()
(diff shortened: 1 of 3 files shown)


source/Octopus.Tentacle/Scripts/PowerShell.cs
```diff
@@ -1,59 +1,24 @@
...
-                    powerShellPath = EnvPowerShellPath;
-            }
-            catch (Exception)
-            {
-                powerShellPath = EnvPowerShellPath;
-            }
-
-            return powerShellPath;
+            var powerShellCore = new PowerShellCore();
+            if (powerShellCore.PowerShellCoreExists)
```
rhysparry, line 13: If we were to do this, we would probably need some sort of configuration to disable/enable.

Also, is it sufficient to find PowerShell core? In the change for the hardening script the branching logic seems to only switch for PowerShell core >= 7.3. Do we need to know the version of PowerShell core?

matt-richardson: The 7.3 check is to work around a change they made to not load the security stuff - https://github.com/PowerShell/PowerShell/issues/18896

Ironically, it was designed to speed up powershell :)

matt-richardson:
The world has moved on a lot since this was done.

PR #830, opened 2024-03-07
Add telemetry to Tentacle Manager

acodrington:
# Background
We currently have no visibility into our customer’s use of Tentacle Manager. This makes it difficult to decide how to prioritize potential future work, as there is no evidence on which to base these decisions.

This PR makes the first step to improve this situation by introducing telemetry to Tentacle Manager.

# Results
Tentacle Manager will now send telemetry to Amplitude (via Octopus Server) when:

_An instance of a polling Tentacle (deployment target, or worker) is installed/registered/configured._

The telemetry data is very basic and only contains:
* User ID
* Device ID (unique to the instance of Tentacle Manager)
* Communication Style (i.e. polling/listening)
* Machine Type (Worker, or Deployment Target)

Octopus Server will augment the event with some further Server-specific data before sending to Amplitude. This data is standard to every event sent to Amplitude by Octopus Server and includes:
* Installation ID
* License Hash
* Feature toggles
* Feature settings
* Host environment (i.e. on-prem, cloud)
* Version

The new 'Tentacle Manager script executed' event in Amplitude

## Why only send telemetry for installing polling tentacles?
We are constrained by the requirement that Tentacle Manager _only_ communicate with Octopus Server, and no other remote endpoint.

Many (most?) Tentacles are installed within restricted corporate networks, therefore requiring additional outgoing requests from Tentacle Manager would often involve updating firewall policies etc, which is not feasible for this scenario.

There is an existing mechanism, used by the front-end portal, to send Amplitude messages via Octopus Server. While this is adequate for our needs, it does mean that we can only send telemetry when we have _a valid Octopus Server API endpoint_. Tentacle Manager is not inherently associated with an instance of Octopus Server, and only communicates with an instance of Server during the ‘Install Polling Tentacle’ wizard process.

Our metrics show that ~30% of Tentacles are polling, which is a large enough sample size to get some valuable data on Tentacle Manager usage.

## What is the Tentacle Manager Instance ID?
The ability to differentiate between instances of Tentacle Manager could be a useful data point when analyzing the telemetry events.

For example, we see from the telemetry that a user uses Tentacle Manager to configure 3 polling Tentacle instances over 1 week. It may be useful to know if that was 3 polling Tentacles installed on 3 different machines, or if they installed (and uninstalled) a Tentacle 3 times on a single machine.

To provide this data, a very simple instance identifier was implemented for Tentacle Manager. When Tentacle Manager is started, it looks for a specific file in the executable’s current directory called `TentacleManagerInstanceID`. This file only contains a unique identifier (i.e. GUID) which Tentacle Manager can use to populate the ‘DeviceID’ field in Amplitude messages. If this file does not exist, it is automatically generated.

This allows us to differentiate between instances of Tentacle Manager, without relying on any system-specific or identifiable data (e.g. Machine name, IP address, Windows Registry ID value).

# Potential future work
This PR represents the minimum amount of work required to produce useful telemetry data for Tentacle Manager, but there is plenty that has been intentionally not done in the interest of time and scope:

## Telemetry for more events
This PR is the simplest implementation available, but an implementation involving changes to both Tentacle and Server would allow more telemetry events (e.g. Deleting, Modifying, Starting/Stopping services) for both polling and listening Tentacles.

## More telemetry tests
We have a single ‘happy path’ test for sending telemetry. This can be expanded upon in the future as necessary.

## Further refactoring to separate UI and business logic
Working on this PR exposed some more rough edges on how we separate our UI and business logic in Tentacle Manager. This can be improved in the future to allow more comprehensive unit tests (rather than relying on manual testing).

# How to review this PR
:heavy_check_mark: General code quality, and manual testing of the ‘install new Tentacle’ wizard.

# Pre-requisites

### DIFF
--- a/source/Octopus.Manager.Tentacle.Tests/Utils/TentacleManagerInstanceIdentifierServiceFixture.cs
+++ b/source/Octopus.Manager.Tentacle.Tests/Utils/TentacleManagerInstanceIdentifierServiceFixture.cs
@@ -2,6 +2,7 @@
 using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
+using Halibut.Util;
 using NUnit.Framework;
 using Octopus.Manager.Tentacle.Util;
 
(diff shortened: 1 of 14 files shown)


source/Octopus.Manager.Tentacle.Tests/WizardFixtures/SetupTentacleWizardFixture.cs
```diff
@@ -69,17 +69,32 @@ public void WhenSettingUpPollingTentacle_ScriptShouldBeCorrectlyGenerated()
...
+            // TODO: Replace with a less ugly assertion
+            await telemetryService
+                .Received(1)
+                .SendTelemetryEvent(
+                    new Uri("http://localhost"),
+                    Arg.Is<TelemetryEvent>(t =>
+                        t.EventProperties.ContainsKey("Communication Style") &&
+                        t.EventProperties["Communication Style"] == CommunicationStyle.TentacleActive.ToString() &&
+                        t.EventProperties.ContainsKey("Machine Type") &&
+                        t.EventProperties["Machine Type"] == MachineType.DeploymentTarget.ToString()));
```
acodrington, line 97: I am aware this assert is ugly as sin 😅 

We can improve this by further refactoring to avoid most (if not all) of the mocks. This is some minor tech debt that I feel is worth accepting in the interest of completing the Tentacle Manager project in a timely fashion, as it will be very easy to modify in the future.

source/Octopus.Manager.Tentacle.Tests/Utils/TentacleManagerInstanceIdentifierServiceFixture.cs
```diff
@@ -0,0 +1,82 @@
...
+    {
+        readonly string identifierLocation = Path.GetTempPath();
+        
+        [Test]
+        public async Task Given_ThereIsNoExistingIdentifier_When_AnIdentifierIsRequested_Then_ANewIdentifierIsGeneratedAndReturned()
+        {
+            // Arrange: Ensure there is no current identifier
+            var identifierFullPath = Path.Combine(identifierLocation, TentacleManagerInstanceIdentifierService.IdentifierFileName);
+            File.Delete(identifierFullPath);
+            Assert.False(File.Exists(identifierFullPath));
```
rhysparry, line 20: Does this potentially introduce flakiness if tests are run in parallel? Can we use a temporary directory available for just this test instead?

source/Octopus.Manager.Tentacle.Tests/Utils/TentacleManagerInstanceIdentifierServiceFixture.cs
```diff
@@ -0,0 +1,82 @@
...
+
+            // Act: Get the identifier
+            var sut = new TentacleManagerInstanceIdentifierService(new DirectoryInfo(identifierLocation));
+            var identifierValue = await sut.GetIdentifier();
+            
+            // Assert: The identifier returned should match the contents of the file we expect to be created
+            Assert.True(File.Exists(identifierFullPath));
+            var identifierFileContents = File.ReadAllLines(identifierFullPath);
+            Assert.AreEqual(1, identifierFileContents.Length);
+            Assert.AreEqual(identifierValue, identifierFileContents.Single());
```
rhysparry, line 30: Should we check that `identifierValue` is not empty (i.e. has some characters)?

source/Octopus.Manager.Tentacle.Tests/Utils/TentacleManagerInstanceIdentifierServiceFixture.cs
```diff
@@ -0,0 +1,82 @@
...
+            }
+            
+            // Act: Get the identifier
+            var sut = new TentacleManagerInstanceIdentifierService(new DirectoryInfo(identifierLocation));
+            var identifierValue = await sut.GetIdentifier();
+            
+            // Assert: The identifier returned should be a valid ID
+            Assert.DoesNotThrow(() => Guid.Parse(identifierValue));
+            // Assert: The identifier file should not contain the original invalid value(s)
+            var identifierFileContents = File.ReadAllLines(identifierFullPath);
```
rhysparry, line 74: There is a mix of sync and async file methods. For something of this size I doubt it matters, but consistency is good.

source/Octopus.Manager.Tentacle.Tests/Utils/TentacleManagerInstanceIdentifierServiceFixture.cs
```diff
@@ -0,0 +1,82 @@
...
+            // Arrange: Ensure there is an existing but invalid identifier file
+            var identifierFullPath = Path.Combine(identifierLocation, TentacleManagerInstanceIdentifierService.IdentifierFileName);
+            var invalidLine1 = "This is not a valid ID string";
+            var invalidLine2 = "This contains multiple lines";
+            // ReSharper disable once ConvertToUsingDeclaration
+            using (var streamWriter = File.CreateText(identifierFullPath))
+            {
+                await streamWriter.WriteLineAsync(invalidLine1);
+                await streamWriter.WriteLineAsync(invalidLine2);
+            }
```
rhysparry, line 65: Why not `File.WriteAllLinesAsync` to set up this invalid file? Might make things a bit more succinct and avoids the need to suppress the ReSharper message.

source/Octopus.Manager.Tentacle.Tests/WizardFixtures/SetupTentacleWizardFixture.cs
```diff
@@ -31,20 +31,20 @@ public void WhenSettingUpListeningTentacle_ScriptShouldBeCorrectlyGenerated()
...
-                    Environment.NewLine + $"{pathToTentacleExe} service --instance \"{model.InstanceName}\" --install --stop --start";
+                Environment.NewLine + $"{pathToTentacleExe} new-certificate --instance \"{model.InstanceName}\" --if-blank" +
+                Environment.NewLine + $"{pathToTentacleExe} configure --instance \"{model.InstanceName}\" --reset-trust" +
+                Environment.NewLine + $"{pathToTentacleExe} configure --instance \"{model.InstanceName}\" --app \"C:\\Octopus\\Applications\\{model.InstanceName}\" --port \"{model.ListenPort}\" --noListen \"False\"" +
+                Environment.NewLine + $"{pathToTentacleExe} configure --instance \"{model.InstanceName}\" --trust \"{model.OctopusThumbprint}\"" +
+                Environment.NewLine + $"\"netsh\" advfirewall firewall add rule \"name=Octopus Deploy Tentacle\" dir=in action=allow protocol=TCP localport={model.ListenPort}" +
+                Environment.NewLine + $"{pathToTentacleExe} service --instance \"{model.InstanceName}\" --install --stop --start";
```
rhysparry, line 39: Did you consider using an array/list of strings and then joining? It might make things clearer than this massive string concatenation.

source/Octopus.Manager.Tentacle/Util/TelemetryService.cs
```diff
@@ -0,0 +1,31 @@
...
+    public interface ITelemetryService
+    {
+        Task<bool> SendTelemetryEvent(Uri serverUri, TelemetryEvent eventObj);
+    }
+    
+    public class TelemetryService: ITelemetryService
+    {
+        public async Task<bool> SendTelemetryEvent(Uri serverUri, TelemetryEvent eventObj)
+        {
+            using (var httpClient = new HttpClient())
```
rhysparry, line 19: Do we need to do anything special here? E.g. to deal with proxies?

rhysparry (commented):
Just a few thoughts.

raynhamdev (approved):
I've gone through the review with @acodrington over a Zoom meeting.
We've implemented some changes to take care of exceptions when sending Telemetry:
- Do not send Telemetry when setting up Listening Tentacles
- Ignore exceptions (eg. network issues) if we have them while sending to Telemetry to Server. We don't want Tentacle Manager to break if this happens.

Other than that, the rest of the changes in this PR look good to me. The implementation of  `SetupTentacleWizardModelBuilder` in the test project looks cool.

source/Octopus.Manager.Tentacle.Tests/WizardFixtures/SetupTentacleWizardFixture.cs
```diff
@@ -96,5 +96,23 @@ await telemetryService
...
+            var model = new SetupTentacleWizardModelBuilder()
+                .WithTelemetryService(telemetryService)
+                .Build();
+
+            model.CommunicationStyle = CommunicationStyle.TentaclePassive;
+
+            _ = await model.ReviewAndRunScriptTabViewModel.GenerateAndExecuteScript();
+
+            await telemetryService
+                .Received(0)
```
rhysparry, line 114: Would `DidNotReceive()` be clearer here?

https://nsubstitute.github.io/help/received-calls/#check-a-call-was-not-received

source/Octopus.Manager.Tentacle.Tests/Utils/TentacleManagerInstanceIdentifierServiceFixture.cs
acodrington: Good call, `GetTempPath` will probably return the same temp path each time. I'll revise.

source/Octopus.Manager.Tentacle.Tests/WizardFixtures/SetupTentacleWizardFixture.cs
acodrington: I didn't modify this, the whitespace must've gotten modified. Having said that, it still would be good to tidy this up.

source/Octopus.Manager.Tentacle/Util/TelemetryService.cs
acodrington: Hmmm good point. The wizard does take proxy settings, so we should use them here. Nice one!

source/Octopus.Manager.Tentacle.Tests/WizardFixtures/SetupTentacleWizardFixture.cs
acodrington: It would 🙇🏻

source/Octopus.Manager.Tentacle.Tests/WizardFixtures/SetupTentacleWizardFixture.cs
rhysparry: Yeah, this was just a general thing I noticed while looking through. This Pull Request has the disadvantage of being completely comprehensible without much additional context.

source/Octopus.Manager.Tentacle.Tests/Utils/TentacleManagerInstanceIdentifierServiceFixture.cs
acodrington: I'm updating these tests to use `File.WriteAllLines` as suggested in another comment, and that will make all usages sync.

rhysparry: Yeah, I agree sync is acceptable for these.

source/Octopus.Manager.Tentacle.Tests/Utils/TentacleManagerInstanceIdentifierServiceFixture.cs
acodrington: I resolved this by having `TentacleManagerInstanceIdentifierService` take the ID file path instead of a directory, which allows the tests to pass in a unique temp file path each test run.

source/Octopus.Manager.Tentacle.Tests/WizardFixtures/SetupTentacleWizardFixture.cs
acodrington: Haha I'm going to take that disadvantage as a compliment! 😁

PR #832, opened 2024-03-07
Add more Tentacle Manager unit tests

raynhamdev:
# Background

Before enabling Tentacle Manager to build on ,NET 6, we had added some happy path tests to ensure core functionalities should be working after the migration.

[sc-72557]

# Results

More use cases are now tested, such as:
- Reinstall, restart, start, stop an existing Tentacle instance.
- Create and install a new Tentacle instance
- Setting up a polling tentacle with server registration being skipped.
- Dynamically hide proxy tab when creating a polling tentacle instance.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Octopus.Manager.Tentacle.Tests/WizardFixtures/SetupTentacleWizardFixture.cs
```diff
@@ -70,6 +70,61 @@ public void WhenSettingUpPollingTentacle_ScriptShouldBeCorrectlyGenerated()
             script.Should().Be(expectedOutput);
         }
         
+        [Test]
+        public void WhenSettingUpPollingTentacle_AndServerRegistrationIsSkipped_ScriptShouldBeCorrectlyGenerated()
```
acodrington, line 74: I don't think skipping the registration is supported.

The `TentacleActiveDetailsTab` control does not enable the skip button, which means that the `SkipServerRegistration` property will never be set to `true` as [this happens as part of the `OnSkip` handler](https://github.com/OctopusDeploy/OctopusTentacle/blob/964282f313065081392018917f44d0fe75343897/source/Octopus.Manager.Tentacle/TentacleConfiguration/SetupWizard/Views/TentacleActiveDetailsTab.xaml.cs#L32).

Rather than add a test for functionality that has been deprecated, I think it would be better to add a TODO to remove the skip server registration functionality.

Adding a test implies that this functionality is important and should work, when in fact it just hasn't been cleaned up properly. We don't want to pay the cost of maintaining functionality and test code for a feature we never use.

source/Octopus.Manager.Tentacle.Tests/ConfiguringTentacleServiceFixture.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            // Arrange
+            var model = CreateTestTentacleManagerViewModel();
+            
+            // Act
+            var script = model.ServiceWatcher.GetRepairCommands().ToCommandLineString();
+
+            // Assert
+            var pathToTentacleExe = $"\"{CommandLinePath}\"";
+            var expectedOutput = $"{pathToTentacleExe} service --instance \"{model.InstanceName}\" --stop" +
```
acodrington, line 65: The multiple string concatenations feels a bit messy. Can we update to use a `StringBuilder` instead?

acodrington (changes requested):
This is mostly fine, but I think we should remove that test and add a TODO before merging.

.gitignore
```diff
@@ -54,3 +54,17 @@ _test
...
+
+build/.idea/.idea.build.dir/.idea/projectSettingsUpdater.xml
+
+build/.idea/.idea.build.dir/.idea/vcs.xml
+
+build/.idea/.idea.build.dir/.idea/workspace.xml
+
+installer/.idea/.idea.Installer/.idea/projectSettingsUpdater.xml
+
+installer/.idea/.idea.Installer/.idea/workspace.xml
```
acodrington, line 70: I'm nitpicking a bit now, but could you please update these lines to this:

`**/.idea/*`

That should exclude all the same files, and ensure that any new `.idea` files are also excluded. Better double-check that the the ignore still works correctly though, I'm not :100: about the syntax

acodrington (approved):

PR #835, opened 2024-03-13
Add v2 ExecuteScript overload to ITentacleClient

rhysparry:
Add an overload for ITentacleClient.ExecuteScript that accepts a `StartScriptCommandV2`

Using this overload will ensure a V2 endpoint is used to execute the script, falling back to V1 if V2 is unsupported.

This is intended to reduce confusion from the existing method which takes a v3alpha version of the command.

[SC-69441]

# Background

The `ExecuteScript` method in `ITentacleClient` was changed to accept a `StartScriptCommandV3Alpha`. Depending on the available support, the underlying orchestrator would map that back to a V2 (or V1) command. This led to confusion, as the method required users to configure the client to explicitly disable support for the alpha command.

With a new overload accepting `StartScriptCommandV2`, a user can call the method without the risk of using the v3 alpha code paths.

# Results

- New overload: `ITentacleClient.ExecuteScript` accepting a `StartScriptCommandV2`
- New overload: `IScriptOrchestrator.ExecuteScript` accepting a `StartScriptCommandV2`
- Internal `ScriptOrchestratorFactory` now considers a maximum script service version

## Before

- Confusion as to how to avoid the v3 alpha script service.
- No unit test coverage of orchestrator factory logic

## After

- Pit of success path to avoid the v3 alpha script service
- New unit tests covering the `ScriptOrchestratorFactory` logic for creating the appropriate `IScriptOrchestrator`

# How to review this PR

- Quality :heavy_check_mark:
- Unit tests were created in a separate commit and verify existing logic. They were updated in the subsequent commit to handle the new cases.

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Client.Tests/ScriptOrchestratorFactoryFixture.cs
+++ b/source/Octopus.Tentacle.Client.Tests/ScriptOrchestratorFactoryFixture.cs
@@ -78,5 +78,37 @@ namespace Octopus.Tentacle.Client.Tests
             orchestrator.Should().NotBeNull();
             orchestrator.Should().BeOfType<ScriptServiceV1Orchestrator>();
         }
+
+        [Test]
+        public async Task CreateOrchestratorReturnsV2WhenV2IsMaxVersion()
+        {
+            // Arrange
+            var factory = new ScriptOrchestratorFactoryBuilder()
+                .WithMaxScriptServiceVersion(ScriptServiceVersion.Version2)
+                .Build();
+
+            // Act
+            var orchestrator = await factory.CreateOrchestrator(CancellationToken);
+
+            // Assert
+            orchestrator.Should().NotBeNull();
+            orchestrator.Should().BeOfType<ScriptServiceV2Orchestrator>();
+        }
+
+        [Test]
+        public async Task CreateOrchestratorReturnsV1WhenV1IsMaxVersion()
+        {
+            // Arrange
+            var factory = new ScriptOrchestratorFactoryBuilder()
+                .WithMaxScriptServiceVersion(ScriptServiceVersion.Version1)
+                .Build();
+
+            // Act
+            var orchestrator = await factory.CreateOrchestrator(CancellationToken);
+
+            // Assert
+            orchestrator.Should().NotBeNull();
+            orchestrator.Should().BeOfType<ScriptServiceV1Orchestrator>();
+        }
     }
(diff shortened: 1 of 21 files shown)


source/Octopus.Tentacle.Client.Tests/ScriptOrchestratorFactoryFixture.cs
```diff
@@ -0,0 +1,114 @@
...
+using Octopus.Tentacle.Client.Scripts;
+using Octopus.Tentacle.Client.Tests.Builders;
+using Octopus.Tentacle.Contracts.ScriptServiceV2;
+
+namespace Octopus.Tentacle.Client.Tests
+{
+    [TestFixture]
+    public class ScriptOrchestratorFactoryFixture
+    {
+        readonly CancellationTokenSource cancellationTokenSource = new(TimeSpan.FromSeconds(42));
```
rhysparry, line 15: I applied this rather generous timeout to the tests. The substitutes used in these tests will mean there are no network requests, so I'm not expecting the cancellation token to ever be used.

Please let me know if there is a preferred approach to take.

rhysparry:
This will conflict with #799 so we probably don't want to proceed with this until that's resolved.

source/Octopus.Tentacle.Client.Tests/ScriptOrchestratorFactoryFixture.cs
LukeButters: Drop this comment in the code.

Also `42` ha nice

source/Octopus.Tentacle.Client/Scripts/ScriptServiceV3AlphaOrchestrator.cs
```diff
@@ -46,6 +47,11 @@ class ScriptServiceV3AlphaOrchestrator : ObservingScriptOrchestrator<StartScript
             this.logger = logger;
         }
 
+        protected override StartScriptCommandV3Alpha Map(StartScriptCommandV2 command)
+        {
+            throw new NotSupportedException("ScriptServiceV3AlphaOrchestrator does not support V2 commands");
```
LukeButters, line 52: Why not? I think in its current form v3 is a super set of v2.

LukeButters (approved):

rhysparry:
This has been made obsolete by #845

PR #836, opened 2024-03-14
Fix bug where pod monitoring stops working after a while for Kubernetes agent

flin-8:
[sc-73237]

# Background

https://github.com/OctopusDeploy/OctopusTentacle/pull/824 starting using a Watcher to get updates for Pods. Unfortunately the Kubernetes API Server will [close the connection after some time](https://stackoverflow.com/questions/51399407/watch-in-k8s-golang-api-watches-and-get-events-but-after-sometime-doesnt-get-an) (we can't have the watch open forever).

# Results

This PR adds a short timeout of 10min and reloads the status after that.

We also add a try/catch around the entire status loading so we don't silently lose track of Pod statuses.

Screenshot of Pod status loading at startup:
CleanShot 2024-03-14 at 13 26 44@2x

Status is reloaded after 10min:
CleanShot 2024-03-14 at 13 27 35@2x

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Tests/Scripts/ExponentialBackoffFixture.cs
+++ b/source/Octopus.Tentacle.Tests/Scripts/ExponentialBackoffFixture.cs
@@ -9,8 +9,6 @@ namespace Octopus.Tentacle.Tests.Scripts
     [TestFixture]
     public class ExponentialBackoffFixture
     {
-        readonly Foo foo = new Foo();
-
         [TestCase(1, 1)]
         [TestCase(2, 2)]
         [TestCase(3, 4)]
--- a/source/Octopus.Tentacle.Tests/Scripts/ExponentialBackoffFixture.cs
+++ b/source/Octopus.Tentacle.Tests/Scripts/ExponentialBackoffFixture.cs
@@ -1,0 +1,31 @@
+using System;
+using FluentAssertions;
+using NUnit.Framework;
+using Octopus.Tentacle.Diagnostics;
+using Octopus.Tentacle.Time;
+
+namespace Octopus.Tentacle.Tests.Scripts
+{
+    [TestFixture]
+    public class ExponentialBackoffFixture
+    {
+        readonly Foo foo = new Foo();
+
+        [TestCase(1, 1)]
+        [TestCase(2, 2)]
+        [TestCase(3, 4)]
+        [TestCase(4, 8)]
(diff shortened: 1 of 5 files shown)


source/Octopus.Tentacle/Kubernetes/KubernetesPodMonitor.cs
```diff
@@ -33,6 +35,21 @@ public KubernetesPodMonitor(IKubernetesPodService podService, ISystemLog log)
         }
 
         async Task IKubernetesPodMonitor.StartAsync(CancellationToken cancellationToken)
+        {
+            var maxDurationSeconds = 70;
```
scme0, line 39: nit: should this be a const?

scme0 (approved):
Nice, LGTM

source/Octopus.Tentacle.Tests/Scripts/ExponentialBackoffFixture.cs
```diff
@@ -0,0 +1,31 @@
...
+using NUnit.Framework;
+using Octopus.Tentacle.Diagnostics;
+using Octopus.Tentacle.Time;
+
+namespace Octopus.Tentacle.Tests.Scripts
+{
+    [TestFixture]
+    public class ExponentialBackoffFixture
+    {
+        readonly Foo foo = new Foo();
```
scme0, line 12: Should this be removed?

LukeButters: Don't touch that, its a load bearing `Foo` :stuck_out_tongue:

LukeButters (approved):
Since you need a SaST approval LGTM

Although I only looked at things outside of the k8s folder

source/Octopus.Tentacle.Tests/Scripts/ExponentialBackoffFixture.cs
flin-8: haha... I extracted that to a different class ...

source/Octopus.Tentacle/Kubernetes/KubernetesPodMonitor.cs
flin-8: 👍

PR #839, opened 2024-03-17
Replace bootstrapRunner with one written in golang

scme0:
# Background

#project-k8s-agent

[sc-73266]

The original bootstrapRunner is a bash script with the main problem being that we needed to add a sleep at the end of the script to ensure that the logs would be received successfully. This isn't scalable as we found that it could still fail under load.

We decided the best solution for this was to write an applet in golang to do the same thing. This means that we don't have to worry about what tooling is installed on the image we are running the pod with (especially if a user selects the image in a script step) and go binaries can be made very small (approx half a MB).

# Results

I've taken a go script that @liam-mackie put together for us (thanks Liam!) and I've added in the infrastructure to build it for arm64 and amd64, put it into the kubernetes tentacle container and use it as the bootstrapRunner. So far it seems to work in my tests.

This version doesn't include any additional functionality (like line numbers) as I wanted to get this in as a separate piece of work.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


docker/kubernetes-tentacle/Dockerfile
```diff
@@ -1,3 +1,26 @@
...
+
+#download upx and unpack
+RUN curl -OL "https://github.com/upx/upx/releases/download/v${UPX_VERSION}/upx-${UPX_VERSION}-${TARGETARCH}_${TARGETOS}.tar.xz"
+RUN tar -x -f "upx-${UPX_VERSION}-${TARGETARCH}_${TARGETOS}.tar.xz"
+
+# build bootstrapRunner
+# Note: the given ldflags remove debug symbols
+RUN go build -ldflags "-s -w" -o "bin/bootstrapRunner"
+# upx reduces the size of the exe
+RUN "./upx-${UPX_VERSION}-${TARGETARCH}_${TARGETOS}/upx" "bin/bootstrapRunner"
```
flin-8, line 22: Can you do a quick test and see what the error stack trace looks like? Want to make sure we can troubleshoot this thing

scme0: Screenshot 2024-03-18 at 14 24 02

scme0: This is what I get when I give it an invalid working directory.

scme0:
The ADR has been approved by Rob and is now merged 👍 
https://github.com/OctopusDeploy/adr/pull/3

docker/kubernetes-tentacle/Dockerfile
APErebus, line 22: Is there any performance issues with using `upx`?

docker/kubernetes-tentacle/bootstrapRunner/bootstrapRunner.go
```diff
@@ -0,0 +1,93 @@
...
+	err = cmd.Wait()
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "Error waiting for Cmd", err)
+		os.Exit(1)
+	}
+}
+
+func reader(scanner *bufio.Scanner, writer *bufio.Writer, done *chan bool) {
+	for scanner.Scan() {
+		message := fmt.Sprintf("%s|%s\n", time.Now().UTC().Format(time.RFC3339Nano), scanner.Text())
```
APErebus, line 82: I don't love that this doesn't have the UTC timezone in the timestamp (ISO 8601). We'll just need to make sure the Tentacle code parses these and knows they are UTC

source/Octopus.Tentacle/Kubernetes/KubernetesConfig.cs
```diff
@@ -22,6 +22,8 @@ public static class KubernetesConfig
         public static string HelmChartVersionVariableName => $"{EnvVarPrefix}__HELMCHARTVERSION";
         public static string HelmChartVersion => GetRequiredEnvVar(HelmChartVersionVariableName, "Unable to determine Helm chart version.");
 
+        public static string BootstrapRunnerExecutablePath => GetRequiredEnvVar("BOOTSTRAP_RUNNER_EXECUTABLE_PATH", "Unable to determine Bootstrap Runner Executable Path");
```
APErebus, line 25: Happy to have the discussion but we've typically not had `_` in the names (see helm chart version above). So it'd be `BOOTSTRAPRUNNEREXEPATH`

docker/kubernetes-tentacle/bootstrapRunner/bootstrapRunner.go
scme0: Good point, I'll have a fiddle with this to see if I can get ISO8601 exactly.

source/Octopus.Tentacle/Kubernetes/KubernetesConfig.cs
scme0: That's a fair point for consistency, but it does give me an aneurysm trying to read it with no separators... 😅

docker/kubernetes-tentacle/bootstrapRunner/bootstrapRunner.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	// Wait for output buffering first
+	<-doneStd
+	<-doneErr
+
+	err = cmd.Wait()
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "Error waiting for Cmd", err)
+		os.Exit(1)
```
APErebus, line 76: We should use the original exit code from the failed command here cc. @liam-mackie

docker/kubernetes-tentacle/bootstrapRunner/bootstrapRunner.go
scme0: Ok, apparently `RFC3339` is for all practical applications, ISO 8601 compliant: https://stackoverflow.com/q/522251

I'm getting `2024-03-19T01:28:24.226953Z` output, where the `Z` represents UTC I believe and I think this is compliant with ISO 8601.

docker/kubernetes-tentacle/bootstrapRunner/bootstrapRunner.go
scme0: I'm also thinking we shouldn't add an additional log here, should we? It seems like the bootstrapper script will log errors before it returns an error code?

docker/kubernetes-tentacle/Dockerfile
scme0: I've removed it 👍

docker/kubernetes-tentacle/bootstrapRunner/bootstrapRunner.go
scme0: I mean, the bootstrapper/script being run should manage logging it's errors before returning with a non-success exit code. This applet is only used to intercept logs and format them, not add additional logs. what do you think?

scme0: I've updated this, we now return the exit code we got from the bootstrapper command, and also we only log if it's not an error related to the failure exit code. (io error or some other weird thing).

source/Octopus.Tentacle/Kubernetes/KubernetesConfig.cs
scme0: Fixed 👍

docker/kubernetes-tentacle/bootstrapRunner/bootstrapRunner.go
APErebus: ok yep, that's good then 👍

APErebus (approved):
Looks excellent mate 👍

docker/kubernetes-tentacle/bootstrapRunner/bootstrapRunner.go
liam-mackie: Yeah, this was just a speed-hack to easily handle the error while I was hacking the script together. Some notes on my understanding:
* If it's an `ExitError` (see [here](https://pkg.go.dev/os/exec#ExitError)), we should definitely just exit with the code and not log (we already redirect stdErr, so `Stderr` in the `ExitError` would actually be `null` :shrug:)
* If it's not an `ExitError`, we should _probably_ just panic with the `Err`, since it's an unrecoverable state that we didn't expect, and we _probably_ want to have the stack. Some of the ways this could happen is being unable to write to a pipe or other OS internals. Panicking _also_ ends up running any of our `defer` code, which allows us to send telemetry if we implement it in future :)

scme0: That sounds reasonable. I'll look at making another PR to make it panic if it's not an `ExitError` 👍

PR #840, opened 2024-03-18
Properly propagate SIGTERM signal in Kubernetes Tentacle

APErebus:
# Background

Observation when updating the Kubernetes Agent tentacle container is that it always took 30s to terminate before the new container would start up. During this process, the terminating container would continue to handle halibut messages.

This indicated that Tentacle was not correctly handling the SIGTERM event that Kubernetes should be passing. Investigation indicated that because we were launching the tentacle as a sub-process of bash (which was a sub-process of the shell entrypoint), that this wasn't being passed down.

This PR updates the docker file and script

# Results

Changes the entrypoint of the kubernetes tentacle docker image to be a single bash script, which launches tentacle via `exec`, which replaces bash with the tentacle process. As a result, this becomes process with pid `1`, which is what receives the `SIGTERM` event.

image

This means that Tentacle now shutdowns much faster (<1s)

This was verified by a testing with fixed 10s thread sleep in the app process shutdown handler and observice that termination took 10s.

Shortcut story: [sc-73493]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/docker/kubernetes-tentacle/Dockerfile
+++ b/docker/kubernetes-tentacle/Dockerfile
@@ -18,7 +18,7 @@ COPY _artifacts/docker/tentacle_${BUILD_NUMBER}_${TARGETOS}-${TARGETARCH}${TARGE
 RUN apt-get update
 RUN apt install ./tentacle.deb -y
 RUN apt-get clean
-RUN apt install curl tini -y
+RUN apt install curl -y
 RUN rm -rf /var/lib/apt/lists/*
 
 WORKDIR /
@@ -53,8 +53,7 @@ ENV OCTOPUS__K8STENTACLE__DISABLEAUTOPODCLEANUP="False"
 ENV TentacleHome=""
 ENV TentacleApplications=""
 
-# We use tini here as the entrypoint so it correctly propagates SIGTERM signals to the tentacle child process (allowing for graceful shutdown to occur)
-ENTRYPOINT ["/usr/bin/tini", "-g", "--", "/scripts/configure-and-run.sh"]
+ENTRYPOINT ["/scripts/configure-and-run.sh"]
 
 LABEL \
     org.label-schema.schema-version="1.0" \
--- a/docker/kubernetes-tentacle/Dockerfile
+++ b/docker/kubernetes-tentacle/Dockerfile
@@ -53,7 +53,7 @@ ENV OCTOPUS__K8STENTACLE__DISABLEAUTOPODCLEANUP="False"
 ENV TentacleHome=""
 ENV TentacleApplications=""
 
-# Change the entrypoint to be tini
+# We use tini here as the entrypoint so it correctly propagates SIGTERM signals to the tentacle child process (allowing for graceful shutdown to occur)
 ENTRYPOINT ["/usr/bin/tini", "-g", "--", "/scripts/configure-and-run.sh"]
 
 LABEL \
--- a/docker/kubernetes-tentacle/Dockerfile
+++ b/docker/kubernetes-tentacle/Dockerfile
@@ -18,7 +18,7 @@ COPY _artifacts/docker/tentacle_${BUILD_NUMBER}_${TARGETOS}-${TARGETARCH}${TARGE
 RUN apt-get update
 RUN apt install ./tentacle.deb -y
 RUN apt-get clean
-RUN apt install curl -y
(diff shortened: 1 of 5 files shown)


docker/kubernetes-tentacle/scripts/configure-and-run.sh
```diff
@@ -230,13 +230,22 @@ function registerTentacle() {
     tentacle "${ARGS[@]}"
 }
 
+function run() {
+    echo "==============================================="
+    echo "Starting Octopus Deploy Kubernetes Tentacle"
+    echo "==============================================="
+
+    exec tentacle agent --instance Tentacle --noninteractive
```
APErebus, line 238: the `exec` here means that the host bash process is replaced by the `tentacle` process

docker/kubernetes-tentacle/Dockerfile
```diff
@@ -53,7 +53,8 @@ ENV OCTOPUS__K8STENTACLE__DISABLEAUTOPODCLEANUP="False"
 ENV TentacleHome=""
 ENV TentacleApplications=""
 
-CMD /scripts/configure-tentacle.sh && /scripts/run-tentacle.sh
+# Change the entrypoint to be tini
```
scme0, line 56: Maybe update this comment to a "why" instead of a "what"?

scme0 (approved):
Nice! Just one minor comment from me 👍

docker/kubernetes-tentacle/scripts/configure-and-run.sh
scme0: Maybe add this comment as a comment in the bash script?

docker/kubernetes-tentacle/Dockerfile
APErebus: yep, good call!

zentron:
Sounds like a good find, is this problem\solution any different to the standard Tentacle or Octopus Server Docker Image? If so, lets bring the "fix" across there, If not, why not?

docker/kubernetes-tentacle/Dockerfile
scme0: Sorry, I added another comment for the other exec part too 😅

APErebus: done!

APErebus:
> Sounds like a good find, is this problem\solution any different to the standard Tentacle or Octopus Server Docker Image? If so, lets bring the "fix" across there, If not, why not?

I let @OctopusDeploy/team-server-at-scale know about the standard tentacle docker image and they said they'd handle it as part of MQS.

I'm honestly not sure about server 🤔

APErebus:
PR was update to no longer use `tini` as it's not required. This is in-line with how server works (using `exec` to replace the host bash process)

scme0 (approved):
Event better! 🚀

docker/kubernetes-tentacle/scripts/configure-and-run.sh
```diff
@@ -230,13 +230,23 @@ function registerTentacle() {
...
+
+
 function verifyTentacleIsNotRegistered() {
   setupVariablesForRegistrationCheck
   getStatusOfRegistration
 
   if [ "$IS_REGISTERED" == "true" ]; then
       echo "Tentacle is already configured and registered with server."
-      exit 0
+      run
```
flin-8, line 249: Do you still need this if we have `run` at the bottom?

APErebus: yeah, otherwise we need to rejigger the logic flow so it doesn't run the configuration steps again. After the `exec` line, this script is stopped

flin-8: Ah ok... conceptually it's like this right?

```
isRegistered = HasBeenRegistered()

if (!isRegistered)
{
  configure....
}

run
```

How much effort would it be to rejig the flow?

flin-8 (approved):
LGTM + comment

docker/kubernetes-tentacle/scripts/configure-and-run.sh
APErebus: conceptually yes. Hrm, I don't think it would be _that_ hard 🤔
We can do another PR :)

PR #846, opened 2024-03-19
Automatic Cleanup of Orphaned Pods

scme0:
# Background

#project-k8s-agent

[sc-73516]

Sometimes a script pod gets orphaned, where for what ever reason server doesn't care about it anymore so it gets left in a completed or failed state, but it never gets cleaned up.

# Results

I've created a new task which creates a new task which checks for orphaned pods every 10 minutes. A pod is considered orphaned if it's in a non-running state, and has not received an update within 10 minutes.

When `OCTOPUS__K8STENTACLE__DISABLEAUTOPODCLEANUP` is `true`, orphaned pods are not deleted, but we log a verbose message saying we are not deleting the pod.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle/Commands/RunAgentCommand.cs
+++ b/source/Octopus.Tentacle/Commands/RunAgentCommand.cs
@@ -28,6 +28,7 @@ namespace Octopus.Tentacle.Commands
         readonly Lazy<IProxyInitializer> proxyInitializer;
         readonly Lazy<IWorkspaceCleanerTask> workspaceCleanerTask;
         readonly Lazy<IKubernetesPodMonitorTask> kubernetesPodMonitorTask;
+        readonly Lazy<IKubernetesOrphanedPodCleanerTask> kubernetesOrphanedPodCleanerTask;
 
         readonly ISleep sleep;
         readonly ISystemLog log;
@@ -37,7 +38,6 @@ namespace Octopus.Tentacle.Commands
         int wait;
         bool halibutHasStarted;
         bool workspaceCleanerHasStarted;
-        bool kubernetesPodMonitorHasStarted;
 
         public override bool CanRunAsService => true;
 
@@ -54,7 +54,8 @@ namespace Octopus.Tentacle.Commands
             AppVersion appVersion,
             ILogFileOnlyLogger logFileOnlyLogger,
             Lazy<IWorkspaceCleanerTask> workspaceCleanerTask,
-            Lazy<IKubernetesPodMonitorTask> kubernetesPodMonitorTask) : base(selector, log, logFileOnlyLogger)
+            Lazy<IKubernetesPodMonitorTask> kubernetesPodMonitorTask,
+            Lazy<IKubernetesOrphanedPodCleanerTask> kubernetesOrphanedPodCleanerTask) : base(selector, log, logFileOnlyLogger)
         {
             this.halibut = halibut;
             this.configuration = configuration;
@@ -68,6 +69,7 @@ namespace Octopus.Tentacle.Commands
             this.appVersion = appVersion;
             this.workspaceCleanerTask = workspaceCleanerTask;
             this.kubernetesPodMonitorTask = kubernetesPodMonitorTask;
+            this.kubernetesOrphanedPodCleanerTask = kubernetesOrphanedPodCleanerTask;
 
             Options.Add("wait=", "Delay (ms) before starting", arg => wait = int.Parse(arg));
             Options.Add("console", "Don't attempt to run as a service, even if the user is non-interactive", v =>
@@ -137,7 +139,7 @@ namespace Octopus.Tentacle.Commands
             if (!KubernetesConfig.ExecuteScriptsInLocalShell)
             {
                 kubernetesPodMonitorTask.Value.Start();
-                kubernetesPodMonitorHasStarted = true;
+                kubernetesOrphanedPodCleanerTask.Value.Start();
             }
 
             Runtime.WaitForUserToExit();
@@ -167,10 +169,16 @@ namespace Octopus.Tentacle.Commands
                 workspaceCleanerTask.Value.Stop();
             }
 
-            if (kubernetesPodMonitorHasStarted)
+            if (kubernetesPodMonitorTask.IsValueCreated)
             {
                 kubernetesPodMonitorTask.Value.Stop();
             }
+
+            if (kubernetesOrphanedPodCleanerTask.IsValueCreated)
+            {
+                kubernetesOrphanedPodCleanerTask.Value.Stop();
+            }
+
         }
     }
 }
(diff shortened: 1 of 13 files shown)


source/Octopus.Tentacle/Commands/RunAgentCommand.cs
```diff
@@ -167,10 +169,16 @@ protected override void Stop()
                 workspaceCleanerTask.Value.Stop();
             }
 
-            if (kubernetesPodMonitorHasStarted)
+            if (kubernetesPodMonitorTask.IsValueCreated)
             {
                 kubernetesPodMonitorTask.Value.Stop();
```
scme0, line 174: If start hasn't been called, Stop is a noop. We can just check if the Value has been created and if it has, call Stop no matter what. We really just want to avoid creating the instance at Stop time if it's not already created.

source/Octopus.Tentacle/Maintenance/WorkspaceCleanerTask.cs
```diff
@@ -2,84 +2,29 @@
...
         {
-            lock (taskLock)
-            {
-                if (cleanerTask is null) return;
-
-                try
-                {
-                    cancellationTokenSource.Cancel();
-
-                    cleanerTask.Wait();
```
scme0, line 66: I noticed this currently has an infinite wait. After the refactor, this becomes a 30s wait which I think is more reasonable.

source/Octopus.Tentacle/Kubernetes/KubernetesPodMonitor.cs
```diff
@@ -131,39 +152,68 @@ internal async Task OnNewEvent(WatchEventType type, V1Pod pod)
...
+        {
+            lookupLock.Wait();
+            try
+            {
+                return podStatusLookup.Values.ToList();
+            }
+            finally
+            {
+                lookupLock.Release();
+            }
```
APErebus, line 165: What if we made the `podStatusLookup` a `ConcurrentDictionary`? Could we avoid all the locking here?

scme0: I did that first, but `InitialLoadAsync` gets re-called every 10 mins, so we need to make sure that the OrphanedPodCleaner doesn't call in while `podStatusLookup` is being refilled with data.

APErebus (approved):
I think 10min is a good starting point. Could we make it a configurable value via an env var?

source/Octopus.Tentacle/Kubernetes/KubernetesPodMonitor.cs
APErebus: hrm, ok, we can live with this for now :)

scme0: I may have a solution that uses ConcurrentDictionary 👍 I'm just testing it.

source/Octopus.Tentacle/Kubernetes/KubernetesPodMonitor.cs
```diff
@@ -69,27 +75,35 @@ async Task UpdateLoop(CancellationToken cancellationToken)
...
 
                 log.Verbose($"Preloaded pod {pod.Name()}. {status}");
-                podStatusLookup[status.ScriptTicket] = status;
+                newStatuses[scriptTicket] = status;
             }
 
+            // Updating a reference is an atomic operation
+            // and we only add data to this dictionary
+            // in this class so this is thread safe.
+            podStatusLookup = newStatuses;
```
scme0, line 97: @APErebus I've updated the dictionary to be a `ConcurrentDictionary<>` and the main difference is that now we load up a new instance of the dictionary and just set the reference on the field. This is an atomic operation as far as I'm aware and because we don't do any updates to the dictionary outside of this class, we can be sure that no new values are added to the old podStatusLookup while we are building the new instance.

source/Octopus.Tentacle/Kubernetes/KubernetesPodMonitor.cs
```diff
@@ -103,22 +117,22 @@ internal async Task OnNewEvent(WatchEventType type, V1Pod pod)
                 {
                     case WatchEventType.Added or WatchEventType.Modified:
                     {
-                        if (!podStatusLookup.TryGetValue(scriptTicket, out var status))
-                        {
-                            status = new PodStatus(pod.GetScriptTicket());
-                            podStatusLookup[scriptTicket] = status;
-                        }
-
+                        var status = podStatusLookup.GetOrAdd(scriptTicket, st => new PodStatus(st, clock));
```
scme0, line 120: I've changed this to use an atomic method.

source/Octopus.Tentacle/Kubernetes/KubernetesPodMonitor.cs
```diff
@@ -103,22 +117,22 @@ internal async Task OnNewEvent(WatchEventType type, V1Pod pod)
...
                         log.Verbose($"Updated pod {pod.Name()} status. {status}");
 
                         break;
                     }
                     case WatchEventType.Deleted:
-                        log.Verbose($"Removed {type} pod {pod.Name()} status");
-
                         //if the pod is deleted, remove it
-                        podStatusLookup.Remove(scriptTicket);
+                        if (podStatusLookup.TryRemove(scriptTicket, out _))
```
scme0, line 128: handling a failure for Try remove by logging a warning. Any missed pods will be cleaned up at the refresh (every 10 mins).

source/Octopus.Tentacle/Kubernetes/KubernetesPodMonitor.cs
APErebus: yeah, that's a good idea 👍

APErebus (approved):
Looks good!

source/Octopus.Tentacle/Kubernetes/KubernetesConfig.cs
```diff
@@ -14,6 +14,8 @@ public static class KubernetesConfig
         // We default this to true if we can't parse the environment variable
         public static bool ExecuteScriptsInLocalShell => !bool.TryParse(Environment.GetEnvironmentVariable($"{EnvVarPrefix}__EXECUTEINLOCALSHELL"), out var executeInLocalShell) || executeInLocalShell;
         public static int PodMonitorTimeoutSeconds => int.TryParse(Environment.GetEnvironmentVariable($"{EnvVarPrefix}__PODMONITORTIMEOUT"), out var podMonitorTimeout) ? podMonitorTimeout : 10*60; //10min
+
+        public static TimeSpan PodsConsideredOrphanedAfterTimeSpan => TimeSpan.FromMinutes(int.TryParse(Environment.GetEnvironmentVariable($"{EnvVarPrefix}__PODSCONSIDEREDORPHANEDAFTERTIMESPAN"), out var podsConsideredOrphanedAfterTimeSpan) ? podsConsideredOrphanedAfterTimeSpan : 10);
```
APErebus, line 18: `PODSCONSIDEREDORPHANEDAFTERTIMESPAN` probably should be `PODSCONSIDEREDORPHANEDAFTERMINUTES`

scme0: Good call

PR #848, opened 2024-03-19
Force flush to log files before bootstrapRunner exits

APErebus:
# Background

We should be doubly sure that we've written all logs from the buffer before exiting

# Results

Added flush calls to both bufio writers

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/docker/kubernetes-tentacle/bootstrapRunner/bootstrapRunner.go
+++ b/docker/kubernetes-tentacle/bootstrapRunner/bootstrapRunner.go
@@ -80,9 +80,13 @@ func main() {
 	}
 
 	// Perform a final flush of the file buffers, just in case they didn't get flushed before
-	stdoutLogFile.Flush()
-	stderrLogFile.Flush()
-
+	if err := stdoutLogFile.Flush(); err != nil {	
+		fmt.Fprintln(os.Stderr, "bootstrapRunner.go: Failed to perform final flush of stdoutLogFile", err)
+	}
+	if err := stderrLogFile.Flush(); err != nil {	
+		fmt.Fprintln(os.Stderr, "bootstrapRunner.go: Failed to perform final flush of stderrLogFile", err)
+	}
+	
 	os.Exit(cmd.ProcessState.ExitCode())
 }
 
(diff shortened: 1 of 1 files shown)


docker/kubernetes-tentacle/bootstrapRunner/bootstrapRunner.go
```diff
@@ -79,6 +79,9 @@ func main() {
 		fmt.Fprintln(os.Stderr, "bootstrapRunner.go: Failed to execute bootstrap script", err)
 	}
 
+	stdoutLogFile.Flush()
+	stderrLogFile.Flush()
```
rhysparry, line 83: Do we need to do anything about the error returned?

APErebus: hrm, I thought about that, but I figured the program was existing anyway so you kinda don't care 🤔

docker/kubernetes-tentacle/bootstrapRunner/bootstrapRunner.go
```diff
@@ -79,6 +79,9 @@ func main() {
 		fmt.Fprintln(os.Stderr, "bootstrapRunner.go: Failed to execute bootstrap script", err)
 	}
 
+	stdoutLogFile.Flush()
```
scme0, line 82: I think it might be worth adding a comment saying these are here "just in case"

APErebus: done

docker/kubernetes-tentacle/bootstrapRunner/bootstrapRunner.go
rhysparry: I suppose we could try printing to stderr so there is a chance it can be seen somewhere. The most likely error that I can think of would be "no space left on device"

rhysparry (approved):
LGTM

docker/kubernetes-tentacle/bootstrapRunner/bootstrapRunner.go
liam-mackie: I agree - it's worth capturing and logging the error explicitly, even if it's just to let people looking at the code know that it _can_ fail!

APErebus: fair, I've updated them to write to stderr if it fails to flush

APErebus:
@rhysparry I've update the logic to write the any failed flushes to stderr

rhysparry (approved):
Great. Thanks!

scme0 (approved):
LGTM

PR #849, opened 2024-03-20
Use friendlier service decorators for CapabilitiesService

rhysparry:
# Background

We have decorator methods that better indicate the intent of the decorators. Using these decorators requires less juggling of generic type parameters and the need to pass the name of the decorated method.

# Results

- Minor changes made to `GetCapabilitiesClientDecorator` to support decorating a method with both before and after decorations.
- The `AfterGetCapabilities` decorator was updated to execute even if the decorated method raised an exception. The response will be `null` in this situation.
- Updates to tests to leverage these updated decorators

Shortcut: [SC-69397]

## Before

Example test code

```csharp
await using var clientAndTentacle = await tentacleConfigurationTestCase.CreateBuilder()
    .WithTentacleServiceDecorator(new TentacleServiceDecoratorBuilder()
        .HookServiceMethod<IAsyncClientCapabilitiesServiceV2, object, CapabilitiesResponseV2>(nameof(IAsyncClientCapabilitiesServiceV2.GetCapabilitiesAsync),
            null,
            async (_, response) =>
            {
                // Code to run after GetCapabilities goes here
            })
        .Build())
    .Build(CancellationToken);
```

## After

The same code using the decorators

```csharp
await using var clientAndTentacle = await tentacleConfigurationTestCase.CreateBuilder()
    .WithPortForwarder(out var portForwarder)
    .WithTentacleServiceDecorator(new TentacleServiceDecoratorBuilder()
        .DecorateCapabilitiesServiceV2With(d => d
            .AfterGetCapabilities(
                async response =>
                {
                    // Code to run after GetCapabilities goes here
                }))
        .Build())
    .Build(CancellationToken);
```

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Tests.Integration/CapabilitiesServiceV2Test.cs
+++ b/source/Octopus.Tentacle.Tests.Integration/CapabilitiesServiceV2Test.cs
@@ -56,21 +56,23 @@ namespace Octopus.Tentacle.Tests.Integration
             await using var clientAndTentacle = await tentacleConfigurationTestCase.CreateBuilder()
                 .WithPortForwarder(out var portForwarder)
                 .WithTentacleServiceDecorator(new TentacleServiceDecoratorBuilder()
-                    .HookServiceMethod<IAsyncClientCapabilitiesServiceV2, object, CapabilitiesResponseV2>(nameof(IAsyncClientCapabilitiesServiceV2.GetCapabilitiesAsync),
-                        null,
-                        async (_, response) =>
-                        {
-                            await Task.CompletedTask;
-
-                            capabilitiesResponses.Add(response);
-
-                            if (resumePortForwarder)
+                    .DecorateCapabilitiesServiceV2With(d => d
+                        .DecorateGetCapabilitiesWith(
+                            async (inner, options) =>
                             {
-                                // (2) Once a get capabilities call has been made which uses the cached response then resume normal RPC calls
-                                // to allow script execution to continue
-                                portForwarder.Value.ReturnToNormalMode();
-                            }
-                        })
+                                var response = await inner.GetCapabilitiesAsync(options);
+
+                                capabilitiesResponses.Add(response);
+
+                                if (resumePortForwarder)
+                                {
+                                    // (2) Once a get capabilities call has been made which uses the cached response then resume normal RPC calls
+                                    // to allow script execution to continue
+                                    portForwarder.Value.ReturnToNormalMode();
+                                }
+
+                                return response;
+                            }))
                     .Build())
                 .Build(CancellationToken);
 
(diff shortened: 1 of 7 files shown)


source/Octopus.Tentacle.Tests.Integration/CapabilitiesServiceV2Test.cs
```diff
@@ -56,21 +56,24 @@ public async Task CapabilitiesResponseShouldBeCached(TentacleConfigurationTestCa
...
-                        {
-                            await Task.CompletedTask;
+                    .DecorateCapabilitiesServiceV2With(d => d
+                        .AfterGetCapabilities(
+                            async response =>
+                            {
+                                await Task.CompletedTask;
 
-                            capabilitiesResponses.Add(response);
+                                if (response != null)
```
rhysparry, line 65: I added the null check here to handle the case where there is no response (due to an exception).

If it was `null` due to an exception, the exception will still be thrown by the service.

source/Octopus.Tentacle.Tests.Integration/Util/Builders/Decorators/CapabilitiesServiceV2DecoratorBuilder.cs
```diff
@@ -44,23 +39,47 @@ public CapabilitiesServiceV2DecoratorBuilder DecorateGetCapabilitiesWith(GetCapa
...
+                try
+                {
+                    await beforeGetCapabilities(inner);
+                    response = await getCapabilitiesFunc(inner, options);
+                    return response;
+                }
+                finally
+                {
+                    await afterGetCapabilities(response);
+                }
```
rhysparry, line 82: The various script service decorators didn't need to do this as a new builder is created for before and after actions, rather than the shared builder approach used here.

source/Octopus.Tentacle.Tests.Integration/CapabilitiesServiceV2Test.cs
```diff
@@ -56,21 +56,24 @@ public async Task CapabilitiesResponseShouldBeCached(TentacleConfigurationTestCa
             await using var clientAndTentacle = await tentacleConfigurationTestCase.CreateBuilder()
                 .WithPortForwarder(out var portForwarder)
                 .WithTentacleServiceDecorator(new TentacleServiceDecoratorBuilder()
-                    .HookServiceMethod<IAsyncClientCapabilitiesServiceV2, object, CapabilitiesResponseV2>(nameof(IAsyncClientCapabilitiesServiceV2.GetCapabilitiesAsync),
-                        null,
-                        async (_, response) =>
-                        {
-                            await Task.CompletedTask;
+                    .DecorateCapabilitiesServiceV2With(d => d
+                        .AfterGetCapabilities(
```
LukeButters, line 60: I think in this case we just want to actually decorate it with the regular decorator pattern e.g.

.DecorateCapabilitiesServiceV2With(d => d.DecorateGetCapabilities(

async (paramasToGetCapabilities, innerGetCapbilitiesService) => {
   // do stuff
    var response = innerGetCapabilitiesService.GetCapabilities(paramasToGetCapabilities);
    capabilitiesResponses.Add(response);
    return response;
    
}

));

LukeButters (approved):

PR #845, opened 2024-03-19
Change TentacleClient.ExecuteScript to use new ExecuteScriptCommand type

APErebus:
# Background

Currently, `TentacleClient.ExecuteScript()`  takes a `StartScriptCommandV3Alpha`, which ties the contract for executing a script to a specific script service version (in this case V3Alpha).

This means that the contract of TentacleClient is tied to the latest version of the Script Service, which then means over the halibut-wire contracts are being propagated throughout Server's codebase. There is an additional problem then for the upcoming `KubernetesScriptServiceV1Alpha` as it will have it's own `StartScriptCommand` and `ScriptServiceV3Alpha` will be going away.

This PR introduces a new abstract `ExecuteScriptCommand` and two concrete implementations, `ExecuteShellScriptCommand` and `ExecuteKubernetesScriptCommand` types which are now the public contract for `TentacleClient.ExecuteScript()`.

# Results

Introduced the three new types and updated the TentacleClient and script orchestration to use the new types. Also included the builders I developed in https://github.com/OctopusDeploy/OctopusDeploy/pull/23506 into the TentacleClient so we have consistent builders for consumption

Almost all the changes in the tests are moving away from the `LatestStartScriptCommandBuilder` to a new `TestExecuteShellScriptCommandBuilder`. The `StartScriptCommandV3Alpha` build remains as it's required for the `ScriptServiceV3AlphaFixture`

Shortcut story: [sc-73757]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

- [ ] I have read [How we use GitHub Issues](https://github.com/OctopusDeploy/Issues/blob/main/docs/CONTRIBUTING.internal.md) for help deciding when and where it's appropriate to make an issue.
- [x] I have considered informing or consulting the right people, according to the [ownership map](https://whimsical.com/ownership-map-NzbiD4HJyvhC9jNJNfS6TG).
- [x] I have considered appropriate testing for my change.

PR #852, opened 2024-03-21
[WIP] Rename ScriptServiceV3Alpha to KubernetesScriptServiceV1Alpha

APErebus:
# Background

Now that we are no longer relying on the `StartScriptCommandV3Alpha` in the `TentacleClient.ExecuteScript()` contract (#845), we can rename the `ScriptServiceV3Alpha` to `KubernetesScriptServiceV1Alpha` without changing the public contract of `TentacleClient`.

Apart from the rename, this also changes `KubernetesScriptServiceV1Alpha` to only execute scripts via Kubernetes pods.

# Results

The Capabilities service is what tells `TentacleClient` which  services are supported on the target Tentacle. This PR changes this so that if the tentacle is known be running as part of the Kubernetes Agent, then we don't return any of the non-Kubernetes script services (and visa-versa). This gives us a lot more control about what services are advertised for a given tentacle and how TentacleClient uses this.

Unfortunately this PR does remove the tests for the previous `ScriptServiceV3Alpha` because they were executing via local shell, which is not supported in `KubernetesScriptServiceV1Alpha`. We will be adding tests for this new service in the future as it's quite challenging due to requiring a running Kubernetes cluster

Shortcut story: [sc-73757]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Octopus.Tentacle.Client/Scripts/ScriptOrchestratorFactory.cs
```diff
@@ -68,19 +68,22 @@ public async Task<IScriptOrchestrator> CreateOrchestrator(CancellationToken canc
                 throw new OperationCanceledException("Script execution was cancelled", ex);
             }
 
-            return scriptServiceToUse switch
+            if (scriptServiceToUse == ScriptServiceVersion.ScriptServiceVersion1)
```
APErebus, line 71: `ScriptServiceToUse` is no longer an enum, so can't be switched on

source/Octopus.Tentacle.Client/Scripts/ScriptOrchestratorFactory.cs
```diff
@@ -127,19 +133,11 @@ async Task<CapabilitiesResponseV2> GetCapabilitiesFunc(CancellationToken ct)
 
             logger.Verbose($"Discovered Tentacle capabilities: {string.Join(",", tentacleCapabilities.SupportedCapabilities)}");
 
-            if (tentacleCapabilities.HasScriptServiceV3Alpha())
+            // Check if we support any kubernetes script service.
+            // It's implied (and tested) that GetCapabilities will only return Kubernetes or non-Kubernetes script services, never a mix
+            if (tentacleCapabilities.HasAnyKubernetesScriptService())
```
APErebus, line 138: This is a key feature of this PR, if the Capabilities service returns _any_ kubernetes script service, we only look for versions in _that_ hierarchy.

If not, we only look for versions of in the `ScriptService` hierarchy

source/Octopus.Tentacle.Client/TentacleClientOptions.cs
```diff
@@ -6,11 +6,6 @@ namespace Octopus.Tentacle.Client
     public class TentacleClientOptions
     {
         public RpcRetrySettings RpcRetrySettings { get; }
-        
-        /// <summary>
-        /// Disables the use of ScriptServiceV3Alpha, even if it's supported on the Tentacle
-        /// </summary>
-        public bool DisableScriptServiceV3Alpha { get; set; }
```
APErebus, line 13: We don't need to disable this as the capabilities service now dictates what services are available.

source/Octopus.Tentacle.Contracts/ClientServices/IAsyncClientKubernetesScriptServiceV1Alpha.cs
APErebus, line 1: rename + updated contract types

source/Octopus.Tentacle/Commands/RunAgentCommand.cs
```diff
@@ -53,9 +51,7 @@ public class RunAgentCommand : AbstractStandardCommand
             IWindowsLocalAdminRightsChecker windowsLocalAdminRightsChecker,
             AppVersion appVersion,
             ILogFileOnlyLogger logFileOnlyLogger,
-            Lazy<IWorkspaceCleanerTask> workspaceCleanerTask,
-            Lazy<IKubernetesPodMonitorTask> kubernetesPodMonitorTask,
-            Lazy<IKubernetesOrphanedPodCleanerTask> kubernetesOrphanedPodCleanerTask) : base(selector, log, logFileOnlyLogger)
+            IEnumerable<Lazy<IBackgroundTask>> backgroundTasks) : base(selector, log, logFileOnlyLogger)
```
APErebus, line 54: Rather than import each individual `IBackgroundTask`, we now import them all and just lazily build them and only stop if they were lazily created

source/Octopus.Tentacle/Util/PlatformDetection.cs
```diff
@@ -12,9 +13,9 @@ public static class PlatformDetection
         public static class Kubernetes
         {
             /// <summary>
-            /// Indicates if the Tentacle is running inside a Kubernetes cluster.
+            /// Indicates if the Tentacle is running inside a Kubernetes cluster as the Kubernetes Agent. This is done by checking if the namespace environment variable is set
             /// </summary>
-            public static bool IsRunningInKubernetes => bool.TryParse(Environment.GetEnvironmentVariable("OCTOPUS__K8STENTACLE__FORCE"), out var b) && b;
+            public static bool IsRunningAsKubernetesAgent => !string.IsNullOrWhiteSpace(Environment.GetEnvironmentVariable(KubernetesConfig.NamespaceVariableName));
```
APErebus, line 18: Renamed this for clarity as you can run the normal tentacle in Kubernetes, but we are really using this bool when it's running as the kubernetes agent

source/Octopus.Tentacle/Services/Scripts/KubernetesScriptServiceV1Alpha.cs
```diff
@@ -71,33 +70,21 @@ public async Task<ScriptStatusResponseV3Alpha> StartScriptAsync(StartScriptComma
...
                 runningScript.Process = process;
 
-                if (command.DurationToWaitForScriptToFinish != null)
-                {
-                    var waited = Stopwatch.StartNew();
-                    while (process.State != ProcessState.Complete && waited.Elapsed < command.DurationToWaitForScriptToFinish.Value)
-                    {
-                        Thread.Sleep(TimeSpan.FromMilliseconds(10));
-                    }
-                }
```
APErebus, line 88: This property is not on the command anymore and we don't want this logic as we _only_ run in polling mode

source/Octopus.Tentacle/Program.cs
```diff
@@ -57,7 +57,12 @@ public override IContainer BuildContainer(StartUpInstanceRequest startUpInstance
             builder.RegisterModule(new ServicesModule());
             builder.RegisterModule(new VersioningModule(GetType().Assembly));
             builder.RegisterModule(new MaintenanceModule());
-            builder.RegisterModule<KubernetesModule>();
+
+            //we don't register any of the kubernetes _stuff_ if we aren't running as a kubernetes agent
+            if (PlatformDetection.Kubernetes.IsRunningAsKubernetesAgent)
```
APErebus, line 62: Because the `KubernetesScriptServiceV1Alpha` takes a direct dependency on the Kubernetes code, it was causing k8s services to be registered, where they are expecting to be running in a cluster. This switch means that we only register the `KubernetesModule` when we know we are running as the Kubernetes Agent.

APErebus:
Closing as this has been supersceeded by #859

PR #853, opened 2024-03-21
Expose IAsyncCapabilitiesServiceV2, maintaining location of service

LukeButters:
# Background

@samdanaei  This is the interface we needed in your test PR.

We should always have exposed this interface like this.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Contracts/Services/Capabilities/IAsyncCapabilitiesServiceV2.cs
+++ b/source/Octopus.Tentacle.Contracts/Services/Capabilities/IAsyncCapabilitiesServiceV2.cs
@@ -4,10 +4,13 @@ using System.Threading.Tasks;
 using Halibut.Transport.Caching;
 using Octopus.Tentacle.Contracts.Capabilities;
 
+// Don't "fix" this namespace, this is what we originally did and perhaps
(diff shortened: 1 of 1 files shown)


source/Octopus.Tentacle.Contracts/Services/Capabilities/IAsyncCapabilitiesServiceV2.cs
```diff
@@ -4,10 +4,13 @@
 using Halibut.Transport.Caching;
 using Octopus.Tentacle.Contracts.Capabilities;
 
+// Don't "fix" this namespace, this is what we originally did and perhaps
+// what we need to keep doing. 
 namespace Octopus.Tentacle.Services.Capabilities
```
LukeButters, line 9: If changing this to what rider wants it to be works and the test passes then it should be safe to do as we have backwards compatibility testing. Also I don't think package location matters for services.

LukeButters: But also this is the most as-is change as possible

samdanaei (approved):

PR #764, opened 2024-01-16
Move the StartScriptCommandBuilders to the Tentacle.Contracts to allow consumers of TentacleClient to use the builders.

nathanwoctopusdeploy:
# Background

[sc-66868]

As part of upgrading DynamicWorkers to use TentacleClient, a set of classes were copied from Server / Tentacle into DynamicWorkers. 

One of the areas was the StartScriptCommandBuilders. The builders provide a convenient way of creating the start script commands.

# Results

This PR exposes a set of StartScriptCommand builders in the Tentacler.Contracts library.
The builders were moved from the testing project and refactored to have better sensible defaults for consumers.

The idea would be that these builders can be used by DyanmicWorkers and also Server as an option refactor.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

- [ ] I have read [How we use GitHub Issues](https://github.com/OctopusDeploy/Issues/blob/main/docs/CONTRIBUTING.internal.md) for help deciding when and where it's appropriate to make an issue.
- [ ] I have considered informing or consulting the right people, according to the [ownership map](https://whimsical.com/ownership-map-NzbiD4HJyvhC9jNJNfS6TG).
- [ ] I have considered appropriate testing for my change.

PR #855, opened 2024-03-21
Move start script command builders to test utils

APErebus:
# Background

As part of the feedback for #845 from @LukeButters, he recommended certain changes be a separate PR, so here it is 😄 

This PR moves the start script command builders into the `CommonTestUtils` project. These builders are only used in the tests as well as in Server, however with the follow up PR #845 and https://github.com/OctopusDeploy/OctopusDeploy/pull/23506, these are no longer to be used in Server.

These were move into the contracts library in Jan by @nathanwoctopusdeploy  in #764 so they could be used in DynamicWorkers, in the upcoming PR they aren't used in TentacleClient, so they don't need to be in the contracts library

# Results

I also change 2 default values for the `StartScriptCommandV3Alpha` to be consistent with the `LatestStartScriptCommandBuilder`

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.CommonTestUtils/Builders/StartScriptCommandV2BuilderExtensionMethods.cs
+++ b/source/Octopus.Tentacle.CommonTestUtils/Builders/StartScriptCommandV2BuilderExtensionMethods.cs
@@ -1,7 +1,5 @@
 ﻿using System;
 using System.Text;
-using Octopus.Tentacle.Contracts.Builders;
-using Octopus.Tentacle.Tests.Integration.Util.Builders;
 
 namespace Octopus.Tentacle.CommonTestUtils.Builders
 {
@@ -14,23 +12,6 @@ namespace Octopus.Tentacle.CommonTestUtils.Builders
             builder.WithScriptBody(scriptBody.ToString());
 
             return builder;
-        }
-
-        public static StartScriptCommandV2Builder WithScriptBody(this StartScriptCommandV2Builder builder, ScriptBuilder scriptBuilder)
-        {
-            var scriptBody = new StringBuilder(scriptBuilder.BuildForCurrentOs());
-
-            builder.WithScriptBody(scriptBody.ToString());
-
-            return builder;
-        }
-
-        public static StartScriptCommandV2Builder WithScriptBody(this StartScriptCommandV2Builder builder, Action<ScriptBuilder> builderFunc)
-        {
-            var scriptBuilder = new ScriptBuilder();
-            builderFunc(scriptBuilder);
-
-            return builder.WithScriptBody(scriptBuilder);
         }
     }
 }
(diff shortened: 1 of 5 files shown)


source/Octopus.Tentacle.CommonTestUtils/Builders/StartScriptCommandV2BuilderExtensionMethods.cs
```diff
@@ -15,22 +13,5 @@ public static StartScriptCommandV2Builder WithScriptBodyForCurrentOs(this StartS
...
-            return builder;
-        }
-
-        public static StartScriptCommandV2Builder WithScriptBody(this StartScriptCommandV2Builder builder, Action<ScriptBuilder> builderFunc)
-        {
-            var scriptBuilder = new ScriptBuilder();
-            builderFunc(scriptBuilder);
-
-            return builder.WithScriptBody(scriptBuilder);
-        }
```
APErebus, line 34: These were unused

scme0 (approved):
LGTM

source/Octopus.Tentacle.CommonTestUtils/Builders/StartScriptCommandV3AlphaBuilder.cs
```diff
@@ -1,22 +1,24 @@
...
+namespace Octopus.Tentacle.CommonTestUtils.Builders
 {
     public class StartScriptCommandV3AlphaBuilder
     {
         readonly List<ScriptFile> files = new();
         readonly List<string> arguments = new();
         readonly Dictionary<ScriptType, string> additionalScripts = new();
         StringBuilder scriptBody = new(string.Empty);
-        ScriptIsolationLevel isolation = ScriptIsolationLevel.FullIsolation;
+        ScriptIsolationLevel isolation = ScriptIsolationLevel.NoIsolation;
```
rhysparry, line 16: Why did this change? I would have thought that `FullIsolation` would be a safe default. Just after some clarification, thanks.

rhysparry (approved):
LGTM

source/Octopus.Tentacle.CommonTestUtils/Builders/StartScriptCommandV3AlphaBuilder.cs
APErebus: So chasing this through, It looks like Nathan changed this in #764 as he wanted to use the builders in [OctopusDeploy/DynamicWorkers](https://github.com/OctopusDeploy/DynamicWorkers/pull/683)

They were previous to that change had no isolation, but then were changed to full isolation as a default for the now-public builders.

I've followed up with CloudPT to let them know dynamic workers will need updating with the new contract changes in #845

PR #856, opened 2024-03-26
Trap SIGTERM and send to tentacle

### LINKED ISSUE #865: Container Tentacle does not shut down gracefully on SIGTERM
### Team

- [X] I've assigned a team label to this issue

### What happened?

When a container is stopped, SIGTERM is sent to the container to allow it to shut down before it is forcibly killed. The tentacle container image does not correctly forward this signal to the tentacle.

### Reproduction

Run the tentacle container and stop it using `docker container stop`

### Error and Stacktrace

_No response_

### More Information

_No response_

### Workaround

_No response_

rhysparry:
# Background

When a container is shut down (e.g., via `docker stop`), the root process is sent the `SIGTERM` signal and given a grace period to shut down before being sent `SIGKILL`. Tentacle has support for handling this event, but due to script nesting, the signal is not propagated to the Tentacle process.

Including Docker-In-Docker introduces additional complexity, as both processes need to receive the signal. The container should also coordinate the independent failure of either process and signal the other to terminate.

# Results

- Graceful shutdown will occur when stopping a Tentacle container.
- If the Docker in Docker daemon terminates (e.g. crashes), the Tentacle will be terminated gracefully.
- Shortcut [SC-73505]
- Resolves #865 

# How this was tested

To test the PR, I created a small application that would run in the container, listen for and log signals and terminate on SIGTERM. I replaced `tentacle` and `dockerd-entrypoint.sh` with this application.

I then ran various scenarios against the container:

- Terminate the container using the `docker container stop` command to see both processes handle the signal and the container exits.
- Terminate the `tentacle` process within the container to see that `dockerd-entrypoint.sh` is terminated and the container exits.
- Terminate the `dockerd-entrypoint.sh` process within the container to see that `tentacle` is terminated and the container exits.

I also ran the tentacle from the CI build in multiple configurations to verify it was working correctly:

- Listening Worker
- Polling Worker
- Listening Tentacle
- Polling Tentacle

I successfully ran scripts against the tentacle, including using docker where it was enabled.

# How to review this PR

Quality :heavy_check_mark:

I've signposted the diff significantly.

Please consider any risks associated with terminating the tentacle if Docker-In-Docker is terminated.

# Pre-requisites

### DIFF
--- a/docker/linux/Dockerfile
+++ b/docker/linux/Dockerfile
@@ -42,6 +42,8 @@ RUN rm -rf /tmp/*
 
 ENV OCTOPUS_RUNNING_IN_CONTAINER=Y
 ENV ACCEPT_EULA=N
+ENV AsKubernetesTentacle=""
+ENV BearerToken=""
 ENV CustomPublicHostName=""
 ENV DISABLE_DIND=N
 ENV ListeningPort=""
--- a/docker/linux/Dockerfile
+++ b/docker/linux/Dockerfile
@@ -64,7 +64,7 @@ ENV TargetWorkerPool=""
 
 VOLUME /var/lib/docker
 
-CMD /scripts/configure-tentacle.sh && /scripts/run-tentacle.sh
+ENTRYPOINT [ "/scripts/configure-and-run.sh" ]
 
 LABEL \
     org.label-schema.schema-version="1.0" \
--- a/docker/linux/scripts/configure-and-run.sh
+++ b/docker/linux/scripts/configure-and-run.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+set -eu
+
+/scripts/configure-tentacle.sh
+CONFIGURE_RESULT=$?
+if [ "$CONFIGURE_RESULT" != "0" ]; then
+  exit $CONFIGURE_RESULT
+fi
+
+exec /scripts/run-tentacle.sh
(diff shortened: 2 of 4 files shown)


acodrington (approved):
This area is unfamiliar to me, so Rhys and I paired over Zoom to go through the PR. I understand the changes now and they LGTM.

There is a subtle change in behaviour regarding the exit codes i.e. Docker-in-Docker error exit code was previously being ignored, so the container could keep running even if there was an error, but the error code is now being handled and will shutdown the container.

As we discussed over Zoom, it would be great to verify the best method of communicating this change to customers. Even though the change is sensible, we don't want consumers of this container to be blind-sided by the different behaviour.

rhysparry:
If users experience excessive Tentacle container terminations as a result of this change, the following options are available:

Set `DISABLE_DIND=Y` in the environment. This will disable Docker in the container, but if it was failing anyway this shouldn't change much.

They could build a custom container that restores the original behaviour. The Dockerfile would be relatively straightforward:

```Dockerfile
FROM octopusdeploy/tentacle:8.1.1226 AS old-scripts
FROM octopusdeploy/tentacle:latest

COPY --from=old-scripts /scripts/*.sh /scripts/

ENTRYPOINT ["/bin/sh", "-c", "/scripts/configure-tentacle.sh && /scripts/run-tentacle.sh"]
```

This would overwrite the updated scripts and reconfigure the entrypoint.

sahanocto:
Just a data point, Cloud Portal had some E2E tests that spun up a Tentacle using the `octopusdeploy/tentacle:latest` Docker image, and they broke due to this change. Thanks for the helpful comments in this PR, we were able to get things going again with the `DISABLE_DIND` environment variable.

PR #858, opened 2024-03-26
Only register Kubernetes dependencies when running as Kubernetes Agent

APErebus:
# Background

To reduce the possibility of Kubernetes Agent-specific code being executed during normal Tentacle operation, the `KubernetesModule` is now registered only when running as the agent.

# Results

There is one place that this caused issues. The `RunAgentCommand` starts two kubernetes-specific background tasks, the `IKubernetesPodMonitorTask` and the `IKubernetesOrphanedPodCleanerTask`.

To fix this, we now just add _all_ registered `IBackgroundTask` types and start them all (and stop them if they were started). This means we don't need to worry about which tasks have been registered or not

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle/Commands/RunAgentCommand.cs
+++ b/source/Octopus.Tentacle/Commands/RunAgentCommand.cs
@@ -1,9 +1,12 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Runtime.InteropServices;
 using System.Security.Cryptography;
 using System.Security.Principal;
 using Octopus.Diagnostics;
+using Octopus.Tentacle.Background;
 using Octopus.Tentacle.Communications;
 using Octopus.Tentacle.Configuration;
 using Octopus.Tentacle.Configuration.Instances;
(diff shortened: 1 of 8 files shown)


scme0 (approved):
LGTM

source/Octopus.Tentacle/Commands/RunAgentCommand.cs
```diff
@@ -164,19 +156,9 @@ protected override void Stop()
...
-                workspaceCleanerTask.Value.Stop();
-            }
-
-            if (kubernetesPodMonitorTask.IsValueCreated)
-            {
-                kubernetesPodMonitorTask.Value.Stop();
-            }
-
-            if (kubernetesOrphanedPodCleanerTask.IsValueCreated)
+            foreach (var backgroundTaskLazy in backgroundTasks.Where(bt => bt.IsValueCreated))
```
APErebus, line 159: We can just check to see if the Lazy value has been created before we call `Stop()`.

evolutionise (approved):

PR #860, opened 2024-03-27
Change the WorkspaceCleaner to load script services by new interface

APErebus:
# Background

Workspace cleaner is resolving the script services by name via a service locator-style pattern, which is generally considered an [anti-pattern](https://blog.ploeh.dk/2010/02/03/ServiceLocatorisanAnti-Pattern/).

This will also cause an issue in a forthcoming PR as the `KubernetesScriptServiceV1Alpha` will not be registered for normal tentacles, so we need a way to not resolve the services directly.

# Results

This PR introduces a new interface, `IRunningScriptReporter` which is added to the appropriate script services and exposes a single `IsRunningScript(ScriptTicket)` method.

Then the workspace cleaner receives a list of registered `IRunningScriptReporter`'s, which it then can do an `Any()` on to see if any of the services are running that script ticket.

To support this change, halibut services are now registered with the `AsImplementedInterfaces()`

Shortcut story: [[sc-73757](https://app.shortcut.com/octopusdeploy/story/73757)]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle/Communications/AutofacServiceFactory.cs
+++ b/source/Octopus.Tentacle/Communications/AutofacServiceFactory.cs
@@ -13,7 +13,7 @@ namespace Octopus.Tentacle.Communications
     /// However, before resolving, one or more IAutofacServiceSources must also be registered.
     /// This is so that only explicitly specified services will be resolved, and the types of all messages are known when the service is instantiated.
     /// </summary>
-    public class AutofacServiceFactory : IServiceFactory, IServiceRegistration, IDisposable
+    public class AutofacServiceFactory : IServiceFactory, IDisposable
     {
         // Must never be modified as it is required for backwards compatability in BackwardsCompatibleCapabilitiesV2Decorator
         const string TentacleServiceShuttingDownMessage = "The Tentacle service is shutting down and cannot process this request.";
@@ -70,11 +70,6 @@ namespace Octopus.Tentacle.Communications
         public void Dispose()
         {
             scope.Dispose();
-        }
-
-        public T GetService<T>()
-        {
-            return scope.Resolve<T>();
         }
     }
 }
--- a/source/Octopus.Tentacle/Maintenance/WorkspaceCleaner.cs
+++ b/source/Octopus.Tentacle/Maintenance/WorkspaceCleaner.cs
@@ -1,5 +1,8 @@
 ﻿using System;
+using System.Collections;
+using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Octopus.Diagnostics;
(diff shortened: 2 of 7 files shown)


source/Octopus.Tentacle/Maintenance/WorkspaceCleaner.cs
```diff
@@ -51,9 +48,10 @@ public async Task Clean(CancellationToken cancellationToken)
 
                 try
                 {
-                    if (scriptService.IsRunningScript(workspace.ScriptTicket)) continue;
-                    if (scriptServiceV2.IsRunningScript(workspace.ScriptTicket)) continue;
-                    if(scriptServiceV3Alpha.IsRunningScript(workspace.ScriptTicket)) continue;
+                    if (runningScriptReporters.Any(x => x.IsRunningScript(workspace.ScriptTicket)))
```
flin-8, line 51: Interesting - each service keeps its own running scripts in memory.

This is one more in-memory thing we need to take out for K8s?

APErebus: The implementation in the k8s script service will just look for an existing pod via the PodStatusMonitor

source/Octopus.Tentacle/Maintenance/WorkspaceCleaner.cs
```diff
@@ -51,9 +48,11 @@ public async Task Clean(CancellationToken cancellationToken)
 
                 try
                 {
-                    if (scriptService.IsRunningScript(workspace.ScriptTicket)) continue;
-                    if (scriptServiceV2.IsRunningScript(workspace.ScriptTicket)) continue;
-                    if(scriptServiceV3Alpha.IsRunningScript(workspace.ScriptTicket)) continue;
+                    log.Info($"runningScriptReporters: {runningScriptReporters.Count()}");
```
sburmanoctopus, line 51: Nit: This is double enumeration. If logging count is important, maybe we should make `runningScriptReporters` a list, and ToList it in the constructor?

source/Octopus.Tentacle/Communications/AutofacServiceFactory.cs
```diff
@@ -49,8 +37,12 @@ public IServiceLease CreateService(string serviceName)
             {
                 if (knownServices.TryGetValue(serviceName, out var knownService))
                 {
-                    //because the service implementations are registered `AsSelf()`, we can resolve them directly
-                    return new Lease(scope.Resolve(knownService.ServiceImplementationType));
+                    //create a nested scope for the service lease. Halibut will automatically dispose of this lease at the end of the RPC call,
+                    //so pass the nested scope to the lease for disposal
+                    var nestedScope = scope.BeginLifetimeScope();
```
sburmanoctopus, line 42: Just wondering, what was the motivation behind creating a nested scope here instead of just resolving the service from `scope`?

APErebus:
#861 fixes the issue with the failing tests as the servers aren't registered in the expected scope

source/Octopus.Tentacle/Services/ServicesModule.cs
```diff
@@ -39,8 +39,18 @@ protected override void Load(ContainerBuilder builder)
                 .Select(x => new KnownService(x.ServiceImplementationType, x.ServiceAttribute!.ContractType))
                 .ToArray();
 
-            var assemblyServices = new KnownServiceSource(knownServices);
-            builder.RegisterInstance(assemblyServices).AsImplementedInterfaces();
+            var knownServiceSources = new KnownServiceSource(knownServices);
+            builder.RegisterInstance(knownServiceSources).AsImplementedInterfaces();
+
+            //register all halibut services with the root container
+            foreach (var knownServiceSource in knownServices)
```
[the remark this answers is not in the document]
APErebus: See #860

source/Octopus.Tentacle/Communications/AutofacServiceFactory.cs
APErebus: See #860

APErebus:
@sburmanoctopus I've updated the branch with autofac scope changes (#861), so the tests should pass now and is ready for a review :)

sburmanoctopus (approved):
LGTM

Issue #325, opened 2021-09-16
Halibut deserialization


andrew-at-octopus:
It was discovered that the halibut protocol used by Tentacle could be used to de-serialize a .NET type which did not form part of the defined service and client contracts.  This could be used for certain security exploits in .NET framework.  While our current versions all use .NET core and there are no known exploits, this security issue (CVE-2021-31819) needs to be addressed.

andrew-at-octopus:
This has been fixed in:
- Halibut 4.4.7
- Octopus Shared 10.3.3
- Octopus Server 2021.3.2035, 2021.2.7428, 2021.1.7737, 2020.6.5394
- Tentacle 6.1.938

octoreleasebot:
Release Note: Halibut protocol has been enhanced with a type whitelist and will no longer de-serialize any types not explicitly included in the communication interface.

PR #861, opened 2024-03-27
Change the Autofac scope Halibut services are registered in

APErebus:
# Background

Halibut services were being registered in a nested scope of the main lifetime scope. This caused an issue in #860 where they then couldn't be resolved in another part of the application due to the incorrect scoping.

# Results

This PR changes Halibut services to be registered in the root lifetime scope, however still maintains the nested scope for _each_ execution of an RPC call.

Because all halibut services are registered as `SingleInstance`, it means they are shared to all nested scopes. If there are services that will need to be per RPC call, we create a nested scope for each call to `CreateService`, which returns a `IServiceLease` that lives for the lifetime of the RPC call in halibut and is disposed at the end.

Ideally, each service wouldn't be registered as `SingleInstance` but Tentacle services are not stateless in that way (see `RunningScriptWrapper` etc), but this opens up the possibility

See https://github.com/OctopusDeploy/Halibut/blob/8c5546305ab73b81e1d92e9979dfcff121994efe/source/Halibut/ServiceModel/ServiceInvoker.cs#L26

This also solves the issue where we were performing an service-locator anti-pattern to resolve script services in the WorkspaceCleaner, as opposed to resolving a common interface etc.

# How to review this PR

I feel comfortable that the change regarding scopes isn't going to be breaking in any way and that automated testing will cover the scenarios.

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Tests/Communications/AutofacServiceFactoryFixture.cs
+++ b/source/Octopus.Tentacle.Tests/Communications/AutofacServiceFactoryFixture.cs
@@ -69,7 +69,7 @@ namespace Octopus.Tentacle.Tests.Communications
         }
 
         [Test]
-        public void Resolved_WithPerLifetimeScopeSingleSource_EachCreateServiceReturnsNewInstance()
+        public void Resolved_WithPerLifetimeScopeSingleSource_EachCreateServiceReturnsSameInstance()
         {
             var builder = new ContainerBuilder();
             builder.RegisterType<AutofacServiceFactory>().AsImplementedInterfaces().SingleInstance();
@@ -85,7 +85,7 @@ namespace Octopus.Tentacle.Tests.Communications
             var service1 = factory.CreateService(nameof(ISimpleService));
             var service2 = factory.CreateService(nameof(ISimpleService));
 
-            service1.Service.Should().NotBeSameAs(service2.Service);
+            service1.Service.Should().BeSameAs(service2.Service);
         }
 
         [Test]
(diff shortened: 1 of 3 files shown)


APErebus:
I had a discussion with @sburmanoctopus regarding the change and we both agree that there shouldn't be an issue removing the nested scope and resolving from the root scope.

I did some digging and was able to determine that the change in the scope was made in [this PR](https://github.com/OctopusDeploy/OctopusShared/pull/149) in 2021.

Based on Andrew W's comment [here](https://github.com/OctopusDeploy/OctopusShared/pull/149#discussion_r688182538) there doesn't appear to be a technical reason the nested scope was added, more a design choice to segregate the halibut services from the rest of Tentacle's DI. This restriction is now causing issues in #860

source/Octopus.Tentacle.Tests/Communications/AutofacServiceFactoryFixture.cs
```diff
@@ -66,12 +68,54 @@ public async Task Resolved_WithSingleSource_CanCreateAsyncServices()
             greeting.Should().Be("Hello Fred!");
         }
 
+        [Test]
+        public void Resolved_WithPerLifetimeScopeSingleSource_EachCreateServiceReturnsSameInstance()
```
sburmanoctopus, line 72: I noticed that the reason this test works this way is because we hold onto the scope within `AutofacServiceFactory`.

Like we discussed, if we want to make it respect `InstancePerLifetimeScope` and return a new instance, then we have to make `AutofacServiceFactory` create a scope when we call `CreateService`. 

But since all of these are singletons, it's not something we need to concern ourselves with now.

I just commented on this for future travellers 🙂

sburmanoctopus (approved):
LGTM

scme0 (approved):
LGTM

APErebus:
I spoke to @tothegills and he didn't remember the details of the design choices and was happy for the design to change as long as a [related CVE](https://github.com/OctopusDeploy/OctopusTentacle/issues/325) stays closed.

I investigated the CVE and it was a vulnerability  in the halibut protocol, so won't be affected by this change

PR #868, opened 2024-04-01
Update orphaned pod cleaner to use the "FinishedAt" property on the Pod

scme0:
# Background

Rather than creating our own date time, we can use the Finished At property on the Pod we get from the API.

### DIFF
--- a/source/Octopus.Tentacle.Tests/Kubernetes/KubernetesOrphanedPodCleanerTests.cs
+++ b/source/Octopus.Tentacle.Tests/Kubernetes/KubernetesOrphanedPodCleanerTests.cs
@@ -62,15 +62,17 @@ namespace Octopus.Tentacle.Tests.Kubernetes
             await podService.Received().Delete(scriptTicket, Arg.Any<CancellationToken>());
         }
 
-        [TestCase(PodState.Succeeded, true)]
-        [TestCase(PodState.Failed, true)]
-        [TestCase(PodState.Running, false)]
+        [TestCase("Succeeded", true)]
+        [TestCase("Failed", true)]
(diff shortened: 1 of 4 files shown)


source/Octopus.Tentacle.Tests/Kubernetes/KubernetesOrphanedPodCleanerTests.cs
```diff
@@ -72,28 +62,15 @@ public async Task OrphanedPodCleanedUpIfOver10MinutesHavePassed()
             await podService.Received().Delete(scriptTicket, Arg.Any<CancellationToken>());
         }
 
-        [TestCase("Succeeded", true)]
-        [TestCase("Failed", true)]
-        [TestCase("Running", false)]
+        [TestCase(PodState.Succeeded, true)]
```
flin-8, line 65: Add a case for `Pending` state?

scme0: The `PodState` enum only has three states:
```csharp
    public enum PodState
    {
        Running,
        Succeeded,
        Failed
    }
```

scme0: I see what you mean now - the Phase can have 5 different values.

scme0: I've added additional tests for those phases 👍

flin-8 (approved):
LGTM

PR #872, opened 2024-04-04
Add End-of-Stream (EOS) Marker to speed up Script Pod completion

flin-8:
[sc-71685]
# Background

Once a Pod has completed running its script, the Kubernetes API Server can take a few seconds to reflect the status update.

This PR adds an End-of-Stream (EOS) message to the bootstrapper script so we can detect this quicker than the Kubernetes API.

# Performance
For a deployment process with 7 steps:

- No EOS: **1m50s**
- With EOS: **1m43s**

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/docker/kubernetes-tentacle/bootstrapRunner/bootstrapRunner.go
+++ b/docker/kubernetes-tentacle/bootstrapRunner/bootstrapRunner.go
@@ -71,8 +71,7 @@ func main() {
 	Write("stdout", "Kubernetes Script Pod completed", &lineCounter)
 	Write("stdout", "##octopus[stdout-default]", &lineCounter)
 
-	//TODO: Add this back to speed things up
-	//Write("stdout", fmt.Sprintf("EOS-075CD4F0-8C76-491D-BA76-0879D35E9CFE<<>>%d", exitCode))
+	Write("stdout", fmt.Sprintf("EOS-075CD4F0-8C76-491D-BA76-0879D35E9CFE<<>>%d", exitCode), &lineCounter)
 
 	os.Exit(exitCode)
 }
--- a/docker/kubernetes-tentacle/bootstrapRunner/bootstrapRunner.go
+++ b/docker/kubernetes-tentacle/bootstrapRunner/bootstrapRunner.go
@@ -71,7 +71,7 @@ func main() {
 	Write("stdout", "Kubernetes Script Pod completed", &lineCounter)
 	Write("stdout", "##octopus[stdout-default]", &lineCounter)
 
-	Write("stdout", fmt.Sprintf("EOS-075CD4F0-8C76-491D-BA76-0879D35E9CFE<<>>%d", exitCode), &lineCounter)
+	Write("debug", fmt.Sprintf("EOS-075CD4F0-8C76-491D-BA76-0879D35E9CFE<<>>%d", exitCode), &lineCounter)
 
 	os.Exit(exitCode)
 }
--- a/source/Octopus.Tentacle/Kubernetes/PodLogLineParser.cs
+++ b/source/Octopus.Tentacle/Kubernetes/PodLogLineParser.cs
@@ -11,21 +11,29 @@ namespace Octopus.Tentacle.Kubernetes
         public PodLogLine? LogLine { get; }
         public string? Error { get; }
 
-        PodLogParseResult(bool succeeded, PodLogLine? logLine, string? error)
+        public int? ExitCode { get; }
+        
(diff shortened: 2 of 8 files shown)


source/Octopus.Tentacle/Kubernetes/PodLogLineParser.cs
```diff
@@ -1,68 +1,103 @@
...
 using Octopus.Tentacle.Contracts;
+using Octopus.Tentacle.Startup;
+using YamlDotNet.Core.Tokens;
 
 namespace Octopus.Tentacle.Kubernetes
 {
     public record PodLogLine(int LineNumber, ProcessOutputSource Source, string Message, DateTimeOffset Occurred);
 
-    public class PodLogParseResult
+    public abstract class PodLogLineParseResult
```
flin-8, line 10: Change this to subclasses so the properties don't need to be nullable

source/Octopus.Tentacle/Kubernetes/PodLogLineParser.cs
```diff
@@ -1,68 +1,103 @@
...
+                return new InvalidPodLogLineParseResult($"Invalid log line detected. '{logParts[2]}' is not a valid source.");
             }
             
             //add the new line
             var message = logParts[3];
 
-            //TODO: confirm if this line is useful
-            // if (message.StartsWith("EOS-075CD4F0-8C76-491D-BA76-0879D35E9CFE"))
-            //     source = ProcessOutputSource.Debug;
+            if (message.StartsWith("EOS-075CD4F0-8C76-491D-BA76-0879D35E9CFE"))
```
APErebus, line 87: Should/Could this be a named const so it's clear what this string represents?

docker/kubernetes-tentacle/bootstrapRunner/bootstrapRunner.go
```diff
@@ -71,8 +71,7 @@ func main() {
 	Write("stdout", "Kubernetes Script Pod completed", &lineCounter)
 	Write("stdout", "##octopus[stdout-default]", &lineCounter)
 
-	//TODO: Add this back to speed things up
-	//Write("stdout", fmt.Sprintf("EOS-075CD4F0-8C76-491D-BA76-0879D35E9CFE<<>>%d", exitCode))
+	Write("stdout", fmt.Sprintf("EOS-075CD4F0-8C76-491D-BA76-0879D35E9CFE<<>>%d", exitCode), &lineCounter)
```
APErebus, line 74: could this be `debug` rather than `stdout` for consistency (as it's explicitly `debug` in the parsing code)

source/Octopus.Tentacle/Kubernetes/PodLogReader.cs
```diff
@@ -9,41 +9,51 @@ namespace Octopus.Tentacle.Kubernetes
...
+                            haveSeenPodLogEntryMatchingLogSequence = true;
 
-                    results.Add(new ProcessOutput(podLogLine.Source, podLogLine.Message, podLogLine.Occurred));
-                    nextSequenceNumber = podLogLine.LineNumber;
-                }
+                        if (!haveSeenPodLogEntryMatchingLogSequence)
+                            throw new MissingPodLogException();
+
+                        if (validParseResult is EndOfStreamPodLogLineParseResult endOfStreamParseResult)
+                            exitCode = endOfStreamParseResult.ExitCode;
```
APErebus, line 44: if we are parsing the EOS message, should be just abort exit the loop here rather than waiting for the next read line? Should we read logs _after_ the EOS message?

APErebus (approved):
Awesome. Just some non-blocking questions

docker/kubernetes-tentacle/bootstrapRunner/bootstrapRunner.go
flin-8: Yeah good point - I didn't think of that 😁

source/Octopus.Tentacle/Kubernetes/PodLogReader.cs
flin-8: It's still possible for the Pod to error after this line if something goes wrong - I think we still want to show whatever comes after this message?

source/Octopus.Tentacle/Kubernetes/PodLogLineParser.cs
flin-8: Yes it should

PR #873, opened 2024-04-05
Update image to new script pod image

scme0:
For now I'm just using the artifactory image but we can publish to dockerhub and use that image once we've confirmed that it doesn't need any tweaks.

### DIFF
--- a/source/Octopus.Tentacle/Octopus.Tentacle.csproj
+++ b/source/Octopus.Tentacle/Octopus.Tentacle.csproj
@@ -52,7 +52,7 @@
 		<PackageReference Include="KubernetesClient" Version="10.1.19" />
 	</ItemGroup>
 	<ItemGroup>
-		<PackageReference Include="Octopus.Client" Version="14.3.1277" />
+		<PackageReference Include="Octopus.Client" Version="14.3.1329-scme-add-tentacle-details-resource-props" />
 	</ItemGroup>
 	<ItemGroup>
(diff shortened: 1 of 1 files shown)


source/Octopus.Tentacle/Kubernetes/KubernetesPodContainerRegistry.cs
```diff
@@ -36,7 +36,7 @@ public async Task<string> GetContainerImageForCluster()
 
             var tag = tagVersion?.ToString(2) ?? "latest";
 
-            return $"octopuslabs/k8s-workertools:{tag}";
+            return $"docker.packages.octopushq.com/octopusdeploy/k8s-agent-script-execution-image:{tag}";
```
APErebus, line 39: What testing have we done on this new image?

scme0: I've tested the health check, raw yaml and helm step so far. I was hoping we could get this in so we can test it with different setups that we (as developers) are using. Then we can fix things as they come up. I can probably spin up a quick kustomize test too.

APErebus: Yep, we need to make sure that this container is going to work for all k8s steps

As mentioned in Slack, I also don't like the name of this image

scme0:
I'll remake this PR

PR #874, opened 2024-04-05
Remove usage of ApplicationInstanceSelector from Tentacle Manager

acodrington:
# Background

We've had a bug turn up in Tentacle Manager [sc-74641], which occurs in the specific scenario where:
1. There are multiple instances of Tentacle installed, _and_
2. There is no default instance i.e. an instance named **(Default)**

This bug manifests as a nasty looking error in the UI which also prevents the user from being able to configure the Tentacle(s):

image

The bug is caused by some [changes to the configuration code in Tentacle itself](https://github.com/OctopusDeploy/OctopusTentacle/pull/674). Tentacle Manager uses these same classes to load configuration and display to the user, but Tentacle and Tentacle Manager do not operate in the same way and these changes are not compatible with Tentacle Manager.

# Results

The fix in this PR does the following:

1. Moves the configuration loading logic from `TentacleManagerModel` into a separate class `InstanceConfigurationLoader`
2. Implements a simplified version of `IApplicationInstanceSelector` which only contains the logic necessary to load configuration in Tentacle Manager. The existing implementation is unchanged and continues to be used in Tentacle itself. 
3. Refactors the relocated code to use the new simplified implementation of `IApplicationInstanceSelector`

I would like to refactor the Tentacle configuration code further in the future to make it simpler to use, but that is beyond the scope of this PR.

Fixes [sc-74641]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Manager.Tentacle/TentacleConfiguration/InstanceConfigurationLoader.cs
+++ b/source/Octopus.Manager.Tentacle/TentacleConfiguration/InstanceConfigurationLoader.cs
@@ -1,0 +1,153 @@
+﻿using System;
+using System.Linq;
+using Octopus.Diagnostics;
+using Octopus.Tentacle;
+using Octopus.Tentacle.Configuration;
+using Octopus.Tentacle.Configuration.Instances;
+using Octopus.Tentacle.Diagnostics;
+using Octopus.Tentacle.Util;
+
+namespace Octopus.Manager.Tentacle.TentacleConfiguration
+{
+    public interface IInstanceConfigurationLoader
+    {
+        InstanceConfiguration LoadConfiguration(ApplicationInstanceRecord applicationInstanceRecord);
+    }
+
+    public class InstanceConfiguration
+    {
+        public InstanceConfiguration(
+            IWritableKeyValueStore keyStore,
+            IHomeConfiguration homeConfiguration,
+            ILoggingConfiguration loggingConfiguration,
+            ITentacleConfiguration tentacleConfiguration)
+        {
+            KeyStore = keyStore;
+            HomeConfiguration = homeConfiguration;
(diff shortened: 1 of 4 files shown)


source/Octopus.Manager.Tentacle/TentacleConfiguration/InstanceConfigurationLoader.cs
```diff
@@ -0,0 +1,153 @@
...
+                this.instanceStrategies = instanceStrategies;
+                this.fileSystem = fileSystem;
+                this.log = log;
+                this.configFilePath = configFilePath;
+
+                try
+                {
+                    Current = LoadInstance();
+                    canLoadCurrentInstance = true;
+
```
rhysparry, line 100: nit: blank line before }

source/Octopus.Manager.Tentacle/TentacleConfiguration/InstanceConfigurationLoader.cs
```diff
@@ -0,0 +1,153 @@
...
+                EnsureConfigurationExists(configurationPath);
+
+                log.Verbose($"Loading configuration from {configurationPath}");
+                var writable = new XmlFileKeyValueStore(fileSystem, configurationPath);
+                return (ContributeAdditionalConfiguration(writable), writable);
+            }
+
+            void EnsureConfigurationExists(string configurationPath)
+            {
+                if (!fileSystem.FileExists(configurationPath ?? string.Empty))
```
rhysparry, line 129: thought: can we enable nullable checks?

rhysparry (approved):
LGTM

acodrington:
Fixes https://github.com/OctopusDeploy/OctopusTentacle/issues/875

PR #877, opened 2024-04-08
Only pull Script Pod logs newer than the last set we got

flin-8:
[sc-71685]
# Background

We currently stream all logs from the Script Pod every time. There's a `sinceTime` filter we can use to only get the latest logs.

This PR adds `sinceTime` to the request we send. We also depend on the API Server's timestamp now instead of writing our own in the bootstrapper - in case the clocks are out of sync.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/docker/kubernetes-tentacle/bootstrapRunner/bootstrapRunner.go
+++ b/docker/kubernetes-tentacle/bootstrapRunner/bootstrapRunner.go
@@ -7,7 +7,6 @@ import (
 	"os"
 	"os/exec"
 	"sync"
-	"time"
 )
 
 type SafeCounter struct {
@@ -88,7 +87,7 @@ func Write(stream string, text string, counter *SafeCounter) {
 	//https://go.dev/tour/concurrency/9
 	counter.Mutex.Lock()
 
-	fmt.Printf("%d|%s|%s|%s\n", counter.Value, time.Now().UTC().Format(time.RFC3339Nano), stream, text)
+	fmt.Printf("|%d|%s|%s\n", counter.Value, stream, text)
 	counter.Value++
 
 	counter.Mutex.Unlock()
--- a/source/Octopus.Tentacle/Kubernetes/KubernetesClientExtensionMethods.cs
+++ b/source/Octopus.Tentacle/Kubernetes/KubernetesClientExtensionMethods.cs
@@ -10,7 +10,7 @@ namespace Octopus.Tentacle.Kubernetes
 {
     public static class KubernetesClientExtensionMethods
     {
-        public static async Task<Stream> GetNamespacedPodLogsAsync(this k8s.Kubernetes client, ISystemLog log, string podName, string namespaceParameter, string container, DateTimeOffset? sinceTime, CancellationToken cancellationToken = default)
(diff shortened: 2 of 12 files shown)


docker/kubernetes-tentacle/bootstrapRunner/bootstrapRunner.go
```diff
@@ -88,7 +87,7 @@ func Write(stream string, text string, counter *SafeCounter) {
 	//https://go.dev/tour/concurrency/9
 	counter.Mutex.Lock()
 
-	fmt.Printf("%d|%s|%s|%s\n", counter.Value, time.Now().UTC().Format(time.RFC3339Nano), stream, text)
+	fmt.Printf("|%d|%s|%s\n", counter.Value, stream, text)
```
flin-8, line 90: Remove our own timestamp - rely on the API Server's timestamp instead (it gets added at the front with a space character

source/Octopus.Tentacle/Kubernetes/ScriptPodSinceTimeStore.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace Octopus.Tentacle.Kubernetes
+{
+    public interface IScriptPodSinceTimeStore
+    {
+        DateTimeOffset? GetSinceTime(ScriptTicket scriptTicket);
+        void UpdateSinceTime(ScriptTicket scriptTicket, DateTimeOffset nextSinceTime);
+        void Delete(ScriptTicket scriptTicket);
+    }
+
+    public class ScriptPodSinceTimeStore : IScriptPodSinceTimeStore
```
flin-8, line 14: I decided to not put this on the tracked pods, this means the monitoring is not coupled to the log reading

docker/kubernetes-tentacle/bootstrapRunner/bootstrapRunner.go
APErebus: is it in the correct format/resolution?

source/Octopus.Tentacle/Kubernetes/KubernetesClientExtensionMethods.cs
```diff
@@ -4,21 +4,23 @@
...
+            //Include Server timestamps so we can use it to specify "sinceTime" on the next call
+            var url = $"api/v1/namespaces/{namespaceParameter}/pods/{podName}/log?container={container}&timestamps=true";
 
             if (sinceTime is not null)
             {
                 var sinceTimeStr = sinceTime.Value.ToString("O");
                 url += $"&sinceTime={Uri.EscapeDataString(sinceTimeStr)}";
             }
-
+
```
scme0, line 23: Nit: Remove whitespace please :)

docker/kubernetes-tentacle/bootstrapRunner/bootstrapRunner.go
flin-8: Good question - looks like it - I have a temp build that prints "Parsing line: ..."

CleanShot 2024-04-10 at 17 22 44@2x

APErebus (approved):
Awesome stuff 👍

PR #879, opened 2024-04-09
Add ability for Tentacle to write to the task log

flin-8:
[sc-71685]
# Background

Now that we pull Script Pod logs directly from the K8s API, we need another way for Tentacle to add messages to the task log (previously this was all on the file system).

This PR adds an in-memory log that gets merged with the Script Pod logs. There's a catch that you'll lose the in-memory log messages if Tentacle is restarted (but these are verbose debugging messages, so they shouldn't affect deployment)

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Tests/Kubernetes/KubernetesOrphanedPodCleanerTests.cs
+++ b/source/Octopus.Tentacle.Tests/Kubernetes/KubernetesOrphanedPodCleanerTests.cs
@@ -34,7 +34,7 @@ namespace Octopus.Tentacle.Tests.Kubernetes
             podService = Substitute.For<IKubernetesPodService>();
             log = new InMemoryLog();
             clock = new FixedClock(startTime);
-            monitor = new KubernetesPodMonitor(podService, log);
+            monitor = new KubernetesPodMonitor(podService, log, new TentacleScriptLogProvider());
 
             scriptTicket = new ScriptTicket(Guid.NewGuid().ToString());
 
--- a/source/Octopus.Tentacle.Tests/Kubernetes/KubernetesOrphanedPodCleanerTests.cs
+++ b/source/Octopus.Tentacle.Tests/Kubernetes/KubernetesOrphanedPodCleanerTests.cs
@@ -26,6 +26,7 @@ namespace Octopus.Tentacle.Tests.Kubernetes
         TimeSpan overCutoff;
         TimeSpan underCutoff;
         DateTimeOffset startTime;
+        ITentacleScriptLogProvider scriptLogProvider;
 
         [SetUp]
         public void Setup()
@@ -34,14 +35,22 @@ namespace Octopus.Tentacle.Tests.Kubernetes
             podService = Substitute.For<IKubernetesPodService>();
             log = new InMemoryLog();
             clock = new FixedClock(startTime);
-            monitor = new KubernetesPodMonitor(podService, log, new TentacleScriptLogProvider());
+            scriptLogProvider = Substitute.For<ITentacleScriptLogProvider>();
+            monitor = new KubernetesPodMonitor(podService, log, scriptLogProvider);
 
             scriptTicket = new ScriptTicket(Guid.NewGuid().ToString());
 
-            cleaner = new KubernetesOrphanedPodCleaner(monitor, podService, log, clock);
+            cleaner = new KubernetesOrphanedPodCleaner(monitor, podService, log, clock, scriptLogProvider);
 
             overCutoff = cleaner.CompletedPodConsideredOrphanedAfterTimeSpan + 1.Minutes();
             underCutoff = cleaner.CompletedPodConsideredOrphanedAfterTimeSpan - 1.Minutes();
+        }
+
+        [TearDown]
+        public void TearDown()
+        {
+            Environment.SetEnvironmentVariable("OCTOPUS__K8STENTACLE__DISABLEAUTOPODCLEANUP", null);
+            Environment.SetEnvironmentVariable("OCTOPUS__K8STENTACLE__PODSCONSIDEREDORPHANEDAFTERMINUTES", null);
         }
 
         [Test]
@@ -60,6 +69,7 @@ namespace Octopus.Tentacle.Tests.Kubernetes
 
             //Assert
             await podService.Received().Delete(scriptTicket, Arg.Any<CancellationToken>());
+            scriptLogProvider.Received().Delete(scriptTicket);
         }
 
         [TestCase("Succeeded", true)]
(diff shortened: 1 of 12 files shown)


source/Octopus.Tentacle/Kubernetes/InMemoryTentacleScriptLog.cs
```diff
@@ -0,0 +1,40 @@
...
+                logs.Add(new ProcessOutput(ProcessOutputSource.StdOut, message));
+        }
+
+        public void Error(string message)
+        {
+            lock (logs)
+                logs.Add(new ProcessOutput(ProcessOutputSource.StdErr, message));
+        }
+
+        public IReadOnlyCollection<ProcessOutput> PopLogs()
```
flin-8, line 30: Are we happy to pop the logs like this? We'll lose these logs if there's an exception later on

source/Octopus.Tentacle/Kubernetes/PodLogReader.cs
```diff
@@ -27,33 +27,41 @@ public static async Task<(IReadOnlyCollection<ProcessOutput> Lines, long NextSeq
 
                 var parseResult = PodLogLineParser.ParseLine(line!);
 
-                if (parseResult is ValidPodLogLineParseResult validParseResult)
+                switch (parseResult)
```
flin-8, line 30: Changed this to a `switch`, resulting in a bad diff 😢

source/Octopus.Tentacle/Kubernetes/PodLogReader.cs
```diff
@@ -27,33 +27,41 @@ public static async Task<(IReadOnlyCollection<ProcessOutput> Lines, long NextSeq
...
-                        if (podLogLine.LineNumber == lastLogSequence + 1)
-                            haveSeenPodLogEntryMatchingLogSequence = true;
+                        var podLogLine = validParseResult.LogLine;
 
-                        if (!haveSeenPodLogEntryMatchingLogSequence)
-                            throw new MissingPodLogException();
+                        //Pod log line numbers are 1-based, log sequence is 0-based
+                        if (podLogLine.LineNumber > lastLogSequence)
+                        {
+                            if (podLogLine.LineNumber == lastLogSequence + 1)
```
scme0, line 39: Can we just combine this if with the next one rather than setting the bool?

source/Octopus.Tentacle/Kubernetes/IKubernetesPodLogService.cs
```diff
@@ -61,6 +69,11 @@ async Task<(IReadOnlyCollection<ProcessOutput> Outputs, long NextSequenceNumber,
                     return await PodLogReader.ReadPodLogs(lastLogSequence, reader);
                 }
             }
+            
+            IReadOnlyCollection<ProcessOutput> ReadTentacleLogs()
```
APErebus, line 73: Not sure this method is serving a purpose

source/Octopus.Tentacle/Kubernetes/InMemoryTentacleScriptLog.cs
```diff
@@ -0,0 +1,40 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Octopus.Tentacle.Contracts;
+
+namespace Octopus.Tentacle.Kubernetes
+{
+    public class InMemoryTentacleScriptLog
+    {
+        readonly List<ProcessOutput> logs = new();
```
APErebus, line 10: Without being flip, is this not the appropriate place for a `ConcurrentDictionary`? It will handle all this locking for you so you don't need manual lock statements everywhere?

source/Octopus.Tentacle/Kubernetes/InMemoryTentacleScriptLog.cs
APErebus: I'm comfortable with this

source/Octopus.Tentacle/Kubernetes/TentacleScriptLogProvider.cs
```diff
@@ -0,0 +1,24 @@
+using System;
+using System.Collections.Concurrent;
+using Octopus.Tentacle.Contracts;
+
+namespace Octopus.Tentacle.Kubernetes
+{
+    public class TentacleScriptLogProvider
+    {
+        readonly ConcurrentDictionary<ScriptTicket, InMemoryTentacleScriptLog> logs = new();
```
APErebus, line 9: 😍

source/Octopus.Tentacle/Kubernetes/TentacleScriptLogProvider.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        //Tentacle might restart after a script has started, so we always have to create this logger on demand
+        public InMemoryTentacleScriptLog GetOrCreate(ScriptTicket scriptTicket)
+        {
+            return logs.GetOrAdd(scriptTicket, _ => new InMemoryTentacleScriptLog());
+        }
+
+        //Having this clean up should reduce significant memory leaks
+        //Assuming there's always a chance that GetOrCreate() might be called after Delete()
+        public void Delete(ScriptTicket scriptTicket)
```
APErebus, line 19: Should we also include this in the `OrphanedPodLogCleaner` monitor? Just in case the `CompleteScript` isn't called?

APErebus (approved):
Excellent stuff! Just a few questions, but nothing blocking :)

source/Octopus.Tentacle/Kubernetes/IKubernetesPodLogService.cs
flin-8: It used to...until it got that small 😬

source/Octopus.Tentacle/Kubernetes/PodLogReader.cs
scme0: Ah yep, I see now it's declared before the while loop.

source/Octopus.Tentacle/Kubernetes/InMemoryTentacleScriptLog.cs
flin-8: We want to preserve the order of log messages right? Or are you saying this can be unordered since we merge this with the Pod logs anyway?

source/Octopus.Tentacle/Kubernetes/TentacleScriptLogProvider.cs
flin-8: Can't hurt

source/Octopus.Tentacle.Tests/Kubernetes/KubernetesOrphanedPodCleanerTests.cs
```diff
@@ -126,9 +139,7 @@ public async Task OrphanedPodNotCleanedUpIfPodCleanupIsDisabled()
 
             //Assert
             await podService.DidNotReceive().Delete(scriptTicket, Arg.Any<CancellationToken>());
-
-            //Cleanup
-            Environment.SetEnvironmentVariable("OCTOPUS__K8STENTACLE__DISABLEAUTOPODCLEANUP", null);
```
flin-8, line 131: Cleanup should be done in Teardown in case of test failures

source/Octopus.Tentacle/Kubernetes/InMemoryTentacleScriptLog.cs
flin-8: I think we need a lock for `PopLogs()` anyway - otherwise we might clear the list when someone is adding

PR #881, opened 2024-04-10
Support opening multiple polling connections

APErebus:
# Background

Now we have [safeguards](https://github.com/OctopusDeploy/Halibut/pull/599) in Halibut to stop DoS-ing a server, we can add a new environment value to open multiple polling connections. This improves throughput and performance by having multiple connections polling for work.

# Results

The env var name is `TentaclePollingConnectionCount`.

I also considered adding a hard max per tentacle (say 16 or something), but felt like people setting this env var should understand what that means to the target server. As mentioned, we also have the new safety nets in Octopus Server to handle this.

Will be tested via the k8s agent against a branch instance with the Server limit set

Shortcut story: [sc-74991]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle/Communications/HalibutInitializer.cs
+++ b/source/Octopus.Tentacle/Communications/HalibutInitializer.cs
@@ -83,6 +83,8 @@ namespace Octopus.Tentacle.Communications
             }
         }
 
+        const int MaximumPollingConnectionCount = 8;
+
         void AddPollingEndpoints()
         {
             foreach (var pollingEndPoint in GetOctopusServersToPoll())
@@ -104,12 +106,27 @@ namespace Octopus.Tentacle.Communications
                 var serviceEndPoint = new ServiceEndPoint(pollingEndPoint.Address, pollingEndPoint.Thumbprint, halibutProxy, halibutTimeoutsAndLimits);
 
                 //Open multiple polling connections if the env var is set to a non-zero/negative number
-                var connectionCount = int.TryParse(Environment.GetEnvironmentVariable(EnvironmentVariables.TentaclePollingConnectionCount), out var count)
-                    ? count
-                    : 1;
+                var connectionCount = 1u;
+                if (uint.TryParse(Environment.GetEnvironmentVariable(EnvironmentVariables.TentaclePollingConnectionCount), out var count))
+                {
+                    log.InfoFormat("Requested polling connection count: {0}", count);
+                    connectionCount = count;
+                }
 
-                // negative numbers and 0 are coerced to 1
-                connectionCount = Math.Max(connectionCount, 1);
+                //Coerce the requested value as it might be outside our max & min
+                switch (connectionCount)
+                {
+                    case > MaximumPollingConnectionCount:
+                        log.InfoFormat("Maximum polling connection count: {0}", MaximumPollingConnectionCount);
+                        connectionCount = 8;
+                        break;
+                    case 0:
+                        log.InfoFormat("Minimum polling connection count: {0}", 1);
+                        connectionCount = 1;
+                        break;
+                }
+
+                log.InfoFormat("Opening {0} polling connections", connectionCount);
 
                 for (var i = 0; i < connectionCount; i++)
                 {
--- a/source/Octopus.Tentacle/Communications/HalibutInitializer.cs
+++ b/source/Octopus.Tentacle/Communications/HalibutInitializer.cs
@@ -105,34 +105,40 @@ namespace Octopus.Tentacle.Communications
                 var halibutTimeoutsAndLimits = new HalibutTimeoutsAndLimits();
                 var serviceEndPoint = new ServiceEndPoint(pollingEndPoint.Address, pollingEndPoint.Thumbprint, halibutProxy, halibutTimeoutsAndLimits);
 
-                //Open multiple polling connections if the env var is set to a non-zero/negative number
-                var connectionCount = 1u;
-                if (uint.TryParse(Environment.GetEnvironmentVariable(EnvironmentVariables.TentaclePollingConnectionCount), out var count))
-                {
-                    log.InfoFormat("Requested polling connection count: {0}", count);
-                    connectionCount = count;
-                }
-
-                //Coerce the requested value as it might be outside our max & min
-                switch (connectionCount)
-                {
-                    case > MaximumPollingConnectionCount:
-                        log.InfoFormat("Maximum polling connection count: {0}", MaximumPollingConnectionCount);
-                        connectionCount = 8;
-                        break;
-                    case 0:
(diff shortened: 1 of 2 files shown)


source/Octopus.Tentacle/Communications/HalibutInitializer.cs
```diff
@@ -100,7 +102,19 @@ void AddPollingEndpoints()
...
                 var serviceEndPoint = new ServiceEndPoint(pollingEndPoint.Address, pollingEndPoint.Thumbprint, halibutProxy, halibutTimeoutsAndLimits);
-                halibut.Poll(new Uri(pollingEndPoint.SubscriptionId), serviceEndPoint, CancellationToken.None);
+
+                //Open multiple polling connections if the env var is set to a non-zero/negative number
+                var connectionCount = int.TryParse(Environment.GetEnvironmentVariable(EnvironmentVariables.TentaclePollingConnectionCount), out var count)
+                    ? count
+                    : 1;
+
+                // negative numbers and 0 are coerced to 1
+                connectionCount = Math.Max(connectionCount, 1);
```
pawelpabich, line 112: It might be worth adding a hard limit here, e.g. 8 or 16 connections.  Just trying to make sure that people won't shoot themselves in the foot. We can always relax this limit  if there is a customer with a reasonable use case. 

Also, it would be worth logging what the effective `connectionCount`  is, especially if we apply a hard limit.

source/Octopus.Tentacle/Communications/HalibutInitializer.cs
```diff
@@ -100,7 +102,19 @@ void AddPollingEndpoints()
...
                 var halibutTimeoutsAndLimits = new HalibutTimeoutsAndLimits();
                 var serviceEndPoint = new ServiceEndPoint(pollingEndPoint.Address, pollingEndPoint.Thumbprint, halibutProxy, halibutTimeoutsAndLimits);
-                halibut.Poll(new Uri(pollingEndPoint.SubscriptionId), serviceEndPoint, CancellationToken.None);
+
+                //Open multiple polling connections if the env var is set to a non-zero/negative number
+                var connectionCount = int.TryParse(Environment.GetEnvironmentVariable(EnvironmentVariables.TentaclePollingConnectionCount), out var count)
+                    ? count
+                    : 1;
+
+                // negative numbers and 0 are coerced to 1
```
pawelpabich, line 111: Nitpick, Would using uint32 help eliminate the need to check for negative values?

pawelpabich (commented):
LGTM

source/Octopus.Tentacle/Communications/HalibutInitializer.cs
APErebus: yeah, I did think about that. I wasn't sure. Happy to set it to `8` for now and log it out 👍

source/Octopus.Tentacle/Communications/HalibutInitializer.cs
APErebus: probably yes. It'd still need to check for 0, so not sure it'll gain much?

source/Octopus.Tentacle/Communications/HalibutInitializer.cs
APErebus: @pawelpabich Implemented a hard limit of 8 and also log the requested and resulting value (after validation/coercion) in [f251b2b](https://github.com/OctopusDeploy/OctopusTentacle/pull/881/commits/f251b2b48defb7687450d602d93e9641a73022e5)

source/Octopus.Tentacle/Communications/HalibutInitializer.cs
APErebus: Changed to `uint`

source/Octopus.Tentacle/Communications/HalibutInitializer.cs
```diff
@@ -100,7 +104,34 @@ void AddPollingEndpoints()
...
+                {
+                    case > MaximumPollingConnectionCount:
+                        log.InfoFormat("Maximum polling connection count: {0}", MaximumPollingConnectionCount);
+                        connectionCount = 8;
+                        break;
+                    case 0:
+                        log.InfoFormat("Minimum polling connection count: {0}", 1);
+                        connectionCount = 1;
+                        break;
+                }
```
scme0, line 127: Nit: it might be worth breaking this code out into it's own method.

scme0 (approved):
LGTM - just one nit

source/Octopus.Tentacle/Communications/HalibutInitializer.cs
APErebus: Done in [025f3ff](https://github.com/OctopusDeploy/OctopusTentacle/pull/881/commits/025f3ffed6571f06587c2cac2e8956c41e737aff)

source/Octopus.Tentacle/Communications/HalibutInitializer.cs
```diff
@@ -100,8 +104,41 @@ void AddPollingEndpoints()
...
+            {
+                log.InfoFormat("Requested polling connection count: {0}", count);
+                connectionCount = count;
             }
+
+            //Coerce the requested value as it might be outside our max & min
+            switch (connectionCount)
+            {
+                case > MaximumPollingConnectionCount:
+                    log.InfoFormat("Maximum polling connection count: {0}", MaximumPollingConnectionCount);
```
flin-8, line 131: Probably worth saying something like "The requested polling connection count is more than the maximum of 8, setting to 8 instead" etc.

Or just always print "Maximum allowed polling connection count" and the min allowed

flin-8 (approved):
😀

source/Octopus.Tentacle/Communications/HalibutInitializer.cs
```diff
@@ -104,12 +106,27 @@ void AddPollingEndpoints()
...
+                    log.InfoFormat("Requested polling connection count: {0}", count);
+                    connectionCount = count;
+                }
+
+                //Coerce the requested value as it might be outside our max & min
+                switch (connectionCount)
+                {
+                    case > MaximumPollingConnectionCount:
+                        log.InfoFormat("Maximum polling connection count: {0}", MaximumPollingConnectionCount);
+                        connectionCount = 8;
```
pawelpabich: Nit: Just set it to `MaximumPollingConnectionCount` so the change of the limit in the future can be done just in one place.

APErebus: Derp. That's what it was meant to be. I'll fix it in a following PR 🤦

APErebus: Fixing in #882

PR #886, opened 2024-04-15
Fix bug where Deployment hangs if the Script Pod is deleted

flin-8:
[sc-75597]
# Background

This PR makes `KubernetesScriptServiceV1Alpha.GetStatusAsync()` throw if the Script Pod isn't found in the in-memory lookup, this stops the deployment from hanging if the Script Pod isn't found. 

As a consequence, a new in-memory dictionary has been added to ensure that newly created Pods aren't going to be missing.

Also a manual reset event has been added to ensure the script service still returns the correct results during Tentacle startup by blocking until the Script Pods have been loaded from the Kubernetes API.

There is a _small_ chance of us throwing an exception if:
- A Script Pod is newly created
- Tentacle restarts
- The newly Script Pod isn't returned by the Kubernetes API (hence it won't be in the lookup)
- Server calls `GetStatus()`

@liam-mackie and I think the chances of this happening is rare, so not worth worrying about (one solution would be to change the service script contract to allow an "I don't know" answer, and let Server retry and fail after a number of retries.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Octopus.Tentacle/Kubernetes/KubernetesPodMonitor.cs
```diff
@@ -85,7 +103,7 @@ async Task UpdateLoop(CancellationToken cancellationToken)
 
         internal async Task<string> InitialLoadAsync(CancellationToken cancellationToken)
         {
-            log.Verbose("Preloading pod statuses");
+            log.Verbose("Loading pod statuses");
```
APErebus, line 107: Are we missing an `initialLoadLock.Reset()` here? won't the next time the initial load kicks in, the lock will already be set and thus won't block anything?

flin-8: I initially had a reset - but I think we are ok on subsequent loads - I created another dictionary to hold the pending Pods, so `podStatusLookup` is only responsible for holding what comes from the K8S API

source/Octopus.Tentacle/Kubernetes/KubernetesPodMonitor.cs
```diff
@@ -156,11 +179,36 @@ internal async Task OnNewEvent(WatchEventType type, V1Pod pod, CancellationToken
...
-            podStatusLookup.Values.Cast<ITrackedScriptPod>().ToList();
+        IList<ITrackedScriptPod> IKubernetesPodStatusProvider.GetAllTrackedScriptPods()
+        {
+            WaitForInitialLoadToFinish();
+
+            return podStatusLookup
+                .Concat(pendingPodStatusLookup)
+                .ToLookup(p => p.Key)
+                .Select(g => (ITrackedScriptPod)g.First().Value) //Deduplicate
+                .ToList();
```
APErebus, line 190: Rather than having dupes, should we remove the pod out of pending once it's been added via the API/Watch? Then we should never have duplicates?

source/Octopus.Tentacle/Services/Scripts/Kubernetes/KubernetesScriptServiceV1Alpha.cs
```diff
@@ -86,8 +86,7 @@ public async Task<KubernetesScriptStatusResponseV1Alpha> GetStatusAsync(Kubernet
             var trackedPod = podStatusProvider.TryGetTrackedScriptPod(request.ScriptTicket);
             return trackedPod != null
                 ? await GetResponse(trackedPod, request.LastLogSequence, cancellationToken)
-                //if we are getting the status of an unknown pod, return that it's still pending
-                : new KubernetesScriptStatusResponseV1Alpha(request.ScriptTicket, ProcessState.Pending, 0, new List<ProcessOutput>(), request.LastLogSequence);
+                : throw new Exception("Can't find Script Pod for: " + request.ScriptTicket);
```
APErebus, line 89: Yeah, I think this is ok 👍 

Can we just improve the error message, and maybe make a custom exception type as this will be reported back to the task log in Server, so might be nice to have a slightly friendlier message?

source/Octopus.Tentacle/Kubernetes/KubernetesPodMonitor.cs
flin-8: There's also the batch load that happens every 10min - which we would can do - just most likely needs a `lock` statement to ensure we sync the 2 dictionaries

source/Octopus.Tentacle/Kubernetes/KubernetesPodMonitor.cs
APErebus: and I guess we are doing an atomic swap of that dictionary 👍

source/Octopus.Tentacle/Services/Scripts/Kubernetes/KubernetesScriptServiceV1Alpha.cs
flin-8: CleanShot 2024-04-15 at 16 30 55@2x

APErebus: like it!

source/Octopus.Tentacle/Kubernetes/KubernetesPodMonitor.cs
```diff
@@ -85,82 +106,133 @@ async Task UpdateLoop(CancellationToken cancellationToken)
...
+                   status = new TrackedScriptPod(scriptTicket);
+               }
+
+               status.Update(pod);
+
+               log.Verbose($"Loaded pod {pod.Name()} ({status})");
+               newStatuses[scriptTicket] = status;
+           }
+
+           lock (theLock)
```
flin-8, line 129: I think we actually need to use a lock, since we don't want to be modifying the collections while the other methods are in the middle of reading

flin-8: @APErebus happy to talk through this when you are free

PR #887, opened 2024-04-15
Move common test code to CommonTestUtils

APErebus:
# Background

In preparation for Kubernetes Tentacle/Agent integration tests (WIP PR #870), there is some code that is shared between the two Integration test projects.

# Results

This PR moves those types into the `CommonTestUtils` project so the following PR is smaller and easier to understand (just adding the new tests)

Shortcut story: [sc-66281]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Octopus.Tentacle.CommonTestUtils/TestCertificates.cs
```diff
@@ -2,9 +2,9 @@
 using System.IO;
 using System.Security.Cryptography.X509Certificates;
 
-namespace Octopus.Tentacle.Tests.Integration.Support
+namespace Octopus.Tentacle.CommonTestUtils
 {
-    internal class Certificates
+    public class TestCertificates
```
APErebus, line 7: This was clashing with the folder name, so just renamed for clarity

source/Octopus.Tentacle.CommonTestUtils/TemporaryDirectory.cs
```diff
@@ -27,7 +28,7 @@ string GetTempBasePath()
             var path = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData, Environment.SpecialFolderOption.DoNotVerify);
             fileSystem.EnsureDirectoryExists(path);
 
-            path = Path.Combine(path, Assembly.GetEntryAssembly() != null ? Assembly.GetEntryAssembly()!.GetName().Name! : "Octopus");
+            path = Path.Combine(path, Assembly.GetEntryAssembly()?.GetName()?.Name ?? "Octopus");
```
flin-8, line 31: were you getting null ref's before?

APErebus: no, but I think this is a more modern way to handle this statement

flin-8: So the change of behaviour is intended here? Now it will use `Octopus` if anything in the chain is `null`

APErebus: Let me check the nullability annotations, but I think this is a safe change

Worth mentioning that this is just for creating a temp directory, but let me double check

APErebus: I double checked and I probably just have some extra `?` in the chain I don't actually need. The nullability check on the `GetEntryAssembly()` is required, but from then on, `GetName()` will return a non-null type and then `Name` is nullable. 

There is a slight behavioural difference in this case as previously, if you had an entry assembly, but `.Name` returned null, it would blow up due to the `!`.

The change now will handle this case.

flin-8 (approved):
LGTM

raynhamdev (approved):
Changes LGTM 👍

PR #889, opened 2024-04-16
Add NFS watchdog to script pods

liam-mackie:
# Background
When NFS goes down, we can get corrupted file shares, which are impossible to read from, but can be written to (but not persisted over the network). This causes all sorts of issues, so we've decided that deleting the pod (which also deletes the mount) is the safest way to recover from this error, so that we don't expect that we've completed work which hasn't persisted. 

This led to the creation of the Kubernetes Agent NFS Watchdog, which checks that there are no read errors coming from the agent. It's designed to run as a lightweight sidecar, and will delete the pod it's running in if there are read errors.

# Results

This PR adds the NFS watchdog to all script containers spawned by the tentacle agent. 

If NFS _or_ the watchdog itself is not in use, the environment variables that we use to determine the image don't get set, and `CreateWatchdogContainer()` returns null. The `null` container can't be added to the containerList because there's no `null` handling in the library itself, so I made a quick extension method to only add an item to a list if it's not null.

I've manually tested this extensively. The only time this breaks is if the environment variable is overwritten, but the actual watchdog is not.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


scme0 (approved):
Nice! LGTM

PR #891, opened 2024-04-16
Update ScriptPod default image

scme0:
# Background

#project-k8s-agent

Upgrade the default script pod image to use a slimmed down image that only has the required dependencies.

[sc-70343]

### DIFF
--- a/source/Octopus.Tentacle/Kubernetes/KubernetesPodContainerRegistry.cs
+++ b/source/Octopus.Tentacle/Kubernetes/KubernetesPodContainerRegistry.cs
@@ -36,7 +36,7 @@ namespace Octopus.Tentacle.Kubernetes
 
             var tag = tagVersion?.ToString(2) ?? "latest";
 
-            return $"octopuslabs/k8s-workertools:{tag}";
+            return $"docker.packages.octopushq.com/octopusdeploy/kubernetes-agent-tools-base:{tag}";
         }
     }
 }


source/Octopus.Tentacle/Kubernetes/KubernetesPodContainerRegistry.cs
```diff
@@ -36,7 +36,7 @@ public async Task<string> GetContainerImageForCluster()
 
             var tag = tagVersion?.ToString(2) ?? "latest";
 
-            return $"octopuslabs/k8s-workertools:{tag}";
+            return $"docker.packages.octopushq.com/octopusdeploy/kubernetes-agent-tools-base:{tag}";
```
APErebus, line 39: Are we pushing this image to dockerhub yet? Or are we going to wait until we've tested it more?

scme0: I was thinking we'd wait until we've tested it and then just before we release we can push it to dockerhub.

APErebus: yep, I think that's a fair call. Can you put a ticket in Shortcut so we don't forget to do this?

scme0:
New shortcut story to review this image before we release the agent EAP: https://app.shortcut.com/octopusdeploy/story/76030/review-the-new-script-pod-container-image

PR #892, opened 2024-04-17
Sanitize Kubernetes cluster version

tleed5:
# Background
#project-k8s-agent

EKS returns the cluster version as, where the minor version number has a non-numeric character which causes the cluster version parsing to fail with a `System.FormatException`
```
{
"major": "1",
"minor": "29+",
"gitVersion": "v1.29.1-eks-b9c9ed7",
"gitCommit": "07600c74de018baffb16c82771a48adcb843a932",
"gitTreeState": "clean",
"buildDate": "2024-03-02T03:46:35Z",
"goVersion": "go1.21.6",
"compiler": "gc",
"platform": "linux/amd64"
}
```

[SC-75848]

# Results
- Added `KubernetesVersionParser` class to handle the parsing of version numbers related to K8s and refactored `KubernetesClusterService` to use it.
- Moved `ClusterVersion` into `KubernetesVersionParser`
- Added tests for `KubernetesVersionParser` 

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle/Kubernetes/KubernetesVersionParser.cs
+++ b/source/Octopus.Tentacle/Kubernetes/KubernetesVersionParser.cs
@@ -6,6 +6,8 @@ namespace Octopus.Tentacle.Kubernetes
 {
     public static class KubernetesVersionParser
     {
+        static Regex clusterVersionRegex = new Regex("[^0-9]");
+        
         public static ClusterVersion ParseClusterVersion(VersionInfo versionInfo)
         {
             return new ClusterVersion(SanitizeAndParseVersionNumber(versionInfo.Major), SanitizeAndParseVersionNumber(versionInfo.Minor));
@@ -13,11 +15,7 @@ namespace Octopus.Tentacle.Kubernetes
         
         static int SanitizeAndParseVersionNumber(string version)
         {
-            if (int.TryParse(version, out var result))
-                return result;
-            
             var sanitized = Regex.Replace(version, "[^0-9]", "");
-
             return int.Parse(sanitized);
         }
 
--- a/source/Octopus.Tentacle/Kubernetes/KubernetesVersionParser.cs
+++ b/source/Octopus.Tentacle/Kubernetes/KubernetesVersionParser.cs
@@ -6,8 +6,6 @@ namespace Octopus.Tentacle.Kubernetes
 {
     public static class KubernetesVersionParser
     {
-        static Regex clusterVersionRegex = new Regex("[^0-9]");
-        
         public static ClusterVersion ParseClusterVersion(VersionInfo versionInfo)
         {
             return new ClusterVersion(SanitizeAndParseVersionNumber(versionInfo.Major), SanitizeAndParseVersionNumber(versionInfo.Minor));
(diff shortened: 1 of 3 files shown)


APErebus (approved):
Excellent. Thanks!

source/Octopus.Tentacle/Kubernetes/KubernetesVersionParser.cs
```diff
@@ -0,0 +1,26 @@
...
+        {
+            return new ClusterVersion(SanitizeAndParseVersionNumber(versionInfo.Major), SanitizeAndParseVersionNumber(versionInfo.Minor));
+        }
+        
+        static int SanitizeAndParseVersionNumber(string version)
+        {
+            if (int.TryParse(version, out var result))
+                return result;
+            
+            var sanitized = Regex.Replace(version, "[^0-9]", "");
```
liam-mackie, line 19: Is there a reason we don't just compile the Regex and always sanitise, rather than trying to parse, then using the regex if it can't?

liam-mackie (approved):
LGTM - one minor comment about Regexes, but trust your judgement. According to SemVer, the minor including `+` breaks the spec, as it **MUST** be a positive integer. I don't see a reason we would need to do anything more.

source/Octopus.Tentacle/Kubernetes/KubernetesVersionParser.cs
tleed5: There isn't a proper reason 😂 , I just went with the method that co-pilot generated initially. Ill update it to always sanitize since I think that will be more readable and I doubt we get any performance gains by doing the `TryParse` first anyway

APErebus: @liam-mackie I had that thought but this check is only run once (as it's cached in a `AsyncLazy`), so the optimization of a compiled regex isn't that necessary. We probably don't need the `TryParse`, but I don't mind it as it's the expected result

liam-mackie: @APErebus I was just curious (for my own purposes) - I assumed that we didn't run the Regex constantly because there was a performance concern if we parsed regardless, so wondered why we'd do the `TryParse` instead of a precompiled Regex (I.E. is a precompiled regex still terrible to use 😅 )

source/Octopus.Tentacle/Kubernetes/KubernetesVersionParser.cs
```diff
@@ -0,0 +1,24 @@
+using System;
+using System.Text.RegularExpressions;
+using k8s.Models;
+
+namespace Octopus.Tentacle.Kubernetes
+{
+    public static class KubernetesVersionParser
+    {
+        static Regex clusterVersionRegex = new Regex("[^0-9]");
```
APErebus, line 9: We can drop this :)

david-staniec-octopus (approved):
LGTM. I did notice you sanitize the major and minor version. Is this required only for the minor versions or does it impact both?

tleed5:
> LGTM. I did notice you sanitize the major and minor version. Is this required only for the minor versions or does it impact both?

@david-staniec-octopus We only observed this issue with the minor version but been overly cautious I opted for to sanitize both of them just in case

PR #893, opened 2024-04-17
Fix bug where bootstrapper copy fails when restarting Tentacle

flin-8:
[sc-75987]
# Background

If Tentacle is restarted just after it copies the Golang bootstrapper, it'll fail to copy it on the next try because we are not allowing the file copy to overwrite an existing file.

This PR allows the file to be overwritten like how [PrepareWorkspace works](https://github.com/OctopusDeploy/OctopusTentacle/blob/85a209e9b54f079edfbe256790679d946392a94a/source/Octopus.Tentacle/Scripts/ScriptWorkspaceFactory.cs#L94) - where it always overwrites.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


scme0 (approved):

APErebus (approved):
great!

PR #900, opened 2024-04-22
Use container status, not pod status to determine phase

liam-mackie:
# Background
If there are multiple containers in a script pod, we would wait for all containers to exit cleanly. This may not always be the case. For example, the NFS watchdog runs as a sidecar, and if the cluster has a service mesh with sidecar injection, the same issue would arise.
Fixes [sc-76109]

# Results
Now, when we check the pod's status, we check the container status of the script pod to determine whether the job completed successfully or not. We check this container status if the pod is in `Running`, `Failed`, or `Successful`, as the container itself may be successfully completed in any of these situations. 

I considered making integration tests for this, but wanted to get the fix in first.

# How to review this PR

Quality :heavy_check_mark:

I've completed smoke testing, but it may be worth completing some more on a live cluster (@flin-8).

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle/Kubernetes/KubernetesPodMonitor.cs
+++ b/source/Octopus.Tentacle/Kubernetes/KubernetesPodMonitor.cs
@@ -240,19 +240,32 @@ namespace Octopus.Tentacle.Kubernetes
         {
             switch (pod.Status?.Phase)
                 {
+                    // The pod isn't yet scheduled
                     case PodPhases.Pending:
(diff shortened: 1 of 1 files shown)


flin-8 (approved):
Looks like a nice targeted change - let me give it a quick test

APErebus (approved):
Love it!

flin-8:
> Looks like a nice targeted change - let me give it a quick test

looks good 👍

PR #901, opened 2024-04-22
Make Kubernetes service idempotent

flin-8:
[sc-74858]

# Background

There may be times where the same call may be made twice due to Tentacle restarts (it performs some action but restarts before returning the result, making Server retry the same call). We want to ensure the deployment still behaves correctly when this happens.

This PR makes the script cancellation idempotent and adds notes in the code around assumptions made around idempotency.

- StartScript
  - Already idempotent assuming the K8s API reflects the new Pod's existence quickly enough
- CancelScript
  - The Pod deletion can now be called more than once
- GetStatus
  - This is a read operation, no change required
- CompleteScript
  - Already idempotent 

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Octopus.Tentacle/Services/Scripts/Kubernetes/KubernetesScriptServiceV1Alpha.cs
```diff
@@ -58,7 +58,16 @@ public async Task<KubernetesScriptStatusResponseV1Alpha> StartScriptAsync(StartK
...
+                //We shouldn't really need to worry about starting the same script twice,
+                //since we wait for the Pods to be reloaded from the K8s API on Tentacle restart.
+                //We might try to start the same script twice if
+                // - a Pod gets created
+                // - Tentacle restarts before returning
+                // - Server retries StartScriptAsync()
+                // - The Kubernetes API doesn't say that the Pod exists when we query it.
+                
+                //Note: PrepareWorkspace overwrites "command.Files", so it's not strictly idempotent
                 var workspace = await workspaceFactory.PrepareWorkspace(command.ScriptTicket,
```
APErebus, line 71: I think this is fine as it will be overwriting it with the same files anyway

APErebus (approved):

source/Octopus.Tentacle/Services/Scripts/Kubernetes/KubernetesScriptServiceV1Alpha.cs
flin-8: Just depends what happens if a script is trying to read the file as it's being overwritten

PR #904, opened 2024-04-24
Use `du` for space checks in Kubernetes

liam-mackie:
**Note that this change will require a [new helm chart version](https://github.com/OctopusDeploy/helm-charts/pull/125) to get a required environment variable!** 

# Background
When running Octopus Tentacle in the Kubernetes Agent, we can't reliably get disk space on the mounted volume using native disk space tools. This is because in some cases (especially using the default NFS implementation), the volume is actually an NFS export of a folder on the node. This means that disk space tools (like `df`) show the disk space free on the _node_ rather than the disk space that we've configured in Kubernetes to be available. When we exceed the disk space we've configured in Kubernetes, `kubelet` will evict the NFS pod.

As such, we've fallen back to using a simple `du` command to quickly get the disk space used on the filesystem. This is compared to the limit that's set on the PVC for tentacle. This is set as an environment variable by the helm chart.

# Results
_Most of the time, this works every time!™_

image

`du` can have some issues getting space, for example:
* If the metadata is slow (using CIFS with many small files), it can take some time
* If we have no access to a file or folder, it can't count the space it takes up

We attempt to do our best, logging out warnings with the issues if we do experience a problem, but taking the space that it still managed to account for, and returning this. If it can't get _anything_ from DU, we just force the check to succeed, in an attempt to not block deployments where possible.

As a part of this, I've also written a non-static, injectable wrapper around the `SilentProcessRunner` so that I could mock it in the unit tests for the `KubernetesDirectoryInformationProvider`.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Tests/Kubernetes/KubernetesDirectoryInformationProviderFixture.cs
+++ b/source/Octopus.Tentacle.Tests/Kubernetes/KubernetesDirectoryInformationProviderFixture.cs
@@ -1,0 +1,90 @@
+using System;
+using FluentAssertions;
+using NSubstitute;
+using NSubstitute.Extensions;
+using NUnit.Framework;
+using Octopus.Diagnostics;
+using Octopus.Tentacle.Kubernetes;
+using Octopus.Tentacle.Util;
+
+namespace Octopus.Tentacle.Tests.Kubernetes
+{
+    public class KubernetesDirectoryInformationProviderFixture
+    {
+        // Sizes
+        const ulong Megabyte = 1000 * 1000;
+        const ulong Mebibyte = 1024 * 1024;
+        const ulong Gibibyte = 1024 * 1024 * 1024;
+        
+        [SetUp]
+        public void SetUp()
+        {
+            Environment.SetEnvironmentVariable(KubernetesConfig.NamespaceVariableName, "test-namespace");
+        }
+
+        [Test]
+        public void DuOutputParses()
+        {
+            const ulong usedSize = 500 * Megabyte;
+            var spr = Substitute.For<ISilentProcessRunner>();
+            spr.When(x => x.ExecuteCommand("du", "-s -B 1 /octopus", "/", Arg.Any<Action<string>>(), Arg.Any<Action<string>>()))
+                .Do(x =>
+                {
+                    x.ArgAt<Action<string>>(3).Invoke($"{usedSize}\t/octopus");
+                });
+            var sut = new KubernetesDirectoryInformationProvider(Substitute.For<ISystemLog>(), spr);
+            sut.GetPathUsedBytes("/octopus").Should().Be(usedSize);
+        }
+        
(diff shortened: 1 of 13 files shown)


source/Octopus.Tentacle.Tests/Kubernetes/KubernetesDirectoryInformationProviderFixture.cs
```diff
@@ -0,0 +1,90 @@
...
+    {
+        // Sizes
+        const ulong Megabyte = 1000 * 1000;
+        const ulong Mebibyte = 1024 * 1024;
+        const ulong Gibibyte = 1024 * 1024 * 1024;
+        
+        [SetUp]
+        public void SetUp()
+        {
+            Environment.SetEnvironmentVariable(KubernetesConfig.NamespaceVariableName, "test-namespace");
```
APErebus, line 22: Do we need this anymore?

source/Octopus.Tentacle.Tests/Kubernetes/KubernetesDirectoryInformationProviderFixture.cs
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        [TestCase("1Gi", 1 * Gibibyte)]
+        [TestCase("10Gi", 10 * Gibibyte)]
+        [TestCase("10Mi", 10 * Mebibyte)]
+        [TestCase("1M", 1 * Megabyte)]
+        public void CorrectlyParsesTotalSize(string sizeString, ulong expected)
+        {
+            KubernetesUtilities.GetResourceBytes(sizeString).Should().Be(expected);
+        }
```
APErebus, line 87: Can we split this into it's own test class `KubernetesUtilitiesFixture`

APErebus (approved):
Two minor things, but excellent stuff 👍

source/Octopus.Tentacle.Tests/Kubernetes/KubernetesDirectoryInformationProviderFixture.cs
liam-mackie: Ah you're right! Since we just DI it in, we don't :)

source/Octopus.Tentacle/Util/ShellModule.cs
```diff
@@ -9,6 +9,8 @@ public class ShellModule : Module
         protected override void Load(ContainerBuilder builder)
         {
             base.Load(builder);
+            
+            builder.RegisterType<SilentProcessRunnerWrapper>().As<ISilentProcessRunner>().SingleInstance();
```
LukeButters, line 13: Good idea

LukeButters (approved):
Approved from the SaST side, although this doesn't include reviewing the k8s namespaces which will need to be done by the originating team :)

PR #905, opened 2024-04-24
Retry when Pod log line numbers don't match

flin-8:
[sc-76250]
# Background

During test, we found that sometimes the Pod logs returned by the K8s API was missing the first 3 lines, but the lines were there on a retry.

This problem is difficult to reproduce, so we've decided to add a simple retry to cater for the case where the K8s API isn't returning the Pod log lines we're expecting.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Kubernetes.Tests.Integration/KubernetesScriptServiceV1AlphaIntegrationTest.cs
+++ b/source/Octopus.Tentacle.Kubernetes.Tests.Integration/KubernetesScriptServiceV1AlphaIntegrationTest.cs
@@ -16,8 +16,15 @@ public class KubernetesScriptServiceV1AlphaIntegrationTest : KubernetesAgentInte
         // Arrange
         var logs = new List<ProcessOutput>();
         var scriptCompleted = false;
+        string scriptBody = @"echo ""Hello World""
+for i in $(seq 1 50);
+do
+    echo $i
+    sleep 0.1
+done
+";
         var command = new ExecuteKubernetesScriptCommandBuilder(LoggingUtils.CurrentTestHash())
-            .SetScriptBody("echo \"Hello World\"")
+            .SetScriptBody(scriptBody)
             .Build();
 
         //act
--- a/source/Octopus.Tentacle/Kubernetes/IKubernetesPodLogService.cs
+++ b/source/Octopus.Tentacle/Kubernetes/IKubernetesPodLogService.cs
@@ -60,7 +60,10 @@ namespace Octopus.Tentacle.Kubernetes
                 }
                 catch (UnexpectedPodLogLineNumberException ex)
                 {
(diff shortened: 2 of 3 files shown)


source/Octopus.Tentacle.Kubernetes.Tests.Integration/KubernetesScriptServiceV1AlphaIntegrationTest.cs
```diff
@@ -16,8 +16,15 @@ public async Task SimpleHelloWorld()
         // Arrange
         var logs = new List<ProcessOutput>();
         var scriptCompleted = false;
+        string scriptBody = @"echo ""Hello World""
+for i in $(seq 1 50);
```
flin-8, line 20: Make the script slightly longer to make it more "real"

source/Octopus.Tentacle/Kubernetes/IKubernetesPodLogService.cs
```diff
@@ -57,7 +51,28 @@ public async Task<(IReadOnlyCollection<ProcessOutput> Outputs, long NextSequence
...
+            async Task<(IReadOnlyCollection<ProcessOutput> Outputs, long NextSequenceNumber, int? ExitCode)> GetPodLogs()
+            {
+                var sinceTime = scriptPodSinceTimeStore.GetSinceTime(scriptTicket);
+                try
+                {
+                    return await GetPodLogsWithSinceTime(sinceTime);
+                }
+                catch (UnexpectedPodLogLineNumberException ex)
+                {
+                    Log.Warn(ex, "Unexpected Pod log line numbers found, retrying with loading ");
```
flin-8, line 63: Should we try log this in the task log?

APErebus: yeah, I think that's a good idea

APErebus (approved):

source/Octopus.Tentacle/Kubernetes/IKubernetesPodLogService.cs
scme0: I would say it's probably not worth putting in the task log, if it's a "real" issue, then it'll fail again when we retry and that error will bubble up. If it's just a little blip that we handle automatically it's not worth bothering the user about.

scme0 (approved):
Nice, looks good!

source/Octopus.Tentacle/Kubernetes/IKubernetesPodLogService.cs
scme0: I didn't see Al's comment before I submitted my review. Adding it to the task log if it's Verbose is probably fine, I'd avoid making it a warning or info level though. 👍

PR #917, opened 2024-04-30
Add simple cache to directory information provider

liam-mackie:
# Background
When we run `du`, it takes a lot longer to get information back instead of using the normal disk APIs. This can slow down scripts quite a lot when running in parallel. The disk space is unlikely to change much in a short period of time so we plan to cache the results to speed things up a bit. 

[There is an ADR for this change.
](https://github.com/OctopusDeploy/adr/pull/10)
# Results

This PR adds an IMemoryCache that caches for 30s, as well as tests to ensure the cache behaviour is correct. In practice, we've seen the following results:
_For 50 Concurrent Cloud deployments_
| Cache Length | Time per Deployment |
| ------------ | ------------------- |
| No DU        | 8min                |
| No Cache     | 30min               |
| 60s          | 9min                |
| 30s          | 11min               |
| 15s          | 15min               |

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Tests/Kubernetes/KubernetesDirectoryInformationProviderFixture.cs
+++ b/source/Octopus.Tentacle.Tests/Kubernetes/KubernetesDirectoryInformationProviderFixture.cs
@@ -2,6 +2,7 @@ using System;
 using System.Threading;
 using FluentAssertions;
 using Microsoft.Extensions.Caching.Memory;
+using Microsoft.Extensions.Internal;
 using NSubstitute;
 using NSubstitute.Extensions;
 using NUnit.Framework;
@@ -88,7 +89,7 @@ namespace Octopus.Tentacle.Tests.Kubernetes
             spr.When(x => x.ExecuteCommand("du", "-s -B 1 /octopus", "/", Arg.Any<Action<string>>(), Arg.Any<Action<string>>()))
                 .Do(x =>
                 {
-                    x.ArgAt<Action<string>>(3).Invoke($"500\t/octopus");
+                    x.ArgAt<Action<string>>(3).Invoke($"123\t/octopus");
                     x.ArgAt<Action<string>>(3).Invoke($"{usedSize}\t/octopus");
                 });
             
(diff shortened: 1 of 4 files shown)


source/Octopus.Tentacle.Tests/Kubernetes/KubernetesDirectoryInformationProviderFixture.cs
```diff
@@ -65,8 +70,52 @@ public void IfDuFailsCompletelyReturnNull()
...
+            
+            sut.GetPathUsedBytes("/octopus").Should().Be(null);
+        }
+        
+        [Test]
+        public void DuCacheExpiresAfter15Seconds()
+        {
+            var spr = Substitute.For<ISilentProcessRunner>();
+            spr.ReturnsForAll(1);
+            var memoryCache = new MemoryCache(new MemoryCacheOptions());
```
flin-8, line 103: `MemoryCacheOptions` has a `Clock` property where you can override the default clock for testing. This way we won't need to call `Thread.Sleep()`

source/Octopus.Tentacle/Kubernetes/KubernetesModule.cs
```diff
@@ -24,6 +25,7 @@ protected override void Load(ContainerBuilder builder)
             builder.RegisterType<KubernetesOrphanedPodCleanerTask>().As<IKubernetesOrphanedPodCleanerTask>().As<IBackgroundTask>().SingleInstance();
             builder.RegisterType<KubernetesOrphanedPodCleaner>().As<IKubernetesOrphanedPodCleaner>().SingleInstance();
             builder.RegisterType<KubernetesDirectoryInformationProvider>().As<IKubernetesDirectoryInformationProvider>().SingleInstance();
+            builder.RegisterType<MemoryCache>().As<IMemoryCache>().SingleInstance();
```
flin-8, line 28: `MemoryCache` isn't resolvable directly without the options

APErebus (approved):

PR #926, opened 2024-05-08
Remove Octopus.Diagnostics from Tentacle contracts

tothegills:
# Background

The deprecated package `Octopus.Diagnostics` crept into the Tentacle contracts. Octopus Server maintains a version of `Octopus.Diagnostics`, so we had a bit of a dependency collision.

This PR decouples the logging interface used by `Tentacle.Contracts` from `Octopus.Diagnostics`.

### DIFF
--- a/source/Octopus.Tentacle.Contracts/Logging/ITaskLog.cs
+++ b/source/Octopus.Tentacle.Contracts/Logging/ITaskLog.cs
@@ -2,7 +2,7 @@
 
 namespace Octopus.Tentacle.Contracts.Logging
 {
-    public interface IOperationLog
+    public interface ITaskLog
     {
         void Info(string message);
         void Verbose(string message);
--- a/source/Octopus.Tentacle/Diagnostics/Log.cs
+++ b/source/Octopus.Tentacle/Diagnostics/Log.cs
@@ -8,7 +8,7 @@ using Octopus.Tentacle.Util;
 
 namespace Octopus.Tentacle.Diagnostics
 {
-    public abstract class Log : ILog, ITentacleTaskLog, IDisposable
+    public abstract class Log : ILog, IDisposable
     {
         public static ConcurrentBag<ILogAppender> Appenders { get; } = new();
         static IEnumerable<ILogAppender> GetThreadSafeAppenderCollection() => Appenders.ToArray();
--- a/source/Octopus.Tentacle/Diagnostics/Log.cs
+++ b/source/Octopus.Tentacle/Diagnostics/Log.cs
@@ -8,7 +8,7 @@ using Octopus.Tentacle.Util;
 
 namespace Octopus.Tentacle.Diagnostics
 {
-    public abstract class Log : ILog, IOperationLog, IDisposable
+    public abstract class Log : ILog, ITaskLog, IDisposable
     {
         public static ConcurrentBag<ILogAppender> Appenders { get; } = new();
         static IEnumerable<ILogAppender> GetThreadSafeAppenderCollection() => Appenders.ToArray();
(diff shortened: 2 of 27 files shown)


source/Octopus.Tentacle/Diagnostics/Log.cs
```diff
@@ -2,12 +2,13 @@
...
 using System.Collections.Generic;
 using Octopus.Diagnostics;
+using Octopus.Tentacle.Contracts.Logging;
 using Octopus.Tentacle.Diagnostics.Formatters;
 using Octopus.Tentacle.Util;
 
 namespace Octopus.Tentacle.Diagnostics
 {
-    public abstract class Log : ILog, IDisposable
+    public abstract class Log : ILog, ITaskLog, IDisposable
```
sburmanoctopus, line 11: Just wondering, does this have to implement both `ILog` and `ITaskLog`?

sburmanoctopus (approved):
LGTM.

At some stage, we will have to remove `Octopus.Diagnostics` from the other projects as well. But that's for another PR 🙂

source/Octopus.Tentacle.Contracts/Logging/ITaskLog.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+
+namespace Octopus.Tentacle.Contracts.Logging
+{
+    public interface ITaskLog
```
LukeButters, line 5: From tentacles perspective this is not necessarily a task, this should have a different name.

borland: Per the slack thread, what _is_ it if not the task? There wasn't an obvious name :-(

source/Octopus.Tentacle/Diagnostics/Log.cs
LukeButters: I think this is needed for the tests to compile, but really the Integration tests should just all be using the to be renamed `ITaskLog`

source/Octopus.Tentacle.Contracts/Logging/ITaskLog.cs
LukeButters: This client is also used outside of octopus e.g. when configuring a DW.

What is it? A thing the client will log to.

LukeButters: `ITentacleClientLogger` perhaps?

LukeButters: As I think about it more, `ITaskLog` is the same name as an interface use internally in octopus, so by having one here it does feel like we are going to offer the chance of importing the wrong one or this interface being used in places it should not be.

A name like `ITentacleClientLogger` or even `ITentacleClientTaskLog` should prevent it from ever being used in the wrong place, which as I think about it is my larger concern.

source/Octopus.Tentacle/Diagnostics/Log.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using Octopus.Diagnostics;
+using Octopus.Tentacle.Contracts.Logging;
```
LukeButters, line 5: nit: This class still needs the using?

LukeButters (approved):
LGTM, can be merged once we have a PR in server which depends on the updated tentacle client && the FULL chain passes on server.

tothegills:
[Passing server PR](https://github.com/OctopusDeploy/OctopusDeploy/pull/24575) with updated contracts and Tentacle.

PR #927, opened 2024-05-08
Added new options to register machine and import cert

tleed5:
# Background
When trying to create/manage the Kubernetes agent within an automated context i.e.  Terraform there is a situation where you create the deployment target which must have a URI and Thumbprint which is unknown until the agent registers itself with server and updates the deployment target with the URI and Thumbprint that it has generated.

[SC-78187]
Related to https://github.com/OctopusDeploy/helm-charts/pull/148 and https://github.com/OctopusDeployLabs/terraform-provider-octopusdeploy/pull/636

# Results
- Added `--server-subscription-id` option to `RegisterMachineCommand` to allow setting a custom subscription id for a active tentacle
- Added `--from-base64` to `ImportCertificateCommand` to allow importing a cert via a base64 formatted string
- Updated the kubernetes dockerfile and run-and-configure.sh script to pass through these new options if they are set in the env variables.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


docker/kubernetes-tentacle/dev/Dockerfile
```diff
@@ -64,6 +65,7 @@ ENV OCTOPUS__K8STENTACLE__PODVOLUMEJSON=""
 ENV OCTOPUS__K8STENTACLE__DISABLEAUTOPODCLEANUP="False"
 ENV TentacleHome=""
 ENV TentacleApplications=""
+ENV TentacleCertificate=""
```
APErebus, line 68: should we specify that this is Base64?
```suggestion
ENV TentacleCertificateBase64=""
```

docker/kubernetes-tentacle/scripts/configure-and-run.sh
```diff
@@ -103,6 +106,9 @@ function validateVariables() {
     if [[ -n "$DefaultNamespace" ]]; then
         echo " - default namespace '$DefaultNamespace'"
     fi
+    if [[ -n "$TentacleCertificate" ]]; then
+        echo " - tentacle certificate custom"
```
APErebus, line 110: Could we show the first X and last X chars of the base 64 value? Something like `32B...3s=`

source/Octopus.Tentacle/Commands/RegisterMachineCommandBase.cs
```diff
@@ -75,6 +76,7 @@ public abstract class RegisterMachineCommandBase<TRegistrationOperationType> : A
             Options.Add("server-comms-port=", "When using active communication, the comms port on the Octopus Server; the default is " + DefaultServerCommsPort + ". If specified, this will take precedence over any port number in server-comms-address.", s => serverCommsPort = int.Parse(s));
             Options.Add("server-comms-address=", "When using active communication, the comms address on the Octopus Server; the address of the Octopus Server will be used if omitted.", s => serverCommsAddress = s);
             Options.Add("server-web-socket=", "When using active communication over websockets, the address of the Octopus Server, eg 'wss://example.com/OctopusComms'. Refer to http://g.octopushq.com/WebSocketComms", s => serverWebSocketAddress = s);
+            Options.Add("server-subscription-id=", "When using active communication, the subscription id is what Octopus Server uses to identify this Tentacle this must be a valid URI e.g. poll://foobar/. If not specified, a new subscription id will be generated.", s => serverSubscriptionId = s);
```
APErebus, line 79: Is there a way to validate that this is `poll://ABC/`? SO people don't put junk here? Also, I think they are constrained to 20 chars?

APErebus (approved):
Just 1 question, but apart from that 🚀

source/Octopus.Tentacle/Commands/RegisterMachineCommandBase.cs
tleed5: > Is there a way to validate that this is poll://ABC/

Currently it does enforce that it's a valid URI since we pass the value to the Uri constructor but we can take it a step further and validate that the scheme equals `poll`

> Also, I think they are constrained to 20 chars?

I thought so too but even though server mentions 20 characters it doesn't actually restrict it. It only fails when the length is longer than the maximum host name length
Screenshot 2024-05-09 at 2 09 42 pm

APErebus (approved):

source/Octopus.Tentacle/Commands/ImportCertificateCommand.cs
```diff
@@ -28,17 +29,18 @@ public ImportCertificateCommand(Lazy<IWritableTentacleConfiguration> tentacleCon
...
         protected override void Start()
         {
             base.Start();
-            if (!fromRegistry && string.IsNullOrWhiteSpace(importFile))
+            if (!fromRegistry && string.IsNullOrWhiteSpace(importFile) && string.IsNullOrWhiteSpace(importBase64))
                 throw new ControlledFailureException("Please specify the certificate to import.");
 
-            if (fromRegistry && !string.IsNullOrWhiteSpace(importFile))
-                throw new ControlledFailureException("Please specify only one of either from-registry or from-file.");
+            if (fromRegistry && (!string.IsNullOrWhiteSpace(importFile) || !string.IsNullOrWhiteSpace(importBase64)))
```
sburmanoctopus, line 42: It looks like the purpose of this guard is to only allow 1 of the 3 options, no more.

However, if `fromRegistry` is false, I think this `if` statement will happily allow `importFile` and `importBase64` to both be set.

sburmanoctopus: Could we write a test for this as well?

tleed5: Ah yup I think you are right. Ill fix it up

sburmanoctopus (approved):

docker/kubernetes-tentacle/scripts/configure-and-run.sh
zentron: A note on this one. While the update the the cert characters is a good idea, the `MII` character will appear for all X509 certs and the `==` is just empty padding characters so might not be particularly unique. 

Would it have been possible to print the thumbprint?

tleed5: It would be possible but we would have to calculate the thumbprint within the script itself and I don't think it's really worth it considering the thumbprint is logged out when the import-certificate command is run which happens a couple of lines below

```
Importing custom certificate ...
2024-05-10 01:51:53.1827 |  INFO | Importing the certificate via base64...
2024-05-10 01:51:53.2284 |  INFO | Machine key file does not yet exist. Generating key that will be used to encrypt data for this tentacle.
2024-05-10 01:51:53.2629 |  INFO | Certificate with thumbprint A21E3DAA2815C532531D354CA81887E55A8ACBFD imported successfully.
2024-05-10 01:51:53.2652 |  WARN | These changes require a restart of the Tentacle.
```

zentron: If we are logging it below thats good, more to the point that output probably isnt particularly useful :) 
Not something to worry about at the moment since its just a log message and already merged.

PR #929, opened 2024-05-15
Fixing DuringStartScript_ForListeningTentacle_ThatIsRetryingTheRpc_AndConnecting_ScriptExecutionCanBeCancelled 

sburmanoctopus:
[sc-70984]

# Background
The `ClientScriptExecutionCanBeCancelledWhenRetriesAreEnabled.DuringStartScript_ForListeningTentacle_ThatIsRetryingTheRpc_AndConnecting_ScriptExecutionCanBeCancelled` test was flaky. 

On closer inspection, it was found that when it failed, it was forever trying to send a `Cancel` message to the Tentacle when it should not be.

It looks as though what was happening was that an exception was being thrown during "prepare exchange" in Halibut, and this exception was being considered as `ConnectionState.Unknown`. 

The reason we care about the connection state is because if we cancel, and encounter an exception while connecting, then we can walk away. The Tentacle never received any instructions.

But if we encounter an exception after we have connected, then the Tentacle may have received instructions, and we should send a "Cancel" message to the Tentacle.

# Results

## Before

If we managed to get a successful connection, or reused an existing connection, but failed during the `MessageExchangeProtocol.PrepareExchangeAsClientAsync` (i.e. before we ever sent any request information), it would **not** be considered "connecting" in the exception.

This logic is found within Halibut itself.

## After

We now consider an exception here as "connecting". This was fixed in Halibut with [this PR](https://github.com/OctopusDeploy/Halibut/pull/609)

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionCanBeCancelledWhenRetriesAreEnabled.cs
+++ b/source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionCanBeCancelledWhenRetriesAreEnabled.cs
@@ -346,6 +346,7 @@ namespace Octopus.Tentacle.Tests.Integration
                 .WithRetryDuration(TimeSpan.FromHours(1))
                 .WithTentacleServiceDecorator(new TentacleServiceDecoratorBuilder().RecordMethodUsages(tentacleConfigurationTestCase, out var scriptServiceRecordedUsages).Build())
                 .WithHalibutTimeoutsAndLimits(halibutTimeoutsAndLimits)
+                .WithPortForwarder(out var portForwarder)
                 .Build(CancellationToken);
 
             // Arrange
@@ -356,6 +357,8 @@ namespace Octopus.Tentacle.Tests.Integration
 
             Logger.Information("Stop Tentacle so no more requests are picked up");
             await clientAndTentacle.RunningTentacle.Stop(CancellationToken);
+            portForwarder.Value.KillNewConnectionsImmediatlyMode = true;
+            portForwarder.Value.CloseExistingConnections();
 
             var delay = clientAndTentacle.Server.ServerHalibutRuntime.TimeoutsAndLimits.SafeTcpClientPooledConnectionTimeout + TimeSpan.FromSeconds(2);
             Logger.Information($"Waiting for {delay} for any active Pooled Connections for the Tentacle to expire");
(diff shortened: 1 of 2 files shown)


source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionCanBeCancelledWhenRetriesAreEnabled.cs
```diff
@@ -346,6 +346,7 @@ public async Task DuringStartScript_ForListeningTentacle_ThatIsRetryingTheRpc_An
                 .WithRetryDuration(TimeSpan.FromHours(1))
                 .WithTentacleServiceDecorator(new TentacleServiceDecoratorBuilder().RecordMethodUsages(tentacleConfigurationTestCase, out var scriptServiceRecordedUsages).Build())
                 .WithHalibutTimeoutsAndLimits(halibutTimeoutsAndLimits)
+                .WithPortForwarder(out var portForwarder)
```
sburmanoctopus, line 349: Let's validate that the port forwarder change is something worth keeping.

While investigating the flaky test, we were reminded that if multiple tests are running at the same time, then the port will become available again once we stop it.

When we call await clientAndTentacle.RunningTentacle.Stop(CancellationToken);, all it does is cancel the cancellation token and wait for the running tentacle task to finish. This would free up the port the tentacle is using.

I have kept this here, as it is a possible problem (unlikely, but possible).

It does make me wonder if this is something we should consider fixing in some for for all tests in general.

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionCanBeCancelledWhenRetriesAreEnabled.cs
```diff
@@ -377,6 +380,72 @@ public async Task DuringStartScript_ForListeningTentacle_ThatIsRetryingTheRpc_An
...
+            // Assert
+            scriptServiceRecordedUsages.ForStartScriptAsync().Completed.Should().BeGreaterOrEqualTo(2);
+            scriptServiceRecordedUsages.ForGetStatusAsync().Started.Should().Be(0);
+            scriptServiceRecordedUsages.ForCancelScriptAsync().Started.Should().BeGreaterOrEqualTo(0, "Script Execution does not need to be cancelled on Tentacle as it has not started");
+            scriptServiceRecordedUsages.ForCompleteScriptAsync().Started.Should().Be(0);
+        }
+
+        [Test]
+        [TentacleConfigurations(testPolling: false)]
+        public async Task DuringStartScript_ForListeningTentacle_ThatIsRetryingTheRpc_AndConnectingWithPooledConnection_ScriptExecutionCanBeCancelled(TentacleConfigurationTestCase tentacleConfigurationTestCase)
```
sburmanoctopus, line 404: This is the same test as before, but causing it to fail on a connection that was in the pool.

I.e., it's an exact copy except to make it hit the pool:
- Leave `TcpClientPooledConnectionTimeout` as it's 9 minute default (so we don't change it)
- No longer `Delay` after we close the connection

LukeButters (approved):

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionCanBeCancelledWhenRetriesAreEnabled.cs
LukeButters: This test is flaky, attempting to drop it in this PR https://github.com/OctopusDeploy/OctopusTentacle/pull/950

PR #931, opened 2024-05-15
Update to Kubernetes SDK 14 and add new test

APErebus:
# Background

The Kubernetes C# SDK that supports Kubernetes 1.30 has been [released](https://github.com/kubernetes-client/csharp/releases/tag/v14.0.2)

This PR updates the package references and also adds a new integration test which tests what happens when the tentacle pod is killed midway through a script.

# Results

Adds a new `KubeCtlTool` class that is used for interacting with KubeCtl. I think eventually this will have named methods like `DeletePod` etc

We will also need to update the documentation to indicate that v1.30 is a supported Kubernetes version

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Kubernetes.Tests.Integration/KubernetesClusterOneTimeSetUp.cs
+++ b/source/Octopus.Tentacle.Kubernetes.Tests.Integration/KubernetesClusterOneTimeSetUp.cs
@@ -18,12 +18,12 @@ public class KubernetesClusterOneTimeSetUp
         await installer.Install();
 
         //if we are not running in TeamCity, then we need to find the latest local tag and use that if it exists 
-        if (!TeamCityDetection.IsRunningInTeamCity())
+        if (!TeamCityDetection.IsRunningInTeamCity() && bool.TryParse(Environment.GetEnvironmentVariable("KubernetesAgentTests_UseLocalImage"), out var useLocal) && useLocal)
         {
             var imageLoader = new DockerImageLoader(KubernetesTestsGlobalContext.Instance.TemporaryDirectory, KubernetesTestsGlobalContext.Instance.Logger, kindExePath);
             KubernetesTestsGlobalContext.Instance.TentacleImageAndTag = imageLoader.LoadMostRecentImageIntoKind(installer.ClusterName);
         }
-        else
+        else if(TeamCityDetection.IsRunningInTeamCity())
         {
             var tag = Environment.GetEnvironmentVariable("KubernetesAgentTests_ImageTag");
             KubernetesTestsGlobalContext.Instance.TentacleImageAndTag = $"docker.packages.octopushq.com/octopusdeploy/kubernetes-tentacle:{tag}";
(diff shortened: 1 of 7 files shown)


source/Octopus.Tentacle.Kubernetes.Tests.Integration/KubernetesClusterOneTimeSetUp.cs
```diff
@@ -18,12 +18,12 @@ public async Task OneTimeSetUp()
...
 
         //if we are not running in TeamCity, then we need to find the latest local tag and use that if it exists 
-        if (!TeamCityDetection.IsRunningInTeamCity())
+        if (!TeamCityDetection.IsRunningInTeamCity() && bool.TryParse(Environment.GetEnvironmentVariable("KubernetesAgentTests_UseLocalImage"), out var useLocal) && useLocal)
         {
             var imageLoader = new DockerImageLoader(KubernetesTestsGlobalContext.Instance.TemporaryDirectory, KubernetesTestsGlobalContext.Instance.Logger, kindExePath);
             KubernetesTestsGlobalContext.Instance.TentacleImageAndTag = imageLoader.LoadMostRecentImageIntoKind(installer.ClusterName);
         }
-        else
+        else if(TeamCityDetection.IsRunningInTeamCity())
```
flin-8, line 26: So we don't need an `else` branch?

It might be clearer if we extract a method that returns `TentacleImageAndTag`, and then set it in this method like:

`KubernetesTestsGlobalContext.Instance.TentacleImageAndTag = GetTentacleImageAndTag()`

APErebus: yeah, fair

source/Octopus.Tentacle.Kubernetes.Tests.Integration/KubernetesScriptServiceV1IntegrationTest.cs
```diff
@@ -59,4 +59,66 @@ Task ScriptCompleted(CancellationToken ct)
...
+
+        await Task.WhenAll(scriptTask, killTask);
+
+        var result = scriptTask.Result;
+
+        commandResult = await KubeCtl.ExecuteNamespacedCommand("get pods -l app.kubernetes.io/name=octopus-agent -o \"Name\"");
+        var finalPodName = commandResult.StdOut.Single();
+
+        //Assert
+        logs.Should().Contain(po => po.Source == ProcessOutputSource.StdOut && po.Text == "Hello World");
```
flin-8, line 96: Is it possible to also assert that Halibut got an error and is retrying mid way through the script?

APErebus: 🤔 Possibly yes

APErebus: I actually don't think this test is working correctly. I may need to tweak the script so that it waits until the old pod is dead (or something like that)

LukeButters (approved):

APErebus:
There is a bug where the NuGet version is `14.0.9`, but the release in GitHub is `14.0.8`
Raised an issue here: https://github.com/kubernetes-client/csharp/issues/1575

PR #941, opened 2024-05-22
Pass PV storage information to script pods as Environment Variables for Calamari package retention

scme0:
# Background

The Kubernetes Script Pods use a PV mount to get access to the shared file system managed by the Kubernetes Tentacle. When Calamari is acquiring packages, it checks if it should do package cleanup first.

Currently the Calamari implementation cannot do package retention when in the Kubernetes agent because it is not able to calculate the amount of storage on the PV.

[sc-76306]

# Results

This PR gets the Kubernetes Tentacle to add PV storage information (total bytes and free bytes) to the environment variables passed to Script Pods.

There is a Calamari PR () to utilise these variables to correctly calculate if package retention needs to be executed.

# Screen shot:
Screenshot 2024-05-22 at 11 18 00

### DIFF
--- a/source/Octopus.Tentacle/Kubernetes/KubernetesConfig.cs
+++ b/source/Octopus.Tentacle/Kubernetes/KubernetesConfig.cs
@@ -29,8 +29,8 @@ namespace Octopus.Tentacle.Kubernetes
         public static string PersistentVolumeSizeVariableName => $"{EnvVarPrefix}__PERSISTENTVOLUMESIZE";
         public static string PersistentVolumeSize => GetRequiredEnvVar(PersistentVolumeSizeVariableName, "Unable to determine Persistent Volume Size");
 
-        public static string PersistentVolumeSizeBytesVariableName => $"{EnvVarPrefix}__PERSISTENVOLUMETOTALBYTES";
-        public static string PersistentVolumeFreeBytesVariableName => $"{EnvVarPrefix}__PERSISTENVOLUMEFREEBYTES";
+        public static string PersistentVolumeSizeBytesVariableName => $"{EnvVarPrefix}__PERSISTENTVOLUMETOTALBYTES";
+        public static string PersistentVolumeFreeBytesVariableName => $"{EnvVarPrefix}__PERSISTENTVOLUMEFREEBYTES";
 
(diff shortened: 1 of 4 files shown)


source/Octopus.Tentacle/Kubernetes/KubernetesScriptPodCreator.cs
```diff
@@ -234,6 +237,7 @@ void LogVerboseToBothLogs(string message, InMemoryTentacleScriptLog tentacleScri
 
         async Task<V1Container> CreateScriptContainer(StartKubernetesScriptCommandV1 command, IScriptWorkspace workspace, string podName, string scriptName, string homeDir)
         {
+            var spaceInformation = kubernetesPhysicalFileSystem.GetCachedStorageInformation();
```
scme0, line 240: At this point, storage information should always be calculated so we can use the cached version. If it's not calculated, this method will fall back to calculating the storage information itself.

source/Octopus.Tentacle/Kubernetes/KubernetesConfig.cs
```diff
@@ -29,6 +29,9 @@ public static class KubernetesConfig
         public static string PersistentVolumeSizeVariableName => $"{EnvVarPrefix}__PERSISTENTVOLUMESIZE";
         public static string PersistentVolumeSize => GetRequiredEnvVar(PersistentVolumeSizeVariableName, "Unable to determine Persistent Volume Size");
 
+        public static string PersistentVolumeSizeBytesVariableName => $"{EnvVarPrefix}__PERSISTENVOLUMETOTALBYTES";
+        public static string PersistentVolumeFreeBytesVariableName => $"{EnvVarPrefix}__PERSISTENVOLUMEFREEBYTES";
```
APErebus, line 33:
```suggestion
        public static string PersistentVolumeSizeBytesVariableName => $"{EnvVarPrefix}__PERSISTENTVOLUMETOTALBYTES";
        public static string PersistentVolumeFreeBytesVariableName => $"{EnvVarPrefix}__PERSISTENTVOLUMEFREEBYTES";
```

APErebus (changes requested):

APErebus (approved):

PR #942, opened 2024-05-22
Add bootstrap runner exe to the debug tentacle image

kevjt:
# Background

The dev Dockerfile is used to build the image of tentacle for remote debugging. This has not been kept up to date as it is missing the bootstrap runner executable.

# Results

The PR brings the debug tentacle image up to date with the prod one (at least in terms of the bootstrap runner part).


kevjt:
> Awesome. I assume there was no other changes?

There are a few more differences around `RUN mkdir /.dotnet` for something to do with "ensuring ownership of the certificate store". However, I'm not familiar with that so I left it out.

PR #943, opened 2024-05-22
Ensure workspace can be cleaned up even if Disk has less than 500MBs of free space

scme0:


scme0:
Closed as being handled by server at scale

gb-8:
Thanks for posting this PR which helps bring clarity to the issue.

The team has had a think about this and has some suggestions for a variant on this approach:

 - We think the readiness check should be moved into `PrepareWorkspace()` rather than `GetWorkspace()`.
 - `CheckReadiness()` could also be renamed to `EnsureEnoughFreeDiskSpace()` since that is all it does.
 - The flag to opted out of the check should be used to determine whether to call `EnsureEnoughFreeDiskSpace()`.

(See the discussion in the [shortcut story](https://app.shortcut.com/octopusdeploy/story/79762/completescript-checks-for-500mb-space-before-cleaning-up-workspace-tentacle) for the reasoning.)

PR #944, opened 2024-05-23
Fix formatting for configure-and-run.sh

scme0:
I just noticed this bash script was not formatted consistently so I just did a `Format Document` in Visual Studio Code.


kevjt (approved):
lgtm

PR #945, opened 2024-05-23
Allow users to specify multiple Server Comms Addresses to support HA Server configurations

scme0:
# Background

We need to allow users to specify multiple server comms addresses to support Octopus Server running in an HA configuration.

[sc-75000]

# Results

Fortunately, there is already a command in Tentacle that we can leverage to register extra server instances (`poll-server`) so I just added an extra variable which can have multiple values, one for each additional server to register.

To make things backwards compatible, I've kept the original variable and now support both in the following way:
If `ServerCommsAddress` (singular) is set, it will use that for the initial registration. In this case, if `ServerCommsAddresses` (plural) is also set, all those values will be registered via `poll-server`.

If `ServerCommsAddress` is **not** set, the first value of `ServerCommsAddresses` will be used instead, and the remaining values (if any) in `ServerCommsAddresses` will be registered via `poll-server`.

Related Helm Chart PR: https://github.com/OctopusDeploy/helm-charts/pull/160

### DIFF
--- a/docker/kubernetes-tentacle/scripts/configure-and-run.sh
+++ b/docker/kubernetes-tentacle/scripts/configure-and-run.sh
@@ -203,6 +203,7 @@ function registerTentacle() {
   if [[ -n "$ServerCommsAddress" || -n "$ServerCommsAddresses" || -n "$ServerPort" ]]; then
     ARGS+=('--comms-style' 'TentacleActive')
 
+    # If ServerCommsAddress (singular) is not set, use the first value in ServerCommsAddresses (plural)
     if [[ -n "$ServerCommsAddress" ]]; then
       ARGS+=('--server-comms-address' "$ServerCommsAddress")
     elif [[ -n "$ServerCommsAddresses" ]]; then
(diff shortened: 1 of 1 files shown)


docker/kubernetes-tentacle/scripts/configure-and-run.sh
```diff
@@ -245,6 +251,60 @@ function registerTentacle() {
...
+    fi
+
+    IFS=',' read -ra SERVER_ADDRESSES <<<"$ServerCommsAddresses"
+    len=${#SERVER_ADDRESSES[@]}
+
+    if [[ -z "$ServerCommsAddress" && len -eq 1 ]]; then
+      return;
+    fi
+
+    echo "Registering additional HA Servers..."
```
zentron, line 266: It looks like this will allow registering additional instances even non HA?

scme0: The way I have this setup, it uses the `ServerUrl`  to get the thumbprint and stuff, which points to the main server you're registering from. We could allow users to pass in different `ServerUrl`s for each Comms Address, but they would also need to pass in different API keys or bearer tokens for each one too. 🤔

scme0: As discussed [here](https://octopusdeploy.slack.com/archives/C0738NKQN7K/p1716509518356089), I think we can continue with this change and add registering with separate server instances later if there is a user call for it.

APErebus (approved):
Awesome! Can we just add a comment in script for the more complex bash _stuff_?

PR #946, opened 2024-05-24
Fix memory leak when Kubernetes Start Script fails

kevjt:
# Background
There is currently a possible memory leak when `KubernetesScriptServiceV1.StartScript` fails (e.g. pod creation error) because the SemaphoreSlim is not removed from the concurrent dictionary which will grow in size.

[sc-77248]

# Results
This PR encapsulates the SemaphoreSlim and the Dictionary lookup into a single thread-safe `IKeyedLock` class. The outcome is to ensure that disposal of the "KeyedLock" causes the underlying SemaphoreSlim to be both released as well as removed from the dictionary*.

*with the caveat that the semaphore is only removed if **there are no more references to it**, otherwise the lock would not behave as it should since other threads can simply create a new key


source/Octopus.Tentacle/Kubernetes/Synchronisation/Internal/SemaphoreSlimReleaser.cs
```diff
@@ -0,0 +1,24 @@
+using System;
+using System.Threading;
+
+namespace Octopus.Tentacle.Kubernetes.Synchronisation.Internal
+{
+    class SemaphoreSlimReleaser<T> : ISemaphoreSlimReleaser<T> where T : SemaphoreSlim
```
scme0, line 6: Rather than having a Factory and specific interface for this, I'd probably make it a private class in `ReferenceCountingKeyedLock`. Maybe same with `SemaphoreSlimWithReferenceCount`.

source/Octopus.Tentacle/Kubernetes/Synchronisation/KubernetesModule.cs
scme0, line 1: This filename needs to be updated to match the class name.

source/Octopus.Tentacle/Kubernetes/Synchronisation/KubernetesModule.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using Autofac;
+using Octopus.Tentacle.Kubernetes.Synchronisation.Internal;
+
+namespace Octopus.Tentacle.Kubernetes.Synchronisation
+{
+    public class SynchronisationModule : Module
```
scme0, line 7: I don't mind a separate Autofac module but I think if you make the `SemaphoreSlimReleaser` and `SemaphoreSlimWithReferenceCount` classes private, you would only need to register the `ReferenceCountingKeyedLock` so probably not worth having a separate Module.

scme0 (commented):
I've had a first pass, I think this looks pretty good! Nice work 👍 

I have a couple of comments about the code organisation etc but generally it looks like it works. thanks for writing tests as well. Keen to hear back from others as semaphores and locking can be tricky so I won't approve just yet.

source/Octopus.Tentacle/Kubernetes/Synchronisation/Internal/SemaphoreSlimReleaser.cs
kevjt: I'll see what I can do to reduce the number of classes/interfaces. I only added the Factory in order to inject some listeners for testing the dispose callback.

scme0: Ah yep fair enough. You might be able to introduce a seam instead 🤔

zentron (changes requested):
Creating new thread safety constructs can be tough but I think this was a pretty good shot. 

I'm not particularly keen on the tests though. The tests all center on the disposal, but in reality we don't care how many (or if at all) there is disposing that takes place, we just want to validate that the threads block as expected. 

In fact I can get these tests to pass by taking out the semaphore waiting and releasing.

The tests should be rewritten to focus on testing the actual desired behavior, namely that threads block.
I had a stab at that in a [linked PR](https://github.com/OctopusDeploy/OctopusTentacle/pull/951) but you might want to give it a run through on your own first. 

Once we are testing the blocking itself we can remove the hooks into the ctor.

Testing race conditions and thread safety are always a bit tricky, but the simpler code has its own benifits.

kevjt:
> Creating new thread safety constructs can be tough but I think this was a pretty good shot.
> 
> I'm not particularly keen on the tests though. The tests all center on the disposal, but in reality we don't care how many (or if at all) there is disposing that takes place, we just want to validate that the threads block as expected.
> 
> In fact I can get these tests to pass by taking out the semaphore waiting and releasing.
> 
> The tests should be rewritten to focus on testing the actual desired behavior, namely that threads block. I had a stab at that in a [linked PR](https://github.com/OctopusDeploy/OctopusTentacle/pull/951) but you might want to give it a run through on your own first.
> 
> Once we are testing the blocking itself we can remove the hooks into the ctor.
> 
> Testing race conditions and thread safety are always a bit tricky, but the simpler code has its own benifits.



> Creating new thread safety constructs can be tough but I think this was a pretty good shot.
> 
> I'm not particularly keen on the tests though. The tests all center on the disposal, but in reality we don't care how many (or if at all) there is disposing that takes place, we just want to validate that the threads block as expected.
> 
> In fact I can get these tests to pass by taking out the semaphore waiting and releasing.
> 
> The tests should be rewritten to focus on testing the actual desired behavior, namely that threads block. I had a stab at that in a [linked PR](https://github.com/OctopusDeploy/OctopusTentacle/pull/951) but you might want to give it a run through on your own first.
> 
> Once we are testing the blocking itself we can remove the hooks into the ctor.
> 
> Testing race conditions and thread safety are always a bit tricky, but the simpler code has its own benifits.

Your changes look good! Thanks for taking the time to write those. 
I guess I was mainly trying to cover the aspect of disposing to ensure that threads were releasing and there was 'something' being done about cleaning the dictionary. However, even with my tests I wasn't actually checking the contents of the dictionary so it's still hard to say that the keys were being removed properly. One thing I thought about was adding a dictionary size property to the interface, but it felt wrong to introduce a method just for testing purposes (which may or may not be accurate since it introduces code) and that it would break the abstraction of the keyed semaphore concept by revealing its internals. I agree with making the code simpler and testing its main purpose 👍, the dictionary key removal part I could do manually with the debugger anyway

zentron (approved):

PR #947, opened 2024-05-27
Lobs communication spike

tleed5:
# Description 
This is a POC to evaluate ways of handling communication between processes for live kubernetes object status

Includes changes from https://github.com/OctopusDeploy/OctopusTentacle/compare/main...robe/reverse-tentacle
Octopus Server side: https://github.com/OctopusDeploy/OctopusDeploy/compare/robe/reverse-tentacle 

# What it does 
- Configures a kestrel server to run when the tentacle process is running to handle gRPC requests
- Creates a basic gRPC service that forwards along a message to Octopus server using existing Halibut polling connections

# Example with simple Go App hitting the gRPC service in Tentacle all running in a cluster
https://github.com/OctopusDeploy/OctopusTentacle/assets/3272176/f618d078-8d25-4f2f-a6fe-5cc9ee8f0fa6

### DIFF
--- a/source/Octopus.Tentacle/Commands/RunAgentCommand.cs
+++ b/source/Octopus.Tentacle/Commands/RunAgentCommand.cs
@@ -4,7 +4,20 @@ using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
 using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
 using System.Security.Principal;
+using Autofac;
+using Autofac.Extensions.DependencyInjection;
+#if !NETFRAMEWORK
+using Microsoft.AspNetCore.Server.Kestrel.Https;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Server.Kestrel.Core;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Octopus.Tentacle.Communications.gRPC;
+#endif
 using Octopus.Diagnostics;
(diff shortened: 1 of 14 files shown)


source/Octopus.Tentacle/Commands/RunAgentCommand.cs
```diff
@@ -74,6 +97,74 @@ public class RunAgentCommand : AbstractStandardCommand
...
+                        options.ListenAnyIP(5001, listenOptions =>
+                        {
+                            listenOptions.Protocols = HttpProtocols.Http1;
+                            // listenOptions.UseHttps(configuration.Value.TentacleCertificate!);
+                        });
+                        options.ListenAnyIP(5002, listenOptions =>
+                        {
+                            listenOptions.Protocols = HttpProtocols.Http2;
+                            // listenOptions.UseHttps(configuration.Value.TentacleCertificate!);
+                        });
```
tleed5, line 127: This is just so I could get it running locally on my mac https://learn.microsoft.com/en-us/aspnet/core/grpc/troubleshoot?view=aspnetcore-8.0#unable-to-start-aspnet-core-grpc-app-on-macos

source/Octopus.Tentacle/Commands/RunAgentCommand.cs
```diff
@@ -74,6 +97,74 @@ public class RunAgentCommand : AbstractStandardCommand
...
+                    });
+                    webBuilder.UseStartup<Startup>();
+                }).UseServiceProviderFactory(new AutofacChildLifetimeScopeServiceProviderFactory(container));
```
tleed5, line 130: Had to update Autofac so I could easily get the existing container services into the Kestrel's container to allow for DI in the gRPC service

PR #948, opened 2024-05-28
Adjust Kubernetes Tentacle startup to always refresh server comms address registration

scme0:
# Background

Recently we added the ability to define multiple `ServerCommsAddresses` to the Kubernetes Tentacle so that it could be configured to run with an Octopus Server HA Cluster setup.

The missing functionality was allowing users to adjust this configuration after installation.

# Results

With these changes, it's possible to adjust the configuration by doing a helm update:
```bash
helm upgrade --atomic \
--set agent.serverCommsAddresses="{http://host.docker.internal:10943/,http://host.docker.internal:10942/}" \
--reuse-values \
--create-namespace --namespace octopus-agent-scme \
scme \
/Users/scottmerchant/Documents/dev/helm-charts/charts/kubernetes-agent
```

[sc-75000]

### DIFF
--- a/source/Octopus.Tentacle/Commands/PollCommand.cs
+++ b/source/Octopus.Tentacle/Commands/PollCommand.cs
@@ -28,6 +28,7 @@ namespace Octopus.Tentacle.Commands
         int? serverCommsPort = null;
         string serverWebSocketAddress = null!;
         string serverCommsAddress = null!;
+        bool reuseThumbprint = false;
 
         public PollCommand(Lazy<IWritableTentacleConfiguration> configuration,
                            ISystemLog log,
@@ -45,11 +46,12 @@ namespace Octopus.Tentacle.Commands
             this.log = log;
             this.selector = selector;
 
-            api = AddOptionSet(new ApiEndpointOptions(Options));
+            api = AddOptionSet(new ApiEndpointOptions(Options, allowBypass: true));
 
             Options.Add("server-comms-address=", "The comms address on the Octopus Server; the address of the Octopus Server will be used if omitted.", s => serverCommsAddress = s);
             Options.Add("server-comms-port=", "The comms port on the Octopus Server; the default is " + DefaultServerCommsPort + ". If specified, this will take precedence over any port number in server-comms-address.", s => serverCommsPort = int.Parse(s));
             Options.Add("server-web-socket=", "When using active communication over websockets, the address of the Octopus Server, eg 'wss://example.com/OctopusComms'. Refer to http://g.octopushq.com/WebSocketComms", s => serverWebSocketAddress = s);
+            Options.Add("reuse-server-thumbprint", "Reuse the Server Thumbprint from the first trusted server instance currently configured", _ => reuseThumbprint = true);
         }
 
         protected override void Start()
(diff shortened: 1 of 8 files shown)


source/Octopus.Tentacle/Configuration/ITentacleConfiguration.cs
```diff
@@ -129,6 +129,8 @@ public interface IWritableTentacleConfiguration : ITentacleConfiguration
         /// </summary>
         void ResetTrustedOctopusServers();
 
+        bool SetTrustedOctopusServers(IEnumerable<OctopusServerConfiguration>? newServers);
```
kevjt, line 132: Is it worth adding a brief docstring here to say that this will overwrite the trusted servers with the new list?

source/Octopus.Tentacle/Commands/PollCommand.cs
```diff
@@ -45,11 +46,14 @@ public class PollCommand : AbstractStandardCommand
             this.log = log;
             this.selector = selector;
 
-            api = AddOptionSet(new ApiEndpointOptions(Options));
+            // purposefully not adding this with AddOptionSet because
+            // we only want to validate it if reuseThumbprint is false.
+            api = new ApiEndpointOptions(Options);
```
kevjt, line 51: Understandable why this was done but it is hard to follow. If only validation was a separate method within `AbstractCommand` :(

source/Octopus.Tentacle/Commands/PollCommand.cs
```diff
@@ -60,38 +64,60 @@ protected override void Start()
...
+
+            configuration.Value.AddOrUpdateTrustedOctopusServer(serverConfiguration);
+            VoteForRestart();
+
+            log.Info("Polling endpoint configured");
+        }
+
+        OctopusServerConfiguration CreateServerConfigurationFromFirstTrustedServer(Uri serverAddress)
+        {
+            var firstTrustedServer = configuration.Value.TrustedOctopusServers.First();
```
kevjt, line 89: Is this guaranteed to be what the user expects as the first server? i.e. it is the same as either the first server in the list of `server-comms-addresses`

Related question, what happens when the `server-comms-address` was also supplied along with the list of extra comms addresses. Does the "first" trusted server instance refer to the server from the singular server-comms-address because that is preferred in the configure-and-run.sh script?

kevjt (commented):
Looks good, I just have a small question

source/Octopus.Tentacle/Commands/PollCommand.cs
scme0: Yeah it's dumb because that's all that is done when an OptionSet is added via `AddOptionSet`.

source/Octopus.Tentacle/Configuration/ITentacleConfiguration.cs
scme0: Good call, will do

source/Octopus.Tentacle/Commands/PollCommand.cs
scme0: There doesn't seem to be any kind of sorting done to the TrustedServers so the first Server should be the one that is registered first which should be the one that's used when registering the agent with OctopusServer.

If `ServerCommsAddress` (singular) is provided it's used for the initial registration with Server (eg: when the tentacle registers itself as a deployment target with server). If it's not provided, the first entry in `ServerCommsAddresses` (plural) is used. The left over entries in `ServerCommsAddresses` are registered as additional addresses.

So if `ServerCommsAddress` is provided, the first server will be that one. If it's not the first server will be `ServerCommsAddresses[0]`.

Does that make sense?

kevjt: Yea that makes sense. Just wondering if it will be clear to the user 🤔

kevjt (approved):
lgtm

source/Octopus.Tentacle/Commands/PollCommand.cs
scme0: I'm considering deprecating the singular form and we can remove it in the next major release of the helm chart 🤔

PR #949, opened 2024-05-28
Add some more integration tests for Kubernetes Agent

APErebus:
# Background

We need some more test overage for the `KubernetesScriptServiceV1` and it's interactions with `TentacleClient`. This PR does that 😄 

# Results

Adds 4 more tests:
- `TentaclePodIsTerminatedDuringScriptExecution_ShouldRestartAndPickUpPodStatus` which tests that when the tentacle pod is terminated, that the script continues running and that a new tentacle pod is created and continues reading the logs
- `WhenALongRunningScriptIsCancelled_TheScriptShouldStop` which tests that cancellation
- `SimpleScriptExitsWithErrorCode_ScriptFails`  which tests that when a script explicitly exits with an error code, that it is reflected
- `ScriptPodIsTerminatedDuringScriptExecution_ScriptFails` which tests when the script pod deleted that the script exits with an error

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Kubernetes.Tests.Integration/KubernetesClusterOneTimeSetUp.cs
+++ b/source/Octopus.Tentacle.Kubernetes.Tests.Integration/KubernetesClusterOneTimeSetUp.cs
@@ -17,25 +17,32 @@ public class KubernetesClusterOneTimeSetUp
         installer = new KubernetesClusterInstaller(KubernetesTestsGlobalContext.Instance.TemporaryDirectory, kindExePath, helmExePath, kubeCtlPath, KubernetesTestsGlobalContext.Instance.Logger);
         await installer.Install();
 
-        //if we are not running in TeamCity, then we need to find the latest local tag and use that if it exists 
-        if (!TeamCityDetection.IsRunningInTeamCity())
-        {
-            var imageLoader = new DockerImageLoader(KubernetesTestsGlobalContext.Instance.TemporaryDirectory, KubernetesTestsGlobalContext.Instance.Logger, kindExePath);
-            KubernetesTestsGlobalContext.Instance.TentacleImageAndTag = imageLoader.LoadMostRecentImageIntoKind(installer.ClusterName);
-        }
-        else
-        {
-            var tag = Environment.GetEnvironmentVariable("KubernetesAgentTests_ImageTag");
-            KubernetesTestsGlobalContext.Instance.TentacleImageAndTag = $"docker.packages.octopushq.com/octopusdeploy/kubernetes-tentacle:{tag}";
-        }
-
-        if (KubernetesTestsGlobalContext.Instance.TentacleImageAndTag is not null)
-        {
-            KubernetesTestsGlobalContext.Instance.Logger.Information("Using tentacle image: {ImageAndTag}", KubernetesTestsGlobalContext.Instance.TentacleImageAndTag);
-        }
-
+        KubernetesTestsGlobalContext.Instance.TentacleImageAndTag = GetTentacleImageAndTag(kindExePath);
         KubernetesTestsGlobalContext.Instance.SetToolExePaths(helmExePath, kubeCtlPath);
         KubernetesTestsGlobalContext.Instance.KubeConfigPath = installer.KubeConfigPath;
+    }
+
+    string? GetTentacleImageAndTag(string kindExePath)
+    {
+        //By default, we don't override the values in the helm chart. This is useful if you are just writing new tests and not changing Tentacle code.
+        string? imageAndTag = null;
+        if (TeamCityDetection.IsRunningInTeamCity())
+        {
+            //In TeamCity, use the tag of the currently building code
+            var tag = Environment.GetEnvironmentVariable("KubernetesAgentTests_ImageTag");
+            imageAndTag = $"docker.packages.octopushq.com/octopusdeploy/kubernetes-tentacle:{tag}";
+        }
+        else if(bool.TryParse(Environment.GetEnvironmentVariable("KubernetesAgentTests_UseLatestLocalImage"), out var useLocal) && useLocal)
+        {
+            //if we should use the latest locally build image, load the tag from docker and load it into kind
+            var imageLoader = new DockerImageLoader(KubernetesTestsGlobalContext.Instance.TemporaryDirectory, KubernetesTestsGlobalContext.Instance.Logger, kindExePath);
+            imageAndTag = imageLoader.LoadMostRecentImageIntoKind(installer.ClusterName);
+        }
+        
+        if(imageAndTag is not null)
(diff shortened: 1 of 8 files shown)


source/Octopus.Tentacle.Kubernetes.Tests.Integration/Tooling/KubeCtlTool.cs
```diff
@@ -0,0 +1,64 @@
+﻿using System;
+using System.Text;
+using Octopus.Tentacle.CommonTestUtils;
+using Octopus.Tentacle.CommonTestUtils.Logging;
+using Octopus.Tentacle.Util;
+
+namespace Octopus.Tentacle.Kubernetes.Tests.Integration.Tooling;
+
+public class KubeCtlTool
```
APErebus, line 9: Helper class for interacting with `kubectl` to execute commands

source/Octopus.Tentacle.Kubernetes.Tests.Integration/KubernetesClusterOneTimeSetUp.cs
```diff
@@ -17,25 +17,32 @@ public async Task OneTimeSetUp()
...
+        if (TeamCityDetection.IsRunningInTeamCity())
         {
+            //In TeamCity, use the tag of the currently building code
             var tag = Environment.GetEnvironmentVariable("KubernetesAgentTests_ImageTag");
-            KubernetesTestsGlobalContext.Instance.TentacleImageAndTag = $"docker.packages.octopushq.com/octopusdeploy/kubernetes-tentacle:{tag}";
+            imageAndTag = $"docker.packages.octopushq.com/octopusdeploy/kubernetes-tentacle:{tag}";
         }
-
-        if (KubernetesTestsGlobalContext.Instance.TentacleImageAndTag is not null)
+        else if(bool.TryParse(Environment.GetEnvironmentVariable("KubernetesAgentTests_UseLatestLocalImage"), out var useLocal) && useLocal)
```
APErebus, line 35: Now if you want to use a locally built image, you need to set the environment variable `KubernetesAgentTests_UseLatestLocalImage` to `true`

source/Octopus.Tentacle.Kubernetes.Tests.Integration/KubernetesScriptServiceV1IntegrationTest.cs
```diff
@@ -46,6 +53,52 @@ sleep 0.1
...
+        var result = await TentacleClient.ExecuteScript(command, StatusReceived, ScriptCompleted, new InMemoryLog(), CancellationToken.None);
+
+        //Assert
+        logs.Should().Contain(po => po.Source == ProcessOutputSource.StdOut && po.Text == "Hello World");
+        scriptCompleted.Should().BeTrue();
+        result.ExitCode.Should().Be(1);
+        result.State.Should().Be(ProcessState.Complete);
+
+        recordedMethodUsages.For(nameof(IAsyncClientKubernetesScriptServiceV1.StartScriptAsync)).Started.Should().Be(1);
+        recordedMethodUsages.For(nameof(IAsyncClientKubernetesScriptServiceV1.GetStatusAsync)).Started.Should().BeGreaterThan(2).And.BeLessThan(30);
```
kevjt, line 98: Where do the numbers 2 and 30 come from? Is it based on an estimate given the polling nature of get status?
Perhaps it's sufficient to just assert that it's called rather than the number of times, in case of brittleness if the polling interval changes.

source/Octopus.Tentacle.Kubernetes.Tests.Integration/KubernetesScriptServiceV1IntegrationTest.cs
```diff
@@ -1,40 +1,47 @@
...
-    echo $i
-    sleep 0.1
-done
-";
+
         var command = new ExecuteKubernetesScriptCommandBuilder(LoggingUtils.CurrentTestHash())
-            .WithScriptBody(scriptBody)
+            .WithScriptBody(script => script
+                .Print("Hello World")
+                .PrintNTimesWithDelay("Yep", 50, TimeSpan.FromMilliseconds(100)))
```
kevjt, line 44: Is it necessary to run the full 5 seconds? It can probably be shortened and still be a useful test.

source/Octopus.Tentacle.Kubernetes.Tests.Integration/KubernetesScriptServiceV1IntegrationTest.cs
```diff
@@ -59,4 +112,197 @@ Task ScriptCompleted(CancellationToken ct)
...
+        var command = new ExecuteKubernetesScriptCommandBuilder(LoggingUtils.CurrentTestHash())
+            .WithScriptBody(script => script
+                .Print("Hello World")
+                .Sleep(TimeSpan.FromSeconds(1))
+                .Print("waitingtobestopped")
+                .Sleep(TimeSpan.FromSeconds(100)))
+            .Build();
+
+        //act
+        var scriptTask = Task.Run(async () => await TentacleClient.ExecuteScript(command, StatusReceived, ScriptCompleted, new InMemoryLog(), CancellationToken.None));
```
kevjt, line 133: Would it be better to pass in the test's cancellation token here instead of using `.None`? That way if something happens to the script task (it doesn't reach the status received for whatever reason) then we won't risk blocking on the semaphoreSlim indefinitely.

On a related note, I noticed that the cancellation token source in `KubernetesAgentIntegrationTest.SetUp` isn't initialised with a timeout, I think we can set something reasonable there so tests can fail within a time frame.

kevjt (commented):
Looks good, just have a few questions

source/Octopus.Tentacle.Kubernetes.Tests.Integration/KubernetesScriptServiceV1IntegrationTest.cs
APErebus: Yeah, that's a good call. We do want it to run long enough that we exercise the pod monitor etc, but I'll drop down to 3s

source/Octopus.Tentacle.Kubernetes.Tests.Integration/KubernetesScriptServiceV1IntegrationTest.cs
APErebus: Yeah, good call! I'd forgotten we had that cancellation token 😅

source/Octopus.Tentacle.Kubernetes.Tests.Integration/KubernetesScriptServiceV1IntegrationTest.cs
APErebus: I did. It wasn't serving much of a purpose

kevjt (approved):
Nice, thanks for updating

PR #954, opened 2024-05-30
Add ServerCommsAddresses as env var to script pods

scme0:
# Background

As part of adding Kubernetes Agent support for HA clusters, we want to allow health checks to check if the agent is configured correctly with the right number of ServerCommsAddresses.

[sc-80605]

# Results

This partners with a Server change which will compare the number of nodes to the number of configured ServerCommsAddresses and fail if the number is not the same. For backwards compatibility, if the value isn't set, it will be ignored on Server.

Related server change: https://github.com/OctopusDeploy/OctopusDeploy/pull/25129
Related documentation change: https://github.com/OctopusDeploy/docs/pull/2323


kevjt (approved):
LGTM

PR #955, opened 2024-05-31
Add metrics to a config map 

rain-on:
It was decided to insert a json-blob into a config map to track each NFS event in the system.

This PR adds the interfaces required to abstract the underlying persistence mechanism from where metrics events are raised.

At this stage, it is assumed that the underlying config map will be a json array of events, however a "mapper" has been included such that the type persisted in the config map can be abstracted from the data-structure used in the business logic (i.e. business logic sees this as a list of events, the underlying json may be different).


source/Octopus.Tentacle/Diagnostics/Metrics/EventRecord.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+
+namespace Octopus.Tentacle.Diagnostics.Metrics
```
kevjt, line 3: These concepts are quite specific to Kubernetes. Can these files be moved to a `Octopus.Tentacle.Kubernetes.Diagnostics` namespace instead?

source/Octopus.Tentacle.Tests/Diagnostics/Metrics/KubernetesAgentMetricsFixture.cs
```diff
@@ -0,0 +1,124 @@
...
+
+            events.ForEach(e => sut.TrackEvent(e));
+
+            var timeDate = sut.GetLatestEventTimestamp();
+
+            timeDate.Should().BeExactly(events[1].Timestamp);
+        }
+
+        [Test]
+        public void GetLatestTimeStampPropagatesExceptionsIfUnderlyingPersistenceFails()
```
kevjt, line 97: What's the reason we want to propagate exceptions for this `GetLatestTimestamp` method but not the `TrackEvent` one?

source/Octopus.Tentacle/Kubernetes/KubernetesModule.cs
```diff
@@ -28,6 +31,16 @@ protected override void Load(ContainerBuilder builder)
             builder.RegisterType<KubernetesDirectoryInformationProvider>().As<IKubernetesDirectoryInformationProvider>().SingleInstance();
             builder.RegisterType<MemoryCache>().As<IMemoryCache>().SingleInstance();
             builder.RegisterType<MemoryCacheOptions>().As<IOptions<MemoryCacheOptions>>().SingleInstance();
+            builder.RegisterType<PersistenceProvider>()
+                .Named<IPersistenceProvider>("KubernetesAgentMetricsConfigMap")
+                .WithParameter("configMapName", "kubernetes-agent-metrics");
+            builder.RegisterType<KubernetesAgentMetrics>().AsSelf().SingleInstance()
+                .WithParameter(
+                    new ResolvedParameter(
```
kevjt, line 39: Is it necessary to use a named registration in this case given that there is only one implementation?
Also what happens if the name of the PersistenceProvider doesn't line up, will an exception be thrown at runtime when it fails to resolve?

source/Octopus.Tentacle/Diagnostics/Metrics/MapFromConfigMapToEventList.cs
```diff
@@ -0,0 +1,17 @@
+using System.Collections.Generic;
+
+namespace Octopus.Tentacle.Diagnostics.Metrics
+{
+    public class MapFromConfigMapToEventList
+    {
+        public List<EventRecord> FromConfigMap(List<EventRecord> configMapContent)
+        {
+            return configMapContent;
```
kevjt, line 9: I'm guessing this is still WIP so this class is here in case we want to transform the list into a more 'compact' representation for storage?

source/Octopus.Tentacle/Diagnostics/Metrics/PersistenceProvider.cs
```diff
@@ -0,0 +1,44 @@
...
+        readonly ISystemLog log;
+        IDictionary<string, string> ConfigMapData => metricsConfigMap.Value.Data ??= new Dictionary<string, string>();
+
+        public PersistenceProvider(string configMapName, IKubernetesConfigMapService configMapService, ISystemLog log)
+        {
+            this.configMapService = configMapService;
+            this.log = log;
+            this.configMapName = configMapName;
+            metricsConfigMap = new Lazy<V1ConfigMap>(() => configMapService.TryGet(this.configMapName, CancellationToken.None).GetAwaiter().GetResult()
+                ?? throw new InvalidOperationException($"Unable to retrieve Tentacle Configuration from config map for namespace {KubernetesConfig.Namespace}"));
```
kevjt, line 30: Does this exception get propagated out of the constructor?

source/Octopus.Tentacle/Kubernetes/KubernetesModule.cs
rain-on: Yeah - named is a bit much (YAGNI).

It started with not wanting hard-coded names in the ConfigMap - as that impedes integration testing.
Creating a named item _seemed_ to make things nicer as you can now say "KubernetesAgentMetrics depends on config-map-X" - while capability-Y depends on configh-map-y.

source/Octopus.Tentacle/Diagnostics/Metrics/MapFromConfigMapToEventList.cs
rain-on: This is a WIP.
And it depends on how much we want to go for a "normalised" form of the data.

source/Octopus.Tentacle/Diagnostics/Metrics/PersistenceProvider.cs
rain-on: No - it gets propagated when the lazy is evaluated.

Specifically - if the lazy fails on the first execution - the same exception is thrown for every attempt.

source/Octopus.Tentacle/Diagnostics/Metrics/EventRecord.cs
rain-on: Done

source/Octopus.Tentacle.Tests/Diagnostics/Metrics/KubernetesAgentMetricsFixture.cs
rain-on: :thinking: we could propogate it up -and the layer above needs to catch it and handle the error.
I figured there was a risk that this would be called from critical path code, and if it fails, we didn't want to crash the calling function.

Whereas, the "GetValue" - if it doesn't work, we need to indicate failure, as subsequent steps may not be able to be conducted - potentially could do that by returning null - but that hides the fact that that something important down below failed.

I like the GetLatestEvent throwing - and can be convinced in having the TrackEvent throwing too.

There is a slight smell with the interface. -get TimeDate is a "hollow" - it probably should return the whole event at the very least ...

source/Octopus.Tentacle/Kubernetes/KubernetesModule.cs
rain-on: updated following the informal code review to use delegate creation factories in autofac.

source/Octopus.Tentacle/Diagnostics/Metrics/MapFromConfigMapToEventList.cs
rain-on: removed this concept - we store a dictionary of dictionaries as per informal review.

installer/Octopus.Tentacle.Installer/Product.wxs
```diff
@@ -14,7 +14,7 @@
     - Product/Version      : Change this every major build
     - Product/UpgradeCode  : Never change this
   -->
-  <Product Id="*" Name="Octopus Deploy Tentacle" Language="1033" Version="8.1.1518" Manufacturer="Octopus Deploy Pty. Ltd." UpgradeCode="1B32E04F-49C2-4907-8879-A556986F7F16">
+  <Product Id="*" Name="Octopus Deploy Tentacle" Language="1033" Version="8.1.1692" Manufacturer="Octopus Deploy Pty. Ltd." UpgradeCode="1B32E04F-49C2-4907-8879-A556986F7F16">
```
scme0, line 17: I would exclude this change as it's just auto-generated and not required.

source/Octopus.Tentacle.Tests/Diagnostics/Metrics/KubernetesAgentMetricsFixture.cs
```diff
@@ -0,0 +1,128 @@
...
+
+            var eventTimestamp = DateTimeOffset.Now;
+            sut.TrackEvent("Killed", "NFS Pod", eventTimestamp);
+
+            persistenceProvider.Content.Keys.Should().ContainSingle(entry => entry.Equals("Killed"));
+            var item = JsonConvert.DeserializeObject<EventJsonEntry>(persistenceProvider.Content["Killed"]);
+            item.Should().BeEquivalentTo(new EventJsonEntry("NFS Pod", new List<DateTimeOffset>() { eventTimestamp }));
+        }
+
+        // [Test]
```
scme0, line 32: Can we potentially uncomment this and add the `[ignore]` attribute like the others?

source/Octopus.Tentacle/Kubernetes/KubernetesModule.cs
```diff
@@ -32,6 +35,13 @@ protected override void Load(ContainerBuilder builder)
             builder.RegisterType<MemoryCacheOptions>().As<IOptions<MemoryCacheOptions>>().SingleInstance();
             
             builder.RegisterGeneric(typeof(ReferenceCountingKeyedBinarySemaphore<>)).As(typeof(IKeyedSemaphore<>)).SingleInstance();
+
+            var kuberneteseAgentMetricsPersistence = "kubernetese-agent-metrics-persistence";
+            builder.Register<PersistenceProvider>(ctx => ctx.Resolve<PersistenceProvider.Factory>().Invoke(ConfigMapNames.AgentMetrics))
+                .Named<IPersistenceProvider>(kuberneteseAgentMetricsPersistence);
+            builder.Register<KubernetesAgentMetrics>(ctx => 
+                ctx.Resolve<KubernetesAgentMetrics.Factory>()
+                    .Invoke(ctx.ResolveNamed<IPersistenceProvider>(kuberneteseAgentMetricsPersistence)));
```
scme0, line 44: Nice

source/Octopus.Tentacle/Kubernetes/Diagnostics/PersistenceProvider.cs
```diff
@@ -0,0 +1,54 @@
...
+    {
+        public delegate PersistenceProvider Factory(string configMapName);
+        
+        readonly string configMapName;
+        readonly IKubernetesConfigMapService configMapService;
+        readonly Lazy<V1ConfigMap> metricsConfigMap;
+        IDictionary<string, string> ConfigMapData => metricsConfigMap.Value.Data ??= new Dictionary<string, string>();
+
+        
+
```
scme0, line 28: nit: Remove some of these new lines please :)

source/Octopus.Tentacle/Kubernetes/Diagnostics/KubernetesAgentMetrics.cs
```diff
@@ -0,0 +1,72 @@
...
+            this.log = log;
+        }
+
+        public void TrackEvent(string reason, string source, DateTimeOffset occurrence)
+        {
+            try
+            {
+                lock (persistenceProvider)
+                {
+                    var existingEvent = LoadFromPersistence(reason) ?? new EventJsonEntry(source, new List<DateTimeOffset>());
```
scme0, line 29: Maybe I'm not understanding this right but if two events come with the same reason but a different source come in, the second source will be lost?

source/Octopus.Tentacle/Kubernetes/KubernetesConfigMapService.cs
```diff
@@ -29,7 +29,7 @@ public KubernetesConfigMapService(IKubernetesClientConfigProvider configProvider
             {
                 try
                 {
-                    return await Client.CoreV1.ReadNamespacedConfigMapAsync(name, KubernetesConfig.Namespace, cancellationToken: cancellationToken);
+                    return await Client.CoreV1.ReadNamespacedConfigMapAsync(name, configProvider.Get().Namespace, cancellationToken: cancellationToken);
```
scme0, line 32: Does this need to be reverted?

source/Octopus.Tentacle/Kubernetes/Diagnostics/KubernetesAgentMetrics.cs
```diff
@@ -0,0 +1,72 @@
...
+                    Persist(reason, existingEvent);
+                }
+            }
+            catch (Exception e)
+            {
+                log.Error($"Failed to persist a kubernetes event metric, {e.Message}");
+            }
+        }
+
+
```
scme0, line 40: nit: extra line can be removed

scme0 (commented):
Nice! Looking good, just a couple of comments.

source/Octopus.Tentacle.Tests/Diagnostics/Metrics/KubernetesAgentMetricsFixture.cs
rain-on: removed

source/Octopus.Tentacle/Kubernetes/Diagnostics/KubernetesAgentMetrics.cs
rain-on: :thinking: this one could be a bug!

source/Octopus.Tentacle/Kubernetes/KubernetesModule.cs
kevjt, line 44:
```suggestion
            const string kubernetesAgentMetricsPersistence = "kubernetes-agent-metrics-persistence";
            builder.Register<PersistenceProvider>(ctx => ctx.Resolve<PersistenceProvider.Factory>().Invoke(ConfigMapNames.AgentMetrics))
                .Named<IPersistenceProvider>(kubernetesAgentMetricsPersistence);
            builder.Register<KubernetesAgentMetrics>(ctx => 
                ctx.Resolve<KubernetesAgentMetrics.Factory>()
                    .Invoke(ctx.ResolveNamed<IPersistenceProvider>(kubernetesAgentMetricsPersistence)));
```

source/Octopus.Tentacle/Kubernetes/Diagnostics/KubernetesAgentMetrics.cs
```diff
@@ -0,0 +1,93 @@
...
+            this.persistenceProvider = persistenceProvider;
+            this.log = log;
+        }
+
+        public void TrackEvent(string reason, string source, DateTimeOffset occurrence)
+        {
+            DateTimeOffset latestEvent;
+            try
+            {
+                latestEvent = GetLatestEventTimestamp();
```
kevjt, line 29: I'm not a huge fan of public methods depending on other public methods in general.
In this case if you made a separate `GetLatestEventTimestampInternal` method that has all the code of the `GetLatestEventTimestamp` method except for the lock, then you can call the internal one directly in here and only need one lock to wrap the entire `TrackEvent` method. This saves having to unlock and then relock.

kevjt (commented):
Looking good! Some minor cleanup suggestions

source/Octopus.Tentacle/Kubernetes/Diagnostics/KubernetesAgentMetrics.cs
rain-on: Not an idea I'd come across, but I kinda like it.

kevjt (approved):
neat

source/Octopus.Tentacle.Kubernetes.Tests.Integration/Support/TestSupportConfigMapService.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Net;
+using k8s;
+using k8s.Autorest;
+using k8s.Models;
+using Octopus.Diagnostics;
+
+namespace Octopus.Tentacle.Kubernetes.Tests.Integration.Support
+{
+
+    // This is a copy of the production ConfigMapService, but allows the namespace to explicitly
```
scme0, line 11:
```suggestion
    // This is a copy of the production ConfigMapService, but allows the namespace to be explicitly
```

source/Octopus.Tentacle/Kubernetes/Diagnostics/KubernetesAgentMetrics.cs
```diff
@@ -0,0 +1,100 @@
...
+                return DateTimeOffset.Parse(timeStampString!);    
+            }
+            return DateTimeOffset.MinValue;
+        }
+
+        void UpdateLatestTimestamp(DateTimeOffset newEventTime)
+        {
+            try
+            {
+                var latestEvent = GetLatestEventTimestamp();
```
scme0, line 72:
```suggestion
                var latestEvent = GetLatestEventTimestampInternal();
```

scme0: I saw the comment from Kevin but I would probably just call the public method and not have the internal one at all. Much of a muchness though 🤷

source/Octopus.Tentacle/Kubernetes/Diagnostics/PersistenceProvider.cs
```diff
@@ -0,0 +1,50 @@
...
+        readonly IKubernetesConfigMapService configMapService;
+        readonly Lazy<V1ConfigMap> metricsConfigMap;
+        IDictionary<string, string> ConfigMapData => metricsConfigMap.Value.Data ??= new Dictionary<string, string>();
+
+        public PersistenceProvider(string configMapName, IKubernetesConfigMapService configMapService)
+        {
+            this.configMapService = configMapService;
+            this.configMapName = configMapName;
+            metricsConfigMap = new Lazy<V1ConfigMap>(() => configMapService.TryGet(this.configMapName, CancellationToken.None).GetAwaiter().GetResult()
+                ?? throw new InvalidOperationException($"Unable to retrieve Tentacle Configuration from config map for namespace {KubernetesConfig.Namespace}"));
```
scme0, line 31: Maybe in a new PR but I think this should be an `AsyncLazy<>` of some kind. There is one implemented in Server - we can probably copy+paste and use it.

This would cause all the methods to become async though... But I think it's better to avoid the `GetAwaiter().GetResult()` if we can.

scme0 (approved):
LGTM - just a couple of thoughts but nothing blocking 👍

source/Octopus.Tentacle/Kubernetes/Diagnostics/KubernetesAgentMetrics.cs
rain-on: likewise - it all works :shrug:

source/Octopus.Tentacle/Kubernetes/Diagnostics/PersistenceProvider.cs
rain-on: I'll raise a ticket to do - but atm this is the same as the other configmap usage.

scme0: Ah fair enough then 👍

PR #956, opened 2024-06-01
Update Machine registrations to allow Tenant slugs to be passed in

scme0:
Also fix Tenanted Deployment Participation parsing to be case insensitive

[sc-80799]


APErebus (approved):
LGTM

PR #959, opened 2024-06-04
Add KuberentesEvent Monitor & associated classes

rain-on:
This adds a background task which is responsible for polling events (currently from all namespaces), and writing events which affect an octopus element (nfs watchdog, script pod or agent) and writing them to the metrics handler (which in turn writes the event to the metrics config map).

There are a few things it'd be good to sort out, as its not currently 100%:
1. Which timestamp do we take from an event (last, first, or eventTime) - atm just taking the newest of these
2. What parameters define the events we're interested in?
   * Currently trying to find start/kill of nfs server pod via name and reason
   * NfsWatchDogTimeout events by reason - but maybe there should be more?
      * Also - not sure if the nfs watchdog needs to put _more_ information in the timeout event to make it 100% unique.

### DIFF
--- a/installer/Octopus.Tentacle.Installer/Product.wxs
+++ b/installer/Octopus.Tentacle.Installer/Product.wxs
@@ -14,7 +14,7 @@
     - Product/Version      : Change this every major build
     - Product/UpgradeCode  : Never change this
   -->
-  <Product Id="*" Name="Octopus Deploy Tentacle" Language="1033" Version="8.1.1518" Manufacturer="Octopus Deploy Pty. Ltd." UpgradeCode="1B32E04F-49C2-4907-8879-A556986F7F16">
+  <Product Id="*" Name="Octopus Deploy Tentacle" Language="1033" Version="8.1.1810" Manufacturer="Octopus Deploy Pty. Ltd." UpgradeCode="1B32E04F-49C2-4907-8879-A556986F7F16">
     <Package InstallerVersion="200" Compressed="yes" Description="Octopus Deploy Tentacle" Platform="$(var.Platform)" InstallScope="perMachine" />
     <Media Id="1" Cabinet="Files.cab" EmbedCab="yes" />
     <Property Id="MSIFASTINSTALL" Value="3" />
--- a/installer/Octopus.Tentacle.Installer/Product.wxs
+++ b/installer/Octopus.Tentacle.Installer/Product.wxs
@@ -14,7 +14,7 @@
     - Product/Version      : Change this every major build
     - Product/UpgradeCode  : Never change this
   -->
-  <Product Id="*" Name="Octopus Deploy Tentacle" Language="1033" Version="8.1.1797" Manufacturer="Octopus Deploy Pty. Ltd." UpgradeCode="1B32E04F-49C2-4907-8879-A556986F7F16">
+  <Product Id="*" Name="Octopus Deploy Tentacle" Language="1033" Version="8.1.1518" Manufacturer="Octopus Deploy Pty. Ltd." UpgradeCode="1B32E04F-49C2-4907-8879-A556986F7F16">
     <Package InstallerVersion="200" Compressed="yes" Description="Octopus Deploy Tentacle" Platform="$(var.Platform)" InstallScope="perMachine" />
     <Media Id="1" Cabinet="Files.cab" EmbedCab="yes" />
     <Property Id="MSIFASTINSTALL" Value="3" />
--- a/installer/Octopus.Tentacle.Installer/Product.wxs
+++ b/installer/Octopus.Tentacle.Installer/Product.wxs
@@ -14,7 +14,7 @@
     - Product/Version      : Change this every major build
     - Product/UpgradeCode  : Never change this
   -->
-  <Product Id="*" Name="Octopus Deploy Tentacle" Language="1033" Version="8.1.1794" Manufacturer="Octopus Deploy Pty. Ltd." UpgradeCode="1B32E04F-49C2-4907-8879-A556986F7F16">
+  <Product Id="*" Name="Octopus Deploy Tentacle" Language="1033" Version="8.1.1795" Manufacturer="Octopus Deploy Pty. Ltd." UpgradeCode="1B32E04F-49C2-4907-8879-A556986F7F16">
     <Package InstallerVersion="200" Compressed="yes" Description="Octopus Deploy Tentacle" Platform="$(var.Platform)" InstallScope="perMachine" />
     <Media Id="1" Cabinet="Files.cab" EmbedCab="yes" />
     <Property Id="MSIFASTINSTALL" Value="3" />
--- a/installer/Octopus.Tentacle.Installer/Product.wxs
+++ b/installer/Octopus.Tentacle.Installer/Product.wxs
@@ -14,7 +14,7 @@
     - Product/Version      : Change this every major build
     - Product/UpgradeCode  : Never change this
   -->
-  <Product Id="*" Name="Octopus Deploy Tentacle" Language="1033" Version="8.1.1518" Manufacturer="Octopus Deploy Pty. Ltd." UpgradeCode="1B32E04F-49C2-4907-8879-A556986F7F16">
+  <Product Id="*" Name="Octopus Deploy Tentacle" Language="1033" Version="8.1.1803" Manufacturer="Octopus Deploy Pty. Ltd." UpgradeCode="1B32E04F-49C2-4907-8879-A556986F7F16">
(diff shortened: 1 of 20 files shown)


source/Octopus.Tentacle/Kubernetes/KubernetesEventMonitor.cs
```diff
@@ -0,0 +1,109 @@
...
+            {
+                lastEventTime = agentMetrics.GetLatestEventTimestamp();
+            }
+            catch (Exception e)
+            {
+                log.Error($"Failed to determine latest handled event. {e.Message}");
+            }
+            
+            return input.Items
+                .Where(e => e.EventTime.HasValue && e.EventTime.Value.ToUniversalTime() > lastEventTime);
```
kevjt, line 79: I'm guessing this branch hasn't been brought up to date with the latest changes in the other metrics branch.
The `e.EventTime` property cannot always be relied upon because there could be event producers in Kubernetes that do not produce precise events, opting instead to use `first/lastTimestamp` and `count`. In such scenarios we need to consider the `lastTimestamp` somehow. Even then it is tricky because my understanding is that the `EventTime` can still be used by some event producers.. which complicates the logic here because you might need to account for both.

source/Octopus.Tentacle/Kubernetes/KubernetesEventMonitor.cs
```diff
@@ -0,0 +1,84 @@
...
+                    kEvent.FirstTimestamp
+                }.Where(dt => dt.HasValue)
+                .OrderByDescending(dt => dt!.Value)
+                .FirstOrDefault();
+        }
+
+        protected virtual string GetNamespace()
+        {
+            return "blah";
+            //return KubernetesConfig.Namespace;
```
scme0, line 81: Looks like this should be reverted?

source/Octopus.Tentacle/Kubernetes/KubernetesEventMonitor.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        bool IsPodRestartDueToStaleNfs(Corev1Event kubernetesEvent)
+        {
+            return kubernetesEvent.Reason == "NfsWatchdogTimeout";
+        }
+
+        bool IsNfsPodRestart(Corev1Event kubernetesEvent)
+        {
+            var podLifecycleEventsOfInterest = new []{"Started", "Killing"};
+            //TODO(tmm): having this magic event-name stored as a constant somewhere would be great.
```
scme0, line 62: Do you want to do this in this PR?

source/Octopus.Tentacle/Kubernetes/KubernetesModule.cs
```diff
@@ -42,6 +42,9 @@ protected override void Load(ContainerBuilder builder)
             builder.Register<KubernetesAgentMetrics>(ctx => 
                 ctx.Resolve<KubernetesAgentMetrics.Factory>()
                     .Invoke(ctx.ResolveNamed<IPersistenceProvider>(kubernetesAgentMetricsPersistence)));
+            
+            builder.RegisterType<KubernetesEventMonitorTask>().As<IBackgroundTask>().SingleInstance();
```
scme0, line 46: Because this is registered in the `KubernetesModule` it will only run with the Kubernetes agent which is what we want. 👍

scme0 (approved):
Nice, this is looking good!

source/Octopus.Tentacle/Kubernetes/KubernetesEventMonitor.cs
```diff
@@ -0,0 +1,84 @@
...
+                {
+                    var eventTimestamp = GetLatestTimestampInEvent(kEvent)!.Value.ToUniversalTime();
+                    agentMetrics.TrackEvent(kEvent.Reason, kEvent.Name(), eventTimestamp);
+                }
+            }
+        }
+
+        bool IsRelevantForMetrics(Corev1Event kubernetesEvent)
+        {
+            return IsNfsPodRestart(kubernetesEvent) || IsPodRestartDueToStaleNfs(kubernetesEvent);
```
kevjt, line 51: Do we also want to track Tentacle pod restarts? That can help us with issues that aren't necessarily NFS related.

source/Octopus.Tentacle/Kubernetes/KubernetesModule.cs
rain-on: I literally just worked that out :D

source/Octopus.Tentacle/Kubernetes/KubernetesEventMonitor.cs
rain-on: I don't have a good answer at this stage, so happy to leave a TODO .... what's your preference?

source/Octopus.Tentacle/Kubernetes/KubernetesEventMonitor.cs
rain-on: yeah - that's probably smart.

source/Octopus.Tentacle/Kubernetes/KubernetesEventMonitor.cs
scme0: Nah it's ok, leave it as a TODO. we'll sort out little things like this in a subsequent PR.

source/Octopus.Tentacle/Kubernetes/KubernetesEventMonitor.cs
rain-on: yes. true.

installer/Octopus.Tentacle.Installer/Product.wxs
```diff
@@ -14,7 +14,7 @@
     - Product/Version      : Change this every major build
     - Product/UpgradeCode  : Never change this
   -->
-  <Product Id="*" Name="Octopus Deploy Tentacle" Language="1033" Version="8.1.1518" Manufacturer="Octopus Deploy Pty. Ltd." UpgradeCode="1B32E04F-49C2-4907-8879-A556986F7F16">
```
scme0, line 17: Revert this :)

source/Octopus.Tentacle/Kubernetes/KubernetesEventMonitorTask.cs
```diff
@@ -0,0 +1,53 @@
...
+
+namespace Octopus.Tentacle.Kubernetes
+{
+    public class KubernetesEventMonitorTask : BackgroundTask
+    {
+        public delegate KubernetesEventMonitorTask Factory(IKubernetesEventMonitor eventMonitor);
+        
+        readonly IKubernetesEventMonitor eventMonitor;
+        readonly ISystemLog log;
+        readonly TimeSpan taskInterval = TimeSpan.FromMinutes(5);
```
scme0, line 16: Only 5 mins? should it be 10?

source/Octopus.Tentacle/Kubernetes/KubernetesEventMonitorTask.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            if (!KubernetesConfig.MetricsIsEnabled)
+            {
+                log.Info("Event monitoring for agent metrics is not enabled.");
+                return;
+            }
+            
+            //We don't want the monitoring to ever stop
+            var policy = Policy.Handle<Exception>().WaitAndRetryForeverAsync(
+                retry => TimeSpan.FromMinutes(5),
```
scme0, line 33: Maybe use the `taskInterval` value here. It makes sense because it doesn't matter if it's successful or failure, we always want to wait <taskInterval> time before trying again.

source/Octopus.Tentacle/Kubernetes/EventMapper.cs
```diff
@@ -0,0 +1,59 @@
...
+            if (kEvent.Reason.Equals(expectedReason))
+            {
+                var eventTimestamp = EventHelpers.GetLatestTimestampInEvent(kEvent)!.Value.ToUniversalTime();
+                return new EventRecord(expectedReason, source, eventTimestamp);
+            }
+            return null;
+        }
+    }
+    
+    public class AgentKilledEventMapper : IEventMapper
```
scme0, line 30: Maybe `TentacleKilledEventMapper`? I try to think of the Agent as the sum of the parts and the Tentacle is the component we're monitoring in this method.

scme0 (approved):
This is looking awesome! Nice work

Just a couple of nitpicky ones, but nothing blocking from me.

installer/Octopus.Tentacle.Installer/Product.wxs
rain-on: Done.
I hate this file.

PR #960, opened 2024-06-04
Kubernetes agent - Allow basic raw scripts with no dependencies to run without a volume mount

scme0:
# Background

There is a known issue with Server running the upgrade script on the Kubernetes agent where if the helm chart has changed in such a way that an NFS Pod restart is required, the script pod doing the upgrade will be forcefully killed by the nfs watchdog. This is reported to server as a failure so the upgrade fails and there is no follow up health check to get the latest version details. The upgrade itself might actually succeed, but because the task fails in server, it requires the user to do a health check before they can use it.

# Results

~To just fix the issue above, we could just disable the NFS watchdog for scripts that don't need it (raw scripts with no dependencies). However thinking about this holistically, I thought it might be even better to remove the volume mount requirement all together for those pods. The only hurdle for that is how to get the files onto the pod if it we can't use the volume mount. It turns out we can use `tar` with `Client.NamespacedPodExecAsync()` as per this example: https://github.com/kubernetes-client/csharp/pull/962~

After discussion with the team, we've opted for a lighter-weight solution which still acts as you would expect. We're using an init container which mounts the volume and then copies the files to an empty dir shared with the main container. Then the main container runs completely without a mount. This way we don't have to send files manually using the approach above but it still makes sense because the main container does not rely on the volume mount once the init container is complete.

https://github.com/OctopusDeploy/OctopusDeploy/pull/25256
~https://github.com/OctopusDeploy/helm-charts/pull/173~ <- this is no longer necessary with the new approach.

[sc-76765]

### DIFF
--- a/source/Octopus.Tentacle.Contracts/KubernetesScriptServiceV1/StartKubernetesScriptCommandV1.cs
+++ b/source/Octopus.Tentacle.Contracts/KubernetesScriptServiceV1/StartKubernetesScriptCommandV1.cs
@@ -18,7 +18,7 @@ namespace Octopus.Tentacle.Contracts.KubernetesScriptServiceV1
             string? scriptPodServiceAccountName,
             Dictionary<ScriptType, string>? additionalScripts,
             ScriptFile[]? additionalFiles,
-            bool isRawScriptWithNoDependencies)
+            bool readonlyWorkspaceOnly)
         {
             Arguments = arguments;
             TaskId = taskId;
@@ -29,7 +29,7 @@ namespace Octopus.Tentacle.Contracts.KubernetesScriptServiceV1
             IsolationMutexName = isolationMutexName;
             PodImageConfiguration = podImageConfiguration;
             ScriptPodServiceAccountName = scriptPodServiceAccountName;
-            IsRawScriptWithNoDependencies = isRawScriptWithNoDependencies;
+            ReadonlyWorkspaceOnly = readonlyWorkspaceOnly;
 
             if (additionalFiles != null)
                 Files.AddRange(additionalFiles);
@@ -43,7 +43,7 @@ namespace Octopus.Tentacle.Contracts.KubernetesScriptServiceV1
             }
         }
 
-        public bool IsRawScriptWithNoDependencies { get; }
+        public bool ReadonlyWorkspaceOnly { get; }
         public string ScriptBody { get; }
         public string TaskId { get; }
         public ScriptTicket ScriptTicket { get; }
--- a/source/Octopus.Tentacle.Contracts/KubernetesScriptServiceV1/StartKubernetesScriptCommandV1.cs
+++ b/source/Octopus.Tentacle.Contracts/KubernetesScriptServiceV1/StartKubernetesScriptCommandV1.cs
@@ -17,7 +17,8 @@ namespace Octopus.Tentacle.Contracts.KubernetesScriptServiceV1
             PodImageConfigurationV1? podImageConfiguration, 
             string? scriptPodServiceAccountName,
             Dictionary<ScriptType, string>? additionalScripts,
-            ScriptFile[]? additionalFiles)
+            ScriptFile[]? additionalFiles,
+            bool isRawScriptWithNoDependencies)
         {
             Arguments = arguments;
             TaskId = taskId;
@@ -28,6 +29,7 @@ namespace Octopus.Tentacle.Contracts.KubernetesScriptServiceV1
             IsolationMutexName = isolationMutexName;
             PodImageConfiguration = podImageConfiguration;
             ScriptPodServiceAccountName = scriptPodServiceAccountName;
+            IsRawScriptWithNoDependencies = isRawScriptWithNoDependencies;
 
             if (additionalFiles != null)
                 Files.AddRange(additionalFiles);
@@ -41,6 +43,7 @@ namespace Octopus.Tentacle.Contracts.KubernetesScriptServiceV1
             }
         }
 
+        public bool IsRawScriptWithNoDependencies { get; }
         public string ScriptBody { get; }
         public string TaskId { get; }
         public ScriptTicket ScriptTicket { get; }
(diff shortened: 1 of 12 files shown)


source/Octopus.Tentacle.Contracts/KubernetesScriptServiceV1/StartKubernetesScriptCommandV1.cs
```diff
@@ -17,7 +17,8 @@ public class StartKubernetesScriptCommandV1
             PodImageConfigurationV1? podImageConfiguration, 
             string? scriptPodServiceAccountName,
             Dictionary<ScriptType, string>? additionalScripts,
-            ScriptFile[]? additionalFiles)
+            ScriptFile[]? additionalFiles,
+            bool isRawScript)
```
kevjt, line 21: It's understandable that adding a bool flag to this contract is backwards compatible. However, I would like to float the idea of splitting out the rawscript into its own class, something like "StartKubernetesRawScriptCommand". Having a separate class reduces the complexity in both, as there might not be as many optional inputs as well as fewer bool flags.

source/Octopus.Tentacle/Kubernetes/KubernetesScriptPodCreator.cs
```diff
@@ -225,17 +215,70 @@ await CreateScriptContainer(command, workspace, podName, scriptName, homeDir)
 
             var createdPod = await podService.Create(pod, cancellationToken);
             podMonitor.AddPendingPod(command.ScriptTicket, createdPod);
-
             LogVerboseToBothLogs($"Executing script in Kubernetes Pod '{podName}'.", tentacleScriptLog);
         }
 
+        async Task<IList<V1Container>> CreateInitContainers(StartKubernetesScriptCommandV1 command, string podName, string homeDir, string workspacePath)
+        {
+            if (!command.IsRawScript)
```
kevjt, line 223: Analogous to splitting the contract, this class here can be split into two separate implementations for the `IKubernetesScriptPodCreator`. This means that the non raw script pod creator wouldn't need to know about init containers and the code can be simpler.

source/Octopus.Tentacle/Kubernetes/KubernetesScriptPodCreator.cs
kevjt, line 1: Would it be easy to add an integration test that covers raw script commands?

kevjt (commented):
Added a few suggestions to consider, otherwise looks pretty good.
Probably still worth getting a review from someone with a good understanding of Kubernetes init containers 😄

source/Octopus.Tentacle.Contracts/KubernetesScriptServiceV1/StartKubernetesScriptCommandV1.cs
APErebus: @kevjt I tend to agree, although contract versioning is a tricky problem. If we did make a new contract, we'd need to probably do a major version of the agent, which then gets tricky.

Due to this, I think it's pragmatic to just add the extra boolean flag as an optional, additional field.

If/when we need a hard, contract breaking version of the agent, we can assess this then.

source/Octopus.Tentacle/Kubernetes/KubernetesScriptPodCreator.cs
APErebus: @scme0 Yeah, I think this is a good idea. if there is some shared logic, then we can have an `abstract` `KubernetesScriptPodCreator` that they both inherit from

source/Octopus.Tentacle/Kubernetes/KubernetesScriptPodCreator.cs
APErebus: Yeah, this we should definitely have cc @scme0

scme0: Done 👍

source/Octopus.Tentacle/Kubernetes/KubernetesScriptPodCreator.cs
scme0: I've just made a new class `KubernetesRawScriptPodCreator` which inherits from `KubernetesScriptPodCreator` directly. I had to make a few methods `protected virtual` and then I could override them in the raw script pod creator to make it work as expected. Let me know what you think.

source/Octopus.Tentacle.Kubernetes.Tests.Integration/KubernetesScriptServiceV1IntegrationTest.cs
```diff
@@ -241,20 +272,27 @@ Task ScriptCompleted(CancellationToken ct)
     }
 
     [Test]
-    public async Task WhenALongRunningScriptIsCancelled_TheScriptShouldStop()
+    [TestCase(ScriptType.Normal)]
+    [TestCase(ScriptType.Raw)]
```
scme0, line 276: I've made all these tests run with both a raw script and a normal script but this makes the tests run quite a bit longer in teamcity. Is this ok? @APErebus

APErebus: Yep, that's fine.

These are no way the slowest tests we have, so I'm fine with that

source/Octopus.Tentacle/Kubernetes/KubernetesPodService.cs
```diff
@@ -1,11 +1,11 @@
 ﻿using System;
 using System.IO;
 using System.Linq;
-using System.Net;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using ICSharpCode.SharpZipLib.Tar;
```
APErebus, line 7: No needed anymore?

APErebus (approved):

source/Octopus.Tentacle/Kubernetes/KubernetesPodService.cs
scme0: Ah yeah, good catch. Thanks!

PR #962, opened 2024-06-11
Use Artifactory instead of dockerhub to source helm chart

scme0:
# Background

We've been hitting rate limiting because docker is tightening up their rate limits.

# Results

Instead of doing what we do in server which is to return inconclusive when dockerhub rate limits our request, we are just going to use Artifactory. This will eliminate the rate limiting but could cause failures if artifactory is down. Because this test doesn't have to run on the same scale as server, we don't need to stress too much about this. If it does fail, we can just rerun the test.


liam-mackie (approved):
LGTM! 🚀

PR #967, opened 2024-06-21
Call openssl rehash in Kubernetes agent startup script

APErebus:
# Background

To support customers with custom certificates, a new helm-chart PR has been created
https://github.com/OctopusDeploy/helm-charts/pull/188

Debian requires that the certificate be hashed (which creates a symlink to the original file, but with a hashed filename) to be used.

Rather than stuff around with init-containers or lifecycle events, we are including this rehash at the start of the `configure-and-run.sh` script as it's fast and safe.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


liam-mackie (approved):
LGTM!

PR #969, opened 2024-06-25
Add command to support registering the Kubernetes tentacle as an Octopus worker

kevjt:
# Background

We would like to support running the Kubernetes Agent as a worker. This PR embodies the proposed solution in this [Tentacle RFC](https://docs.google.com/document/d/1TrzRaOznEbVsruJdCOLBKTvccVxRM-YttEhYGFjLxIA).

[sc-82021]

### Associated PRs
- [Helm chart](https://github.com/OctopusDeploy/helm-charts/pull/196)
- [Octopus.Client](https://github.com/OctopusDeploy/OctopusClients/pull/858) (dependency)

# Results

- Introduces a `RegisterKubernetesWorkerCommand`, following a similar pattern to how the agent is currently registered as a deployment target
- Deprecates the existing `RegisterKubernetesClusterCommand` and rebrands it as `RegisterKubernetesDeploymentTargetCommand` as we want to be clear in the distinction between the two (the functionality is preserved)
- Modified `configure-and-run.sh` script for the Kubernetes agent to optionally select from either worker or deployment target (but not both)


source/Octopus.Tentacle/Octopus.Tentacle.csproj
```diff
@@ -52,7 +52,7 @@
 		<PackageReference Include="KubernetesClient" Version="13.0.26" />
 	</ItemGroup>
 	<ItemGroup>
-		<PackageReference Include="Octopus.Client" Version="14.3.1436" />
+		<PackageReference Include="Octopus.Client" Version="14.3.1467-kevjt-kubernetes-agent-as-a-worker-20240624172217" />
```
kevjt, line 55: This will be updated once the draft PRs finalise and we get a proper release of the Octopus.Client changes

scme0 (commented):
This looks good to me and I couldn't find anything to comment on but I'll hold off on approval until the official release of `Octopus.Client` is in here and the PR is out of draft.

docker/kubernetes-agent-tentacle/scripts/configure-and-run.sh
```diff
@@ -108,15 +141,19 @@ function validateVariables() {
...
   if [[ -n "$DefaultNamespace" ]]; then
     echo " - default namespace '$DefaultNamespace'"
   fi
-  if [[ -n "$TentacleCertificateBase64" ]]; then
-    echo " - tentacle certificate '${TentacleCertificateBase64:0:3}...${TentacleCertificateBase64: -3}'"
+}
+
+function validateWorkerVariables() {
+  if [[ -z "$WorkerPools" ]]; then
+    echo "Please specify one or more worker pool names (comma delimited) with the 'WorkerPools' environment variable" >&2
```
rain-on, line 152: should this say "names/ids/slugs"?

rain-on (commented):
one nit :) Nice work.
Will let SaST review and approve.

docker/kubernetes-agent-tentacle/scripts/configure-and-run.sh
```diff
@@ -47,6 +47,28 @@ function getPublicHostName() {
 }
 
 function validateVariables() {
+  validateCommonVariables
+
+  if [[ "$DeploymentTargetEnabled" != "true" && "$WorkerEnabled" != "true" ]]; then
```
LukeButters, line 52: For my understanding: Is it that neither of these existed before, and now these do exist. Could the requirement of these variables break a customer or is it that the helm chart always takes care of setting one of these and so no customer can be impacted?

source/Octopus.Tentacle/Commands/ClientModule.cs
```diff
@@ -11,7 +11,11 @@ protected override void Load(ContainerBuilder builder)
         {
             base.Load(builder);
             builder.RegisterType<OctopusClientFactory>().AsSelf().AsImplementedInterfaces();
+#pragma warning disable CS0618 // Type or member is obsolete
```
LukeButters, line 14: It might be worth noting why this is obsolete, and if/when it could be removed or if instead it must remain forever for some reason.

source/Octopus.Tentacle/Commands/RegisterKubernetesClusterCommand.cs
```diff
@@ -9,6 +9,7 @@
 
 namespace Octopus.Tentacle.Commands
 {
+    [Obsolete("This class is being deprecated in favour of RegisterKubernetesDeploymentTargetCommand")]
```
LukeButters, line 12: If possible make a note around when this could be deleted or of it needs to stay around forever.

source/Octopus.Tentacle/Commands/RegisterWorkerCommand.cs
```diff
@@ -11,11 +11,18 @@
...
-    public class RegisterWorkerCommand : RegisterMachineCommandBase<IRegisterWorkerOperation>
+    public class RegisterWorkerCommand : RegisterWorkerCommand<IRegisterWorkerOperation>
     {
-        readonly List<string> workerpoolNames = new List<string>();
+        public RegisterWorkerCommand(Lazy<IRegisterWorkerOperation> lazyRegisterMachineOperation, Lazy<IWritableTentacleConfiguration> configuration, ISystemLog log, IApplicationInstanceSelector selector, Lazy<IOctopusServerChecker> octopusServerChecker, IProxyConfigParser proxyConfig, IOctopusClientInitializer octopusClientInitializer, ISpaceRepositoryFactory spaceRepositoryFactory, ILogFileOnlyLogger logFileOnlyLogger) : base(lazyRegisterMachineOperation, configuration, log, selector, octopusServerChecker, proxyConfig, octopusClientInitializer, spaceRepositoryFactory, logFileOnlyLogger)
+        {
+        }
+    }
+    
+    public class RegisterWorkerCommand<TRegisterWorkerOperation> : RegisterMachineCommandBase<TRegisterWorkerOperation> where TRegisterWorkerOperation : IRegisterWorkerOperation
```
LukeButters, line 21: Should this be `RegisterWorkerCommandBase`? and moved into its own file?

docker/kubernetes-agent-tentacle/scripts/configure-and-run.sh
```diff
@@ -246,15 +261,62 @@ function registerTentacle() {
...
+
+  if [[ -n "$WorkerPools" ]]; then
+    IFS=',' read -ra WORKERPOOLS <<<"$WorkerPools"
+    for i in "${WORKERPOOLS[@]}"; do
+      ARGS+=('--workerpool' "$i")
+    done
+  fi
+
+  echo "${ARGS[@]}"
 }
```
LukeButters, line 320: Nit: This is an impressive bash script. I wonder if it can be pulled apart so that it may be tested and make future changes to it easier.

LukeButters (approved):
Approved although it is not clear to me if this can be merged given the RFC itself is not approved.

source/Octopus.Tentacle.Tests/Commands/RegisterWorkerCommandFixture.cs
```diff
@@ -88,7 +88,7 @@ public void ShouldRegisterListeningWorker()
                 "--force",
                 "--proxy=Proxy");
 
-            Assert.That(operation.WorkerPoolNames.Single(), Is.EqualTo("SomePool"));
+            Assert.That(operation.WorkerPools.Single(), Is.EqualTo("SomePool"));
```
LukeButters, line 91: @rain-on it would make sense and I think is a requirement that the new k8s command is tested similar to this in this repo. (on top of the E2E test in Octopus Deploy)

docker/kubernetes-agent-tentacle/scripts/configure-and-run.sh
kevjt: Good question. We are planning to move to a new [v2 Helm chart](https://github.com/OctopusDeploy/helm-charts/pull/196) which will be setting these env variables. We will also be [updating](https://github.com/OctopusDeploy/helm-charts/pull/220) the current Helm chart (v1) to set these variables so that it can continue using the latest Tentacle versions.

docker/kubernetes-agent-tentacle/scripts/configure-and-run.sh
kevjt: Yea, agreed that this bash script has gotten a bit big. I think it'd be ideal to have less logic in bash and more in C# so it can actually be tested. The script itself is already broken into various functions so I don't think pulling it apart (assuming you meant into separate files) will make it much easier to work with 🤔. Having said that, we will definitely plan to reduce the size of our bash scripts, and rely on scripts that are testable 👍

source/Octopus.Tentacle.Tests/Commands/RegisterWorkerCommandFixture.cs
kevjt: Added the unit tests.

With regards to the E2E tests in Octopus, I am in the process of updating them to cover the Tentacle changes here (using the branch build). Will share the Teamcity link to those once they are available.

LukeButters (approved):

kevjt:
The Octopus E2E tests that cover this change are in this [PR](https://github.com/OctopusDeploy/OctopusDeploy/pull/26342)
[Teamcity test results](https://build.octopushq.com/buildConfiguration/OctopusDeploy_OctopusServer_TestOrchestration_OrchestrationTestE2eLinux/13112691?buildTab=tests&name=kubernetesagent) 

The tests cover both usage as worker (new feature) and deployment target (regression testing).

PR #973, opened 2024-07-02
pull docker image from artifactory

michaelongso:
# Background

Dockerhub is no longer providing rate-limit exemptions for Azure-hosted services. ([slack thread ref](https://octopusdeploy.slack.com/archives/CDANN5QLT/p1719890200247119)) This leads to failures on the **Test: Linux Packages**.

The [team discussed](https://octopusdeploy.slack.com/archives/C01HH8T16G3/p1719897407362199?thread_ts=1719894614.969379&cid=C01HH8T16G3) and to unblock the build pipeline, we will use Octopus artifactory as a proxy for the docker images needed by the build.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

- [ ] I have read [How we use GitHub Issues](https://github.com/OctopusDeploy/Issues/blob/main/docs/CONTRIBUTING.internal.md) for help deciding when and where it's appropriate to make an issue.
- [ ] I have considered informing or consulting the right people, according to the [ownership map](https://whimsical.com/ownership-map-NzbiD4HJyvhC9jNJNfS6TG).
- [ ] I have considered appropriate testing for my change.

PR #974, opened 2024-07-02
Update k8s agent tests to pull from Artifactory

APErebus:
# Background

Similar to the change made in #973, the kubernetes integration tests are intermittently failing due to docker rate limits.

After discussing with the builds team, they let us know that artifactory will proxy and cache the container pull to docker hub if the container is not in artifactory.

# Results

Updated the agent values file to change all the default container image repositories to be artifactory.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

- [ x] I have considered informing or consulting the right people, according to the [ownership map](https://whimsical.com/ownership-map-NzbiD4HJyvhC9jNJNfS6TG).

### DIFF
--- a/source/Octopus.Tentacle.Kubernetes.Tests.Integration/Setup/agent-values.yaml
+++ b/source/Octopus.Tentacle.Kubernetes.Tests.Integration/Setup/agent-values.yaml
@@ -8,6 +8,17 @@ agent:
   targetEnvironments: ["development"]
   targetRoles: ["testing-cluster"]
   
+  image:
+    repository: docker.packages.octopushq.com/octopusdeploy/kubernetes-agent-tentacle
+  
+persistence:
+  nfs:
+    image:
+      repository: docker.packages.octopushq.com/octopusdeploy/nfs-server
+    watchdog:
+      image:
+        repository: docker.packages.octopushq.com/octopusdeploy/kubernetes-agent-nfs-watchdog  
+  
 testing:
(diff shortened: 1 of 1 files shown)


source/Octopus.Tentacle.Kubernetes.Tests.Integration/Setup/agent-values.yaml
```diff
@@ -8,6 +8,17 @@ agent:
   targetEnvironments: ["development"]
   targetRoles: ["testing-cluster"]
   
+  image:
+    repository: docker.packages.octopushq.com/octopusdeploy/kubernetes-agent-tentacle
```
scme0, line 12: Do you need the `/octopusdeploy` bit?

APErebus: yes, I believe we do...

https://github.com/OctopusDeploy/OctopusTentacle/blob/main/source/Octopus.Tentacle.Kubernetes.Tests.Integration/KubernetesClusterOneTimeSetUp.cs#L33

scme0: Ah because we are using Artifactory as a dockerhub cache I guess?

I suppose if it doesn't work the tests will fail so I'll approve 👍

scme0 (approved):

PR #977, opened 2024-07-09
Support configuring script pod resource requirements

APErebus:
# Background

A customer appears to be having resource contention with a large number of script pods. They would like to be able to modify the resource requirements of script pods.

# Results

Adds support for a new environment variable `OCTOPUS__K8STENTACLE__PODRESOURCEJSON` which contains the serialized resource requirements provided by the helm chart in https://github.com/OctopusDeploy/helm-charts/pull/225

We then deserialize this to a `V1ResourceRequirements` and use it

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle/Kubernetes/KubernetesScriptPodCreator.cs
+++ b/source/Octopus.Tentacle/Kubernetes/KubernetesScriptPodCreator.cs
@@ -197,7 +197,7 @@ namespace Octopus.Tentacle.Kubernetes
                 Spec = new V1PodSpec
                 {
                     InitContainers = await CreateInitContainers(command, podName, homeDir, workspacePath),
-                    Containers = await CreateScriptContainers(command, podName, scriptName, homeDir, workspacePath, workspace.ScriptArguments),
+                    Containers = await CreateScriptContainers(command, podName, scriptName, homeDir, workspacePath, workspace.ScriptArguments, tentacleScriptLog),
                     ImagePullSecrets = imagePullSecretNames,
                     ServiceAccountName = serviceAccountName,
                     RestartPolicy = "Never",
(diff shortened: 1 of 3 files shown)


source/Octopus.Tentacle/Kubernetes/KubernetesScriptPodCreator.cs
```diff
@@ -284,14 +288,33 @@ protected async Task<V1Container> CreateScriptContainer(StartKubernetesScriptCom
...
-                    {
-                        ["cpu"] = new("25m"),
-                        ["memory"] = new("100Mi")
-                    }
+                    return KubernetesJson.Deserialize<V1ResourceRequirements>(json);
+                }
+                catch (Exception e)
+                {
+                    //if we can't parse the JSON, fall back to the defaults below and warn the user
+                    log.WarnFormat(e, $"Failed to deserialize env.{KubernetesConfig.PodResourceJsonVariableName} into valid pod resource requirements. Using default values. Value: {json}");
```
kevjt, line 307:
```suggestion
                    log.WarnFormat(e, $"Failed to deserialize env.{KubernetesConfig.PodResourceJsonVariableName} into valid pod resource requirements.\nJSON: {json}\nUsing default values.");
```
Perhaps something like that suggestion to make it clear that the `json` is the thing that cannot be deserialised rather than the default values being applied?

kevjt (approved):
Looks good! One minor comment.

liam-mackie (approved):
LGTM! Thanks for the 🍐 on the review.
For posterity - we talked about validating the resource requests and limits, but there was no easy way to do this in the C# k8s client. In addition, the error returned when this happens is verbose and points to the issue fairly well.

liam-mackie (approved):
LGTM! Thanks for the 🍐 on the review.
For posterity - the test should pass once the tentacle PR is merged and the build is available on dockerhub.

PR #980, opened 2024-07-22
Define pod container image in env var

rain-on:
When Tentacle is running in kubernetes, it is required to be informed of the container image to be used when spawning the script pod.

The script-pods:
* repository
* tag
* Pullpolicy

Are to be defined by environment variables.

If the script-pod repository is not specifed
* Will default to using the kubernetes-agent-tools-base image, whose version matches that of the kubernetes being run (eg 1.3.0)
* If an image is specified - it will used the specified image, and will use the supplied tag if supplied - otherwise "latest"
* If pull policy is not defined, will fall back to kuberentes api standard, otherwise will override with value in env-var.


source/Octopus.Tentacle/Kubernetes/KubernetesRawScriptPodCreator.cs
```diff
@@ -39,6 +40,7 @@ protected override async Task<IList<V1Container>> CreateInitContainers(StartKube
             {
                 Name = $"{podName}-init",
                 Image = command.PodImageConfiguration?.Image ?? await containerResolver.GetContainerImageForCluster(),
+                ImagePullPolicy = KubernetesConfig.ScriptPodPullPolicy.IsNullOrEmpty() ? null : KubernetesConfig.ScriptPodPullPolicy,
```
kevjt, line 43: Can this `.IsNullOrEmpty()` logic be moved into the KubernetesConfig?

source/Octopus.Tentacle/Kubernetes/KubernetesConfig.cs
```diff
@@ -38,6 +38,10 @@ public static class KubernetesConfig
 
         public const string ServerCommsAddressesVariableName = "ServerCommsAddresses";
 
+        public static string? ScriptPodContainerImage => Environment.GetEnvironmentVariable($"{EnvVarPrefix}__SCRIPTPODIMAGE");
+        public static string ScriptPodContainerImageTag => Environment.GetEnvironmentVariable($"{EnvVarPrefix}__SCRIPTPODIMAGETAG") ?? "latest";
```
kevjt, line 42: Should this also be checking for null or empty like you have for the ScriptPodPullPolicy in the pod creator?

source/Octopus.Tentacle/Kubernetes/KubernetesRawScriptPodCreator.cs
rain-on: I actually suspect we just remove the ternary - whatever got passed in, goes here - it'll already be null.

source/Octopus.Tentacle/Kubernetes/KubernetesConfig.cs
rain-on: I saw two options - they both work:
1. Do this - such that the Pod Creator _thinks_ that the required tag has been set
2. Make this nullable - then make Pod Creator choose the default (your suggestion)

I preferred this way, as it means the "indecision" is handled upfront as a configuration - there's no need to check for null otherwise - for the code, there _is_ a known value.

I guess its about:
* Define default in the config
* Define default where config is used.

Not sure on a best practice - or if it just depends.

source/Octopus.Tentacle.Kubernetes.Tests.Integration/Setup/KubernetesAgentInstaller.cs
```diff
@@ -40,10 +40,12 @@ public KubernetesAgentInstaller(TemporaryDirectory temporaryDirectory, string he
 
     public Uri SubscriptionId { get; } = PollingSubscriptionId.Generate();
 
-    public async Task<string> InstallAgent(int listeningPort, string? tentacleImageAndTag)
+    public async Task<string> InstallAgent(int listeningPort, string? tentacleImageAndTag, IDictionary<string, string> customValues)
     {
         var valuesFilePath = await WriteValuesFile(listeningPort);
         var arguments = BuildAgentInstallArguments(valuesFilePath, tentacleImageAndTag);
+        arguments += " ";
+        arguments += ToHelmCommandArgs(customValues);
```
APErebus, line 48: A bit clunky, but it'll do.

source/Octopus.Tentacle.Kubernetes.Tests.Integration/Setup/KubernetesClusterInstaller.cs
```diff
@@ -31,7 +31,7 @@ public KubernetesClusterInstaller(TemporaryDirectory tempDirectory, string kindE
         this.kubeCtlPath = kubeCtlPath;
         this.logger = logger;
 
-        clusterName = $"tentacleint-{DateTime.Now:yyyyMMddhhmmss}";
+        clusterName = $"tentaclient-{DateTime.Now:yyyyMMddhhmmss}";
```
APErebus, line 34: ?

rain-on: hehe - I kept reading "tentacle-cleint" - but now I see its tentacle_int - changing it back!

APErebus (approved):
Excellent!

PR #982, opened 2024-07-24
Updated Kubernetes pod creation to use sh -c to bypass selinux issues

liam-mackie:
# Background
In EKS clusters utilising AWS Bottlerocket nodes, SELinux was blocking the execution of the `bootstrapRunner` on NFS. This was shown in the following logs when attempting to run a script pod:
`kubelet`:
```
"StartContainer from runtime service failed" err="rpc error: code = Unknown desc = failed to create containerd task: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: \"/octopus/Work/<task-dir>/bootstrapRunner\": permission denied: unknown"
```

`Audit`:
```
AVC avc:  denied  { execute } for  pid=<pid> comm="nfsd" name="bootstrapRunner" dev="<device>" ino=<inode> scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:data_t:s0:cxxx,cyyy tclass=file permissive=0
```

# Results

Fixes https://github.com/OctopusDeploy/Issues/issues/8935

To fix this issue, the shim just needs to execute a binary local to the container, and _then_ invoke the binary on the remote filesystem. In this case, we run `sh -c "<command>"` which ensures that the executable is being invoked in the context of the container, not kernel.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


APErebus (commented):
💭 Should we make sure we do a test of this container with an actual deployment so we get the script args being used (things like variables etc)?

APErebus:
Shortcut story: [sc-88150]

APErebus (approved):
@liam-mackie and I did some testing and verified that the scripts were being executed correctly, as well as the script arguments were correctly being passed (such as the encrypted variables password)

PR #988, opened 2024-08-01
Add ability to set script pod security contexts

APErebus:
# Background

Customers have asked if they are configure the Kubernetes agent pods to not run as the root user. K8s has this capability via the pod `securityContext` values (and `runAsUser`).

# Results

This PR exposes a new environment variable `OCTOPUS_K8STENTACLE_PODSECURITYCONTEXTJSON` that is provided by the Helm chart that is then deserialized and set on the script pod.

Shortcut story: [sc-87998]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/docker/kubernetes-agent-tentacle/Dockerfile
+++ b/docker/kubernetes-agent-tentacle/Dockerfile
@@ -49,8 +49,9 @@ RUN chgrp -R 0 /opt /usr /.dotnet && \
 
 # The same as above, but we shouldn't recursively own /etc, as it would add
 # ownership to system files like /etc/passwd
-RUN chgrp 0 /etc && \
-    chmod g=u /etc
+# We own /etc/ssl/certs so we can do a rehash
+RUN chgrp 0 /etc /etc/ssl/certs && \
+    chmod g=u /etc /etc/ssl/certs
 
 ENV BOOTSTRAPRUNNEREXECUTABLEPATH=/bootstrapRunner
 ENV OCTOPUS_RUNNING_IN_CONTAINER=Y
--- a/source/Octopus.Tentacle/Kubernetes/KubernetesScriptPodCreator.cs
+++ b/source/Octopus.Tentacle/Kubernetes/KubernetesScriptPodCreator.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Linq;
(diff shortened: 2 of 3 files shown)


source/Octopus.Tentacle/Kubernetes/KubernetesScriptPodCreator.cs
```diff
@@ -318,58 +320,58 @@ V1ResourceRequirements GetScriptPodResourceRequirements(InMemoryTentacleScriptLo
...
+                "pod tolerations");
+
+        V1PodSecurityContext? ParseScriptPodSecurityContext(InMemoryTentacleScriptLog tentacleScriptLog)
+            => ParseScriptPodJson<V1PodSecurityContext>(
+                tentacleScriptLog,
+                KubernetesConfig.PodSecurityContextJson,
+                KubernetesConfig.PodSecurityContextJsonVariableName,
+                "pod security context");
+        
+        [return: NotNullIfNotNull("defaultValue")]
```
liam-mackie, line 352: I didn't know this existed - that's neat. What happens if the Pod security context is null rather than an empty object in this case though? Will the pod creation fail?

APErebus: I'll double check, but I'm pretty confident that property is nullable right now and because we weren't setting it, was just getting `null`.

liam-mackie (approved):
Nice work! Good call on the generic `ParseScriptPodJson` now - I'm fairly sure we'll be doing this more over time.

docker/kubernetes-agent-tentacle/Dockerfile
```diff
@@ -49,8 +49,9 @@ RUN chgrp -R 0 /opt /usr /.dotnet && \
 
 # The same as above, but we shouldn't recursively own /etc, as it would add
 # ownership to system files like /etc/passwd
-RUN chgrp 0 /etc && \
-    chmod g=u /etc
+# We own /etc/ssl/certs so we can do a rehash
+RUN chgrp 0 /etc /etc/ssl/certs && \
+    chmod g=u /etc /etc/ssl/certs
```
APErebus, line 54: When running as a non-root user, we don't have permission to do the ssl cert rehash at the start. This allows for it

APErebus:
Related Helm chart pr: https://github.com/OctopusDeploy/helm-charts/pull/258

PR #992, opened 2024-08-16
Write script pod events to task log

APErebus:
# Background

We have situations where a script pod gets stuck n the Pending state or is behaving strangely. In these cases, customers have to use `kubectl` or another tool to work out what the issue is (such as a bad image name).

This PR adds functionality so that pod events are now retrieved and written to the task log as verbose messages.

# Results

When we are retrieving the pod logs, we now also make a call to the Kubernetes API to retrieve all the events that have occurred for the pod. We then filter those events to only those we haven't seen since last time we retrieved logs, then interlace them with the rest of the logs.

So we don't lose the "sinceTime" for the events if tentacle restarts, we write the current timestamp to a file in the workspace and only load if from the file if we can't find the value in memory.

If the event is a warning event, we flag it as a warning log message.
If the event is a `Pulling` or `Pulled` event for a previous `Pulling` event, we show it as the `Wait` category, to show users that we are waiting for the pull to complete before running their pods.

These events can be disabled by setting the `OCTOPUS__K8STENTACLE__DISABLEPODEVENTSINTASKLOG` env var to `true`.

Shortcut story: [sc-74374]

## After

Pod events
image

Warning events
image

Pulling/Pulled
image

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Octopus.Tentacle/Kubernetes/IKubernetesPodLogService.cs
```diff
@@ -21,39 +22,65 @@ class KubernetesPodLogService : KubernetesService, IKubernetesPodLogService
...
 
         public async Task<(IReadOnlyCollection<ProcessOutput> Outputs, long NextSequenceNumber)> GetLogs(ScriptTicket scriptTicket, long lastLogSequence, CancellationToken cancellationToken)
         {
             var tentacleScriptLog = scriptLogProvider.GetOrCreate(scriptTicket);
             var podName = scriptTicket.ToKubernetesScriptPodName();
 
-            var podLogs = await GetPodLogs();
+            var podLogsTask = GetPodLogs();
+            var podEventsTask = GetPodEvents(scriptTicket, podName, cancellationToken);
+            await Task.WhenAll(podLogsTask, podEventsTask);
```
liam-mackie, line 49: Is there any reason to await both tasks here when we await both separately below?

source/Octopus.Tentacle/Kubernetes/IKubernetesPodLogService.cs
```diff
@@ -107,5 +185,36 @@ public KubernetesPodLogService(IKubernetesClientConfigProvider configProvider, I
...
+                this.wrapper = wrapper;
+            }
+
+            public IEnumerable<ProcessOutput> Expand()
+            {
+                return new[]
+                {
+                    //we add the service messages one tick before and after so they are correctly ordered
+                    new ProcessOutput(ProcessOutputSource.StdOut, $"##octopus[stdout-{wrapper}]", Occurred.Subtract(OneTick)),
+                    new ProcessOutput(Source, Text, Occurred),
```
liam-mackie, line 206: If there are events that occur in the same microsecond, could this cause issues with service message formatting?

source/Octopus.Tentacle/Kubernetes/KubernetesConfig.cs
```diff
@@ -21,6 +21,7 @@ public static class KubernetesConfig
 
         public static TimeSpan PodsConsideredOrphanedAfterTimeSpan => TimeSpan.FromMinutes(int.TryParse(Environment.GetEnvironmentVariable($"{EnvVarPrefix}__PODSCONSIDEREDORPHANEDAFTERMINUTES"), out var podsConsideredOrphanedAfterTimeSpan) ? podsConsideredOrphanedAfterTimeSpan : 10);
         public static bool DisableAutomaticPodCleanup => bool.TryParse(Environment.GetEnvironmentVariable($"{EnvVarPrefix}__DISABLEAUTOPODCLEANUP"), out var disableAutoCleanup) && disableAutoCleanup;
+        public static bool DisablePodEventsInTaskLog => bool.TryParse(Environment.GetEnvironmentVariable($"{EnvVarPrefix}__DISABLEPODEVENTSINTASKLOG"), out var disable) && disable;
```
liam-mackie, line 24: Are we able to (and is it worth) feature flagging this somehow in the Octopus Server side instead? That way we can turn it off if we find there's an issue? I only worry since if something goes wrong here, there's very little chance of recovery, since we rely on Tentacle to run a deployment to update itself.

liam-mackie (approved):
Minor review comments (mostly about my own understanding), but otherwise LGTM! I was wondering if there's any way for us to test these events work, since it feels like a low hanging fruit, but not enough to block this merge. :shipit:

source/Octopus.Tentacle/Kubernetes/IKubernetesPodLogService.cs
APErebus: the awaits below won't actually re-run the code, it's just a quick way of getting the result (without calling `.Result)

source/Octopus.Tentacle/Kubernetes/IKubernetesPodLogService.cs
APErebus: yeah, technically... I couldn't think of a better way to do this... maybe in the same second... I'm just concerned if _some other code_ does a sort based on `Occurred`

source/Octopus.Tentacle/Kubernetes/KubernetesConfig.cs
APErebus: We _could_, but that's a lot more work as we'd need to update the contracts + add UI to set/handle it.

APErebus:
> Minor review comments (mostly about my own understanding), but otherwise LGTM! I was wondering if there's any way for us to test these events work, since it feels like a low hanging fruit, but not enough to block this merge. :shipit:

I can write a test that runs a script in a container only we care about in that test, observes that it runs the script, then re-run that script and observe that it doesn't write the pulling message a 2nd time

liam-mackie:
> > Minor review comments (mostly about my own understanding), but otherwise LGTM! I was wondering if there's any way for us to test these events work, since it feels like a low hanging fruit, but not enough to block this merge. :shipit:
> 
> 
> 
> I can write a test that runs a script in a container only we care about in that test, observes that it runs the script, then re-run that script and observe that it doesn't write the pulling message a 2nd time

I was honestly more on the side of a small unit test here to ensure that the events were able to be processed properly if we collect them, since the k8s contract for events is well known. We can likely also simply check that any deployment has an event written to the log to test that the collection part works.

source/Octopus.Tentacle/Kubernetes/IKubernetesPodLogService.cs
liam-mackie: I meant that we don't need the `await Task.WhenAll(podLogsTask, podEventsTask);`
It's nicer to only await the tasks when we process them IMO, since we'd only be waiting for the data we're actively using.

source/Octopus.Tentacle/Kubernetes/KubernetesConfig.cs
liam-mackie: Probably not worth it for this change since it's fairly trivial then, but it may be worth considering this in future.

source/Octopus.Tentacle/Kubernetes/IKubernetesPodLogService.cs
APErebus: ah, I see what you mean. Yeah. I can see it.

APErebus:
> We can likely also simply check that any deployment has an event written to the log to test that the collection part works.

I added a check to the simple integration test to verify at least _some_ pod events are being written

PR #994, opened 2024-08-21
Change Kubernetes script cancellation to better handle pod deletion

APErebus:
# Background

Kubernetes script pod cancellation takes into account the grace seconds of the script pod. Which means that after cancellation, the pod can run for up to 30s before termination. This PR changes this logic to delete the pod with `0` grace seconds, it also returns a completed state from the `CancelScript` call so that the script observation stops calling `CancelScript` multiple times.

# Results

Script execution is now forced to not have a grace period. We also immediately mark the task as completed, which then stops the script observation loop from calling `CancelScript` again, which can result in an erroneous error as the pod has been removed.

This is also consistent with the normal shell scripts, which return complete & the canceled error code as soon as the cancellation token is marked as canceled.

image

Old code, note the multiple CancelScript calls
image

New code, note single CancelScript call
image

Shortcut story: [sc-70002]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle/Kubernetes/KubernetesPodMonitor.cs
+++ b/source/Octopus.Tentacle/Kubernetes/KubernetesPodMonitor.cs
@@ -327,6 +327,14 @@ namespace Octopus.Tentacle.Kubernetes
             };
         }
 
+        public static TrackedScriptPodState Canceling()
+        {
+            return new TrackedScriptPodState()
+            {
+                Phase = TrackedScriptPodPhase.Canceling
+            };
+        }
+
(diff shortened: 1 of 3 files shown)


zentron (approved):
LGTM,
 my only suggestion would be if we could link to either a doc\PR\issue\code\slack something that demonstrates this immediate exit is a deliberate behavior and what "normal" deployment tasks do.

APErebus:
@zentron ADR added here: https://github.com/OctopusDeploy/adr/blob/main/team-modern-deployments/kubernetes-agent/adr-11-immediate-pod-cancellation

PR #995, opened 2024-08-22
Makes --proxy added in #46 available to Docker containers

### LINKED ISSUE #996: Proxy configuration support in Docker images
#46 added support for specifying the proxy Octopus should use when communicating with a listening Tentacle

Docker images should also provide this configuration option.

mwalling-dc:
# Background

https://github.com/OctopusDeploy/OctopusTentacle/pull/46 added the `--proxy` CLI arg. This change adds the ability to pass the value for that token as an environment variable into the container.

# Results

- Shortcut: [SC-90394]
- Resolves #996

## After

Specifying `ProxyName` as an environment variable to the container will add it as the value of the  `--proxy` CLI flag.

## Testing

This has been running in production for approximately a year. We're trying to unwind our customizations on top of the upstream image.

### DIFF
--- a/docker/linux/scripts/configure-tentacle.sh
+++ b/docker/linux/scripts/configure-tentacle.sh
@@ -214,9 +214,9 @@ function registerTentacle() {
         if [[ ! -z "$ListeningPort" && "$ListeningPort" != "$internalListeningPort" ]]; then
             ARGS+=('--tentacle-comms-port' $ListeningPort)
         fi
-		if [[ ! -z "$ProxyName" ]]; then
+        if [[ ! -z "$ProxyName" ]]; then
             ARGS+=('--proxy' "$ProxyName")
-		fi
+        fi
     fi
 
     if [[ ! -z "$ServerApiKey" ]]; then
--- a/docker/linux/scripts/configure-tentacle.sh
+++ b/docker/linux/scripts/configure-tentacle.sh
@@ -90,9 +90,9 @@ function validateVariables() {
             echo " - communication mode 'Listening' (Passive)"
         fi
         echo " - registered port $ListeningPort"
-    if [[ ! -z "$ProxyName" ]]; then
-      echo " - proxy '$ProxyName'"
-    fi
+        if [[ ! -z "$ProxyName" ]]; then
+            echo " - proxy '$ProxyName'"
+        fi
     fi
     if [[ ! -z "$TargetWorkerPool" ]]; then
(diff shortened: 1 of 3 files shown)


docker/linux/scripts/configure-tentacle.sh
```diff
@@ -90,6 +90,9 @@ function validateVariables() {
             echo " - communication mode 'Listening' (Passive)"
         fi
         echo " - registered port $ListeningPort"
+    if [[ ! -z "$ProxyName" ]]; then
+      echo " - proxy '$ProxyName'"
+    fi
```
rhysparry, line 95: nit: use indenting consistent with the above

docker/linux/scripts/configure-tentacle.sh
```diff
@@ -211,6 +214,9 @@ function registerTentacle() {
         if [[ ! -z "$ListeningPort" && "$ListeningPort" != "$internalListeningPort" ]]; then
             ARGS+=('--tentacle-comms-port' $ListeningPort)
         fi
+		if [[ ! -z "$ProxyName" ]]; then
+		  ARGS+=('--proxy' $ProxyName)
```
rhysparry, line 218: Quote the `$ProxyName` parameter to ensure spaces in the name are captured

```suggestion
		  ARGS+=('--proxy' "$ProxyName")
```

docker/linux/scripts/configure-tentacle.sh
```diff
@@ -211,6 +214,9 @@ function registerTentacle() {
...
+		fi
```
rhysparry, line 219: nit: as above, use consistent indenting

docker/windows/Scripts/configure-tentacle.ps1
```diff
@@ -132,6 +133,9 @@ function Validate-Variables() {
       Write-Log " - communication mode 'Listening' (Passive)"
     }
     Write-Log " - registered port $ListeningPort"
+    if($null -ne $ProxyName) {
```
rhysparry, line 136: nit:
```suggestion
    if ($null -ne $ProxyName) {
```

rhysparry (changes requested):
Thanks for this PR. With some minor changes we should be able to approve.

Could you please add details of any testing you might have done to your PR description?

Many thanks

rhysparry (approved):
Looks good. Thanks for the contribution.

rhysparry:
Thanks again. This change is available from version 8.1.2115 onwards

PR #997, opened 2024-08-28
Debian 12 Docker Image for Kubernetes Tentacle

kevjt:
# Background

The Kubernetes Tentacle Docker image uses `mcr.microsoft.com/dotnet/runtime-deps:6.0` which is based on Debian 11 ("bullseye-slim").

Debian 11 is getting old and there are numerous CVEs. As such, the goal is to move to a Debian 12 ("bookworm-slim") based Docker image which has fewer security vulnerabilities  (see below for scan results comparison).

Debian 12 ships with OpenSSLv3 which means that SHA1RSA certificates can cause Halibut communication errors when Octopus Server is running on Windows Server 2012 R2. See this [doc](https://docs.google.com/document/d/1jgApNQDZYYbNWLopsyK-GPAclI1Wjv-6ped3euY5H-s/edit?usp=sharing) for more information about how and why this occurs. The issue arises primarily because the cipher suites on the older Windows OS are incompatible with OpenSSLv3 when using the SHA1RSA certificate signature algorithm. 

As such, the intention here is to publish **both** Debian 11 and Debian 12 based Kubernetes Tentacle images, to support any customers who are still running Octopus Server on Windows 2012 R2.

## Image Scans

Scanned using [trivy](https://github.com/aquasecurity/trivy)

[deb11-scan.txt](https://github.com/user-attachments/files/16776905/deb11-scan.txt)
[deb12-scan.txt](https://github.com/user-attachments/files/16776910/deb12-scan.txt)

**Summary comparison**

octopusdeploy/kubernetes-agent-tentacle:my-local-image-debian11 (debian 11.10)
Total: 158 (UNKNOWN: 0, LOW: 89, MEDIUM: 45, HIGH: 20, CRITICAL: 4)

octopusdeploy/kubernetes-agent-tentacle:my-local-image-debian12 (debian 12.6)
Total: 111 (UNKNOWN: 0, LOW: 73, MEDIUM: 30, HIGH: 7, CRITICAL: 1)

The 1 critical vulnerability in Debian 12 is on [CVE-2023-45853](https://security-tracker.debian.org/tracker/CVE-2023-45853) which it shares with Debian 11.

# Results

Configures the build process of the Kubernetes Tentacle Docker image such that both Debian 11 and 12 versions of the image are built and published. 

Debian 12 becomes the new default base for the image, as the Debian 11 tag is suffixed with `-bullseye-slim`


docker/kubernetes-agent-tentacle/Dockerfile
```diff
@@ -1,3 +1,7 @@
+# The default image is bullseye-slim (Debian 11)
+ARG RuntimeDepsTag="6.0-bullseye-slim"
```
kevjt, line 2: This needs to be defined at the very top as a 'global' ARG because this Dockerfile defines a multi-stage build. Each `FROM` statement declares a new scope which lasts until either the next FROM or the end of the file.

kevjt, line 2: This needs to be defined at the very top as a 'global' ARG because this Dockerfile uses a multi-stage build. Each `FROM` statement declares a new scope which lasts until either the next FROM or the end of the file.

docker/kubernetes-agent-tentacle/Dockerfile
```diff
@@ -1,3 +1,6 @@
+ARG RuntimeDepsTag
```
kevjt, line 1: This needs to be defined at the very top as a 'global' ARG because the Dockerfile uses a multi-stage build. Each `FROM` statement declares a variable scope which lasts until either the next FROM or the end of file. This means that putting this line right above the FROM line where it is used will cause it to fall under the previous FROM's scope and hence not work.

kevjt:
I have tested the Deb12 version of Tentacle against a Kubernetes Agent running locally. 
I'll also be running the full set of Kubernetes Agent/Worker E2E tests against the branch build of Tentacle to confirm everything is working.

kevjt:
I have tested the Debian12 version against a Kubernetes Agent running locally. 
I'll also run the full set of Kubernetes Agent/Worker E2E tests against the branch build of Tentacle to confirm everything is working.

kevjt:
> Details

E2Es ran as part of this [build](https://github.com/OctopusDeploy/OctopusDeploy/pull/27199)

build/Build.Pack.cs
```diff
@@ -21,6 +21,8 @@
 
 partial class Build
 {
+    const string KubernetesTentacleContainerRuntimeDepsTag = "6.0-bookworm-slim";
```
APErebus, line 24: now we are building .NET 8, should this be `8.0-bookworm-slim`?

kevjt: Yes, good spot!

samdanaei (approved):
Seems all good from a code owner perspective 👍

APErebus (approved):

PR #1007, opened 2024-09-26
Add script to guide developers through setting up their environment to allow locally debugging the Kubernetes Agent

scme0:
# Background

Previously we haven't had a nice way of debugging the Kubernetes Agent, and we never had a way to debug it locally. Looking forward we might have more people using the Agent so it's a good idea to formalise the process of how to setup a developers environment to allow for local debugging.

# Results

I've created a script, along with a few small changes to the Tentacle code base to help guide engineers through the process of setting up their environment to locally debug the agent.

Caveats: It has only been tested on MacOS - updates are probably required for Windows and Linux.

# Video Walk Through

https://octopus.zoom.us/rec/share/Az-ugVGNR8Mg7_WWZa8FCwiqg4nJwgDkVx9N41EbzaHNPFn9EcygVZwG51sY_skE.Q913692PVXEdOXG3?startTime=1727326867000
Passcode: &429gPqU

### DIFF
--- a/k8s-agent-debug-cluster-config/launchSettings.template.json
+++ b/k8s-agent-debug-cluster-config/launchSettings.template.json
@@ -1,0 +1,29 @@
+{
+  "profiles": {
+    "Octopus Kubernetes Agent": {
+      "commandName": "Project",
+      "commandLineArgs": "agent --instance Tentacle --noninteractive",
+      "framework": "net8.0",
+      "environmentVariables": {
+        "ACCEPT_EULA": "Y",
+        "OCTOPUS__K8STENTACLE__NAMESPACE": "octopus-agent-<agent-name>",
+        "OCTOPUS__K8STENTACLE__PODSERVICEACCOUNTNAME": "octopus-agent-scripts",
+        "OCTOPUS__K8STENTACLE__PODVOLUMECLAIMNAME": "octopus-agent-<agent-name>-pvc",
+        "OCTOPUS__K8STENTACLE__HELMRELEASENAME": "<agent-name>",
+        "OCTOPUS__K8STENTACLE__HELMCHARTVERSION": "2.4.0",
+        "OCTOPUS__K8STENTACLE__DISABLEAUTOPODCLEANUP": "false",
+        "OCTOPUS__K8STENTACLE__DISABLEPODEVENTSINTASKLOG": "false",
+        "OCTOPUS__TENTACLE__LOGLEVEL": "Info",
+        "TentacleHome": "/tmp/k8s-agent-debug-vol",
+        "TentacleApplications": "/tmp/k8s-agent-debug-vol/Applications",
+        "TentaclePollingConnectionCount": "5",
+        "OCTOPUS__K8STENTACLE__ENABLEMETRICSCAPTURE": "false",
+        "OCTOPUS__K8STENTACLE__PERSISTENTVOLUMESIZE": "10Gi",
+        "KUBECONFIG": "<kube-config>",
+        "DefaultLogDirectory": "/tmp/k8s-agent-debug-vol/logs",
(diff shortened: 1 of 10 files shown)


source/Octopus.Tentacle/Diagnostics/OctopusLogsDirectoryRenderer.cs
```diff
@@ -11,7 +11,7 @@ namespace Octopus.Tentacle.Diagnostics
     [LayoutRenderer("octopusLogsDirectory")]
     public class OctopusLogsDirectoryRenderer : LayoutRenderer
     {
-        public static readonly string DefaultLogsDirectory =
+        public static readonly string DefaultLogsDirectory = Environment.GetEnvironmentVariable("DefaultLogDirectory") ??
```
kevjt, line 14: Should this be placed in the EnvironmentVariables class as a const?

source/Octopus.Tentacle/Kubernetes/KubernetesPhysicalFileSystem.cs
```diff
@@ -13,7 +13,7 @@ public class KubernetesPhysicalFileSystem : OctopusPhysicalFileSystem
         // Set like this for now because we don't have a way to get the home directory from the provider without requiring ourselves
         // DI can be painful when circular dependencies happen with constructed classes :sad-panda:
         // When we can get an Injectable KubernetesConfiguration, we can remove this, alternatively, we can pull apart the configuration stores into different implementations
-        const string HomeDir = "/octopus";
+        readonly string HomeDir = Environment.GetEnvironmentVariable("TentacleHome") ?? "/octopus";
```
kevjt, line 16: This can use `EnvironmentVariables.TentacleHome`

kevjt (approved):
Great work man! I tried it and debugging works smoothly. The only trouble I had was the fact that Rider didn't immediately use the target framework specified in the launchsettings.
I just have a few very minor comments.

scme0:
> The only trouble I had was the fact that Rider didn't immediately use the target framework specified in the launchsettings.

Ah yeah actually I don't think that framework variable in the launchsettings is valid. I was trying it to see if it would work but I don't think you can set it from the launch settings file unfortunately. I'll remove the line from the launchsettings. I do have a log line on the last line of the script which says to check the run configuration because I was also having trouble with it 🤷‍♂️

k8s-agent-debug-cluster-config/launchSettings.template.json
```diff
@@ -0,0 +1,29 @@
+{
+  "profiles": {
+    "Octopus Kubernetes Agent": {
+      "commandName": "Project",
+      "commandLineArgs": "agent --instance Tentacle --noninteractive",
+      "framework": "net8.0",
+      "environmentVariables": {
+        "ACCEPT_EULA": "Y",
+        "OCTOPUS__K8STENTACLE__NAMESPACE": "octopus-agent-<agent-name>",
```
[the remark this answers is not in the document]
scme0: Ah yes, didn't think about this. I was considering updating the script to query the cluster and find the agent by finding a helm release that has a matching name or something. But I thought initially this would work. I'm thinking maybe lets get this first version in and we can iterate on it 👍

source/Octopus.Tentacle/Diagnostics/OctopusLogsDirectoryRenderer.cs
scme0: Yep good idea 👍

scme0: Done!

source/Octopus.Tentacle/Kubernetes/KubernetesPhysicalFileSystem.cs
scme0: Done!

PR #1010, opened 2024-10-02
Implement image metadata check for Kubernetes Agent tools to use the latest image tag revision for pull policy workaround

kevjt:


source/Octopus.Tentacle/Kubernetes/KubernetesAgentToolsImageVersionMetadataProvider.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            this.log = log;
+        }
+        
+        public async Task<KubernetesAgentToolsImageVersionMetadata?> TryGetVersionMetadata()
+        {
+            using var httpClient = new HttpClient();
+            try
+            {
+                var response = await httpClient.GetAsync("https://oc.to/kubernetes-agent-tools-image-metadata");
```
kevjt, line 29: Using a short URL allows us to avoid updating Tentacle if the location of the metadata file changes.

source/Octopus.Tentacle/Kubernetes/ClusterVersion.cs
```diff
@@ -0,0 +1,69 @@
...
+                return false;
+
+            return clusterVersion.Major == Major && clusterVersion.Minor == Minor;
+        }
+
+        public override int GetHashCode()
+        {
+#if NET8_0_OR_GREATER
+            return HashCode.Combine(Major, Minor);
+#else
```
kevjt, line 53: I'm not sure if there's a way to avoid doing this. I ran into a compiler error similar to the one described [here](https://stackoverflow.com/questions/65941418/cannot-use-new-hashcode-to-override-gethashcode-in-class) because HashCode is not available in the .NET Framework. Since we only build the Kubernetes Tentacle with .NET 8, this part of the code would never actually be used.

source/Octopus.Tentacle/Kubernetes/KubernetesModule.cs
```diff
@@ -15,6 +15,13 @@ protected override void Load(ContainerBuilder builder)
         {
             builder.RegisterType<KubernetesPodService>().As<IKubernetesPodService>().SingleInstance();
             builder.RegisterType<KubernetesClusterService>().As<IKubernetesClusterService>().SingleInstance();
+
+            const string agentToolsImageMetadataProvider = "AgentToolsImageMetadataProvider";
+            builder.RegisterType<KubernetesAgentToolsImageVersionMetadataProvider>().Named<IToolsImageVersionMetadataProvider>(agentToolsImageMetadataProvider);
+            builder.RegisterDecorator<IToolsImageVersionMetadataProvider>(
+                (context, inner) => new CachingKubernetesAgentToolsImageVersionMetadataProvider(
+                    inner, context.Resolve<IMemoryCache>()), fromKey: agentToolsImageMetadataProvider);
```
APErebus, line 23:
```suggestion
            builder.RegisterDecorator<CachingKubernetesAgentToolsImageVersionMetadataProvider, IToolsImageVersionMetadataProvider>();
```
You _should_ be able to just do this and not need the named service?

source/Octopus.Tentacle/Kubernetes/KubernetesAgentToolsImageVersionMetadataProvider.cs
```diff
@@ -0,0 +1,63 @@
...
+                var json = await response.Content.ReadAsStringAsync();
+                return JsonConvert.DeserializeObject<KubernetesAgentToolsImageVersionMetadata>(json);
+            }
+            catch (Exception ex)
+            {
+                log.Error($"Failed to fetch version metadata for the agent tools container image. Details: {ex.Message}");
```
APErebus, line 58: can `Error()` take the `ex` as the first param?

source/Octopus.Tentacle/Kubernetes/KubernetesModule.cs
kevjt: I tried that at first but was getting a compiler error. It seems that this syntax was introduced in Autofac [version 4.9](https://autofac.readthedocs.io/en/latest/advanced/adapters-decorators.html#simplified-syntax). Tentacle is on version 4.6.2

APErebus: _sigh_ of course 😢

APErebus (approved):
LGTM!

PR #1016, opened 2024-10-04
Fix flaky test: Remove stray whitespace in expected exception messages.

gb-8:
# Background

My build failed due to a flaky test:
https://build.octopushq.com/test/-3830879800243621547?currentProjectId=TeamFireAndMotion_OctopusTentacle_TentacleVLatest_net80&expandTestHistoryChartSection=true

It looks like there is stray whitespace in one of the expected exception messages that is causing the test to occasionally fail.

# Results

This PR removes the stray whitespace.

# How to review this PR

Was that whitespace deliberate?
Do we need to cater for messages both with and without the extra space?


rhysparry (approved):
✅

PR #1024, opened 2024-10-16
Separate chocolateyInstall.ps1 files for self contained and net-framework-dependent tentacles

samdanaei:
# Background

A Chocolatey moderator has [pointed out](https://community.chocolatey.org/packages/OctopusDeploy.Tentacle.SelfContained/8.2.2297#testingResults) that the same `chocolateyinstall.ps1` file is being used for both flavours of tentacle. 

chocolateyinstall.ps1 is a script that actually runs on the user's machine for a particular package install. Therefore sharing the same chocolateyinstall.ps1 is not suitable for both packages. It needs to be unique for each package.

We prevented chocolateyinstall.ps1 calling `Install-ChocolateyPackage` twice **only on the OctopusDeploy.Tentacle package** with #1025. 

# Results

This PR splits the installation scripts into separate folders during packing, resulting in two separate chocolateyinstall.ps1 files.

## Before

The OctopusDeploy.Tentacle.SelfContained nupkg's chocolateyinstall.ps1 calls `Install-ChocolateyPackage` twice. It should only have called the latter

image

## After

The OctopusDeploy.Tentacle.SelfContained nupkg's chocolateyinstall.ps1 only calls `Install-ChocolateyPackage` once

image

The OctopusDeploy.Tentacle nupkg's chocolateyinstall.ps1 also only calls `Install-ChocolateyPackage` once. This is existing behaviour as of #1025, and has not changed with this PR.

image

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


rhysparry (approved):
LGTM ✅

PR #1025, opened 2024-10-16
Remove self-contained tentacle installation from existing chocolateyInstall.ps1 file

samdanaei:
# Background

A Chocolatey moderator has [pointed out](https://community.chocolatey.org/packages/OctopusDeploy.Tentacle.SelfContained/8.2.2297#testingResults) that the same chocolateyinstall.ps1 file is being used for both flavours of tentacle.

chocolateyinstall.ps1 is a script that actually runs on the user's machine. It's not only a helper file for Chocolatey. Therefore it needs to be unique for each package.

# Results

This PR removes the installation of self-contained tentacles for now, to mitigate any effect on the existing chocolatey package, which should only install the .NET Framework-dependant tentacle. 

A follow up PR is being worked on [here](https://github.com/OctopusDeploy/OctopusTentacle/pull/1024), which will split each into its own chocolateyinstall.ps1 file. 

Packaging code in build.pack.cs has not been rolled back, as it will still be used to control the new chocolateyinstall.ps1 file in the upcoming PR.

## Before

We assume that the current state would result in a user getting both flavours installed on Windows. Since the self-contained version is the second one, it would install the self-contained version over the .NET framework dependant version, when a user tries to install the OctopusDeploy.Tentacle package.

## After

A user will get only the  .NET framework dependant version installed from the OctopusDeploy.Tentacle package.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


rhysparry (approved):
LGTM ✅

PR #1044, opened 2024-11-25
Mount agent-upgrade credentials default location in filesystem

rain-on:
# Background

To avoid rate-limiting when polling docker for agent versions/charts - helm must be provided with credentials via the `~/.config/helm/registry/config.json` file (though will fallback to docker config/env).

This change is responsible for mapping the secret in the helm chart, into the filesystem of the script-pod.

It is optional, such that if the secret does not exist, and empty directory is mounted instead.

The change was required to be made to both raw, and normal script pods (though there should be a nicer way of doing this to avoid string-matching across files :( ).

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Octopus.Tentacle/Kubernetes/KubernetesRawScriptPodCreator.cs
```diff
@@ -80,7 +80,24 @@ protected override IList<V1Volume> CreateVolumes(StartKubernetesScriptCommandV1
...
+                        {
+                            new()
+                            {
+                                Key = ".dockerconfigjson",
+                                Path = "config.json"
+                            }
+                        },
+                        Optional = true,
+                    },
+                },
```
APErebus, line 100: Can we define this as a method in the base `KubernetesScriptPodCreator`, so we don't have it defined twice?

APErebus (approved):
LGTM, just 1 suggestion to reduce duplication, but not blocking

PR #1046, opened 2024-11-28
Switch from CliWrap to Shellfish

borland:
**_Are you a customer of Octopus Deploy? Please contact [our support team](https://octopus.com/support) so we can triage your PR, so that we can make sure it's handled appropriately._**

# Background

# Results

## Before

Before

## After

After

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Octopus.Tentacle.Tests.Integration/TentacleCommandLineTests.cs
```diff
@@ -119,7 +118,7 @@ public async Task VersionCommandTextFormat(TentacleConfigurationTestCase tc)
 
             var expectedVersion = GetVersionInfo(tc);
 
-            stdout.Should().Be(expectedVersion.ProductVersion, "The version command should print the informational version as text");
+            stdout.TrimEnd().Should().Be(expectedVersion.ProductVersion, "The version command should print the informational version as text");
```
borland, line 121: CliWrap deals with text output in a slightly different way to Shellfish.

If the app emits text without a trailing newline, CliWrap wouldn't give you a trailing newline either, whereas with Shellfish we get a trailing newline.

This isn't an inherent flaw in Shellfish, but rather just a side effect `WithStdOutTarget(stringBuilder)` output mechanism. The stringbuilder-writer puts a newline on the end of everything to keep things simple. We could do `WithStdOutTarget(someCustomTarget)` and get the output in a more direct way, but that would be harder and more complex than simply trimming the newlines in a few places

LukeButters: For the test I don't think it matters.

LukeButters (approved):

LukeButters:
No doube the build has to actually work reliably before merging ;p

borland:
> No doube the build has to actually work reliably before merging ;p

Yeah I don't know what happened there, the tests just stalled forever on Win10 LTSC with no kind of log or anything, but they worked fine all the other OSes. Not a single test ran... I don't think it had anything to do with the Shellfish change, I'm re-running

PR #1053, opened 2025-01-05
Update dotnet-install.ps1 script for new CDN urls

rhysparry:
# Background

The [.NET install links recently changed](https://devblogs.microsoft.com/dotnet/critical-dotnet-install-links-are-changing/). We maintain a modified copy of the `dotnet-install.ps1` script in this repository to ensure TLS 1.2 is used when the script is run on older operating systems.

This change applies these changes to the current latest script.

# Results

Script updated to reflect the upstream changes. Note: the upstream version includes the new domains, but still has some references to the old domains (`dotnetcli.azureedge.net` and `dotnetbuilds.azureedge.net`). There is an upstream issue tracking this: https://github.com/dotnet/install-scripts/issues/559

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/dotnet-install.ps1
+++ b/dotnet-install.ps1
@@ -3,7 +3,7 @@
 # Licensed under the MIT license. See LICENSE file in the project root for full license information.
 #
 
-# Duplicated from https://dot.net/v1/dotnet-install.ps1 @ 2024-07-01
+# Duplicated from https://dot.net/v1/dotnet-install.ps1 @ 2025-01-06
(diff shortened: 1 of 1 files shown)


gb-8 (approved):
Maybe considered out-of-scope, but the way we are enabling TLS 1.2 forces the disabling of TLS 1.3.

I think it would be better to bitwise AND with TLS 1.2 instead of straight assignment.

gb-8 (approved):
Approved, but with same question about not disabling TLS 1.3 as on the Halibut PR.

PR #1055, opened 2025-01-09
Increase the maximum buffer size of the Kubernetes agent bootstrapRunner's stream readers

kevjt:
# Background

Part of https://github.com/OctopusDeploy/Issues/issues/9202

This PR addresses an issue where long text cannot be written to Octopus output variables when deploying via the Octopus Kubernetes Agent. The problem is caused by the default max buffer size of the `bufio.Scanner `in the bootstrap runner, which is limited to 64 KiB. This causes large outputs (e.g., from Terraform plans) to not be written to service message logs.

[sc-100196]

# Results

- The max buffer size for the Scanner has been increased to allow larger outputs, with the maximum size set to 10 MiB, which aligns with Kubernetes' default max container log size.

- Ensures the reader is closed after scanning (regardless of error) to properly close the pipe, allowing the command process to terminate without hanging while waiting for the pipe to clear before writing stdout.

### DIFF
--- a/docker/kubernetes-agent-tentacle/bootstrapRunner/bootstrapRunner.go
+++ b/docker/kubernetes-agent-tentacle/bootstrapRunner/bootstrapRunner.go
@@ -20,6 +20,12 @@ type SafeCounter struct {
 	Value int
 }
 
+const (
+	// MaxScanTokenSize 10Mi Max token size
+	// 10Mi is the maximum default size of a pod's log file
+	MaxScanTokenSize = 10 * 1024 * 1024
+)
+
 // The bootstrapRunner applet is designed to execute a script in a specific folder
 // and format the script's output from stdout and stderr into the following format:
 // <line number>|<RFC3339Nano date time>|<"stdout" or "stderr">|<original string>
@@ -47,6 +53,11 @@ func main() {
 
 	stdOutScanner := bufio.NewScanner(stdOutCmdReader)
 	stdErrScanner := bufio.NewScanner(stdErrCmdReader)
+
+	// Create an initial buffer to be used by the scanners
+	buffer := make([]byte, 64*1024) // 64K initial buffer
+	stdOutScanner.Buffer(buffer, MaxScanTokenSize)
+	stdErrScanner.Buffer(buffer, MaxScanTokenSize)
 
 	doneStd := make(chan bool)
 	doneErr := make(chan bool)
--- a/docker/kubernetes-agent-tentacle/bootstrapRunner/bootstrapRunner.go
+++ b/docker/kubernetes-agent-tentacle/bootstrapRunner/bootstrapRunner.go
@@ -94,7 +94,7 @@ func reader(readCloser io.ReadCloser, stream string, done *chan bool, counter *S
 	scanner := bufio.NewScanner(readCloser)
 
 	// Create an initial buffer to be used by the scanners
-	buffer := make([]byte, 4096) // 4Ki initial buffer
+	buffer := make([]byte, 4096) // 4Ki initial buffer https://github.com/golang/go/blob/master/src/bufio/scan.go#L84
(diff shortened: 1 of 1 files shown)


docker/kubernetes-agent-tentacle/bootstrapRunner/bootstrapRunner.go
```diff
@@ -48,6 +54,11 @@ func main() {
 	stdOutScanner := bufio.NewScanner(stdOutCmdReader)
 	stdErrScanner := bufio.NewScanner(stdErrCmdReader)
 
+	// Create an initial buffer to be used by the scanners
+	buffer := make([]byte, 64*1024) // 64K initial buffer
+	stdOutScanner.Buffer(buffer, MaxScanTokenSize)
+	stdErrScanner.Buffer(buffer, MaxScanTokenSize)
```
flin-8, line 60: CleanShot 2025-01-10 at 11 03 02@2x

So to be clear we are setting the initial buffer to 64K when it was 4K previously? Is this intentional?

docker/kubernetes-agent-tentacle/bootstrapRunner/bootstrapRunner.go
```diff
@@ -48,6 +54,11 @@ func main() {
 	stdOutScanner := bufio.NewScanner(stdOutCmdReader)
 	stdErrScanner := bufio.NewScanner(stdErrCmdReader)
 
+	// Create an initial buffer to be used by the scanners
+	buffer := make([]byte, 64*1024) // 64K initial buffer
```
flin-8, line 58: Is the 64K meant to match `bufio.MaxScanTokenSize`?

docker/kubernetes-agent-tentacle/bootstrapRunner/bootstrapRunner.go
kevjt: Good point. I'll update it to match the default start buffer size.

docker/kubernetes-agent-tentacle/bootstrapRunner/bootstrapRunner.go
```diff
@@ -87,10 +90,25 @@ func main() {
...
+	// Create an initial buffer to be used by the scanners
+	buffer := make([]byte, 4096) // 4Ki initial buffer
+	scanner.Buffer(buffer, MaxScanTokenSize)
+
 	for scanner.Scan() {
 		Write(stream, scanner.Text(), counter, gcm)
 	}
+
+	err := readCloser.Close()
+	if scanner.Err() != nil {
```
flin-8, line 105: QQ: is it worth printing the scanner error before closing the reader?

docker/kubernetes-agent-tentacle/bootstrapRunner/bootstrapRunner.go
```diff
@@ -87,10 +90,25 @@ func main() {
 	os.Exit(exitCode)
 }
 
-func reader(scanner *bufio.Scanner, stream string, done *chan bool, counter *SafeCounter, gcm cipher.AEAD) {
+func reader(readCloser io.ReadCloser, stream string, done *chan bool, counter *SafeCounter, gcm cipher.AEAD) {
+	scanner := bufio.NewScanner(readCloser)
+
+	// Create an initial buffer to be used by the scanners
+	buffer := make([]byte, 4096) // 4Ki initial buffer
```
flin-8, line 97: Might be good to link to the Bufio source code that has 4096

docker/kubernetes-agent-tentacle/bootstrapRunner/bootstrapRunner.go
kevjt: The scanning error will be available even after closing the reader, because the error is associated with the scanner object itself, not the reader.

PR #1058, opened 2025-01-21
Event driven script orchestrator

sburmanoctopus:
[sc-101635]

# Background

Currently we only allow calling into tentacle client to run a script via the TentacleClient.ExecuteScript method. Although extremely simple it is not re-entrant. When we talk about re-entrancy we are considering the case the client gets turned off and on and wants to continue on. The deficiencies are:
- when using SSv1 could start a script a second time
- it will re-fetch all logs, rather than where or close to the point we left off from.
- it doesn't work well from an event driven orchestrator. 

# Results

This PR was a continuation of the [spike started here](https://github.com/OctopusDeploy/OctopusTentacle/pull/966)

In summary the changes are:

- The ObservingOrchestrator is no longer abstract and becomes a thing that "just" knows how to run the life cycle of the script.
- A new IScriptExecutor interface is created which allows driving script life cycle
  - This is now used by the ObservingOrchestrator
  - Could be used by an event driven orchestrator.
- The ScriptService Orchestrators which inherited from  ObservingOrchestrator, now implement IScriptExecutor. They now  focus on implementing the main stages of script execution following a single interface.
  - They are also responsible from converting from types the client of the client might use to the types that go over the wire.
  - They are responsible for their own "quirks" providing a straight forward interface to interact with.
- A ScriptExecutor is created which:
   - On start script determines which of the other ScriptServices to use.
   -  On the other methods uses the given CommandContext to determine which script service to use.
   - In doing so it can deal with calling the correct service.
- The ObservingOrchestrator is given a `ScriptExecutor`
   - Which is called by TentacleClient.ExecuteScript
   - In this way the existing tests are able to exercise the  ScriptExecutor which is what an event driven orchestrator would call.
   - This reduces what an event driven orchestrator needs to implement to just what is done in ObservingOrchestrator


source/Octopus.Tentacle.Client/EventDriven/CommandContext.cs
```diff
@@ -0,0 +1,21 @@
+using Octopus.Tentacle.Client.Scripts;
+using Octopus.Tentacle.Contracts;
+
+namespace Octopus.Tentacle.Client.EventDriven
+{
+    public class CommandContext
```
LukeButters, line 6: nit: consider documenting what this class is for.

source/Octopus.Tentacle.Client/Scripts/IScriptExecutor.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Threading.Tasks;
+using Octopus.Tentacle.Client.EventDriven;
+using Octopus.Tentacle.Client.Scripts.Models;
+using Octopus.Tentacle.Contracts;
+
+namespace Octopus.Tentacle.Client.Scripts
+{
+    public interface IScriptExecutor 
+    {
+        Task<ScriptOperationExecutionResult> StartScript(ExecuteScriptCommand command,
```
LukeButters, line 12: It might make sense to document that the resulting CommandContext should be used for the next call.

source/Octopus.Tentacle.Client/Scripts/IScriptExecutor.cs
```diff
@@ -0,0 +1,22 @@
...
+            StartScriptIsBeingReAttempted startScriptIsBeingReAttempted,
+            CancellationToken scriptExecutionCancellationToken);
+        
+        Task<ScriptOperationExecutionResult> GetStatus(CommandContext commandContext, CancellationToken scriptExecutionCancellationToken);
```
LukeButters, line 16: nit: I think I actually preferred the tuple because the responses where not something that are really coupled AND I think it improved readability. ie it was clear each method was returning a `CommandContext` which might be enough to hint to the user to use that `CommandContext` for the next call.

source/Octopus.Tentacle.Client/Scripts/ScriptServiceVersion.cs
```diff
@@ -1,6 +1,7 @@
 ﻿namespace Octopus.Tentacle.Client.Scripts
 {
-    record ScriptServiceVersion(string Value)
+    // TODO should not be public.
```
LukeButters, line 3: I assume we will address this in a future

source/Octopus.Tentacle.Client/Scripts/UnsafeStartAttemptException.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace Octopus.Tentacle.Client.Scripts
+{
+    public class UnsafeStartAttemptException : Exception
```
LukeButters, line 5: If this is used we should document what it is about.

source/Tentacle.sln.DotSettings
```diff
@@ -183,6 +183,7 @@
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpPlaceEmbeddedOnSameLineMigration/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpRenamePlacementToArrangementMigration/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpUseContinuousIndentInsideBracesMigration/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002EMemberReordering_002EMigrations_002ECSharpFileLayoutPatternRemoveIsAttributeUpgrade/@EntryIndexedValue">True</s:Boolean>
```
LukeButters, line 186: just dotnet things

LukeButters (approved):

source/Octopus.Tentacle.Client/EventDriven/CommandContext.cs
sburmanoctopus: Good idea. Done

source/Octopus.Tentacle.Client/Scripts/IScriptExecutor.cs
sburmanoctopus: Done

source/Octopus.Tentacle.Client/Scripts/IScriptExecutor.cs
sburmanoctopus: > nit: I think I actually preferred the tuple because the responses where not something that are really coupled
🤔 I feel that the fact we return both implies that they are related.

> I think it improved readability. ie it was clear each method was returning a CommandContext which might be enough to hint to the user to use that CommandContext for the next call.

🤔 I would have thought the property named ContextForNextCommand would be good enough. Plus, adding the documentation from the [previous comment](https://github.com/OctopusDeploy/OctopusTentacle/pull/1058#discussion_r1922992631) should be more than enough.

I personally don't like tuples at all, and feel this is much easier to understand. For now, let's leave it the way it is, and if we find it unhelpful we can address it then.

source/Octopus.Tentacle.Client/Scripts/ScriptServiceVersion.cs
sburmanoctopus: This will be addressed as part of the work to use a serialized version of command context.

But yes, we shouldn't be leaving these TODOs.

source/Octopus.Tentacle.Client/Scripts/UnsafeStartAttemptException.cs
sburmanoctopus: Done

PR #1078, opened 2025-03-18
Link init script pod resource limits to main script pod limits

MissedTheMark:
# Background
The Kubernetes Agent currently supports setting custom resource limits on main script pods, but not init script pods - they have a fixed value defined in the agent itself.

This PR sets the init script pod resource limits to the values specified for the main script pods.

# Results

I added two agents to my local cluster, one with the old version, one with this version. 
They both had a memory limit set: `--set agent.scriptPods.resources.limits.memory="273Mi"`  

Only this version actually applied that limit.

Left is before, right is after:
image

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

- [X] I have read [How we use GitHub Issues](https://github.com/OctopusDeploy/Issues/blob/main/docs/CONTRIBUTING.internal.md) for help deciding when and where it's appropriate to make an issue.
- [X] I have considered informing or consulting the right people, according to the [ownership map](https://whimsical.com/ownership-map-NzbiD4HJyvhC9jNJNfS6TG).
- [X] I have considered appropriate testing for my change.

### DIFF
--- a/source/Octopus.Tentacle/Kubernetes/KubernetesRawScriptPodCreator.cs
+++ b/source/Octopus.Tentacle/Kubernetes/KubernetesRawScriptPodCreator.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Threading.Tasks;
+using k8s;
 using k8s.Models;
 using Octopus.Diagnostics;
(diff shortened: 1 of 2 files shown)


APErebus (approved):
Awesome. Thanks mate

PR #1085, opened 2025-04-03
Allow callers to cancel their request to cancel the script

LukeButters:
# Background

Otherwise callers of the event driven script executor can't bail out on requests to cancel the script.

Which is preventing octopus from shutting down.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Client/Scripts/ObservingScriptOrchestrator.cs
+++ b/source/Octopus.Tentacle.Client/Scripts/ObservingScriptOrchestrator.cs
@@ -79,7 +79,8 @@ namespace Octopus.Tentacle.Client.Scripts
             {
                 if (scriptExecutionCancellationToken.IsCancellationRequested)
                 {
-                    lastResult = await scriptExecutor.CancelScript(lastResult.ContextForNextCommand).ConfigureAwait(false);
+                    // We don't want to cancel this operation as it is responsible for stopping the script executing on the Tentacle
+                    lastResult = await scriptExecutor.CancelScript(lastResult.ContextForNextCommand, CancellationToken.None).ConfigureAwait(false);
                 }
                 else
                 {
--- a/source/Octopus.Tentacle.Client/ScriptExecutor.cs
+++ b/source/Octopus.Tentacle.Client/ScriptExecutor.cs
@@ -60,12 +60,12 @@ namespace Octopus.Tentacle.Client
             return await scriptExecutor.GetStatus(commandContext, cancellationToken);
         }
 
-        public async Task<ScriptOperationExecutionResult> CancelScript(CommandContext commandContext)
+        public async Task<ScriptOperationExecutionResult> CancelScript(CommandContext commandContext, CancellationToken cancellationToken)
         {
             var scriptExecutorFactory = CreateScriptExecutorFactory();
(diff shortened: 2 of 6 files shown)


source/Octopus.Tentacle.Client/Scripts/ObservingScriptOrchestrator.cs
```diff
@@ -79,7 +79,8 @@ async Task<ScriptOperationExecutionResult> ObserveUntilComplete(
             {
                 if (scriptExecutionCancellationToken.IsCancellationRequested)
                 {
-                    lastResult = await scriptExecutor.CancelScript(lastResult.ContextForNextCommand).ConfigureAwait(false);
+                    // We don't want to cancel this operation as it is responsible for stopping the script executing on the Tentacle
+                    lastResult = await scriptExecutor.CancelScript(lastResult.ContextForNextCommand, CancellationToken.None).ConfigureAwait(false);
```
LukeButters, line 83: The `CancellationToken.None` is moved up here.

source/Octopus.Tentacle.Client/ITentacleClient.cs
```diff
@@ -34,37 +34,43 @@ Task<ScriptExecutionResult> ExecuteScript(
             CancellationToken scriptExecutionCancellationToken);
 
         /// <summary>
-        /// Start the script.
+        ///
```
sburmanoctopus, line 37: Why did we delete this summary? Yeah, it's simple, but it's better than none.

source/Octopus.Tentacle.Client/Scripts/ObservingScriptOrchestrator.cs
sburmanoctopus: The reason for this PR is so that we can cancel the cancellation request, so that the cancellation request does not block Octopus Server from shutting down.

But this made me wonder, why is this not an issue for the non-resilient pipeline? Surely it would get told to cancel a running deployment when it shuts down too, right? I would have thought that would also block.

source/Octopus.Tentacle.Client/ITentacleClient.cs
```diff
@@ -34,37 +34,43 @@ Task<ScriptExecutionResult> ExecuteScript(
...
 
         /// <summary>
         /// Complete the script.
         /// </summary>
         /// <param name="commandContext">The CommandContext from the previous command</param>
         /// <param name="logger"></param>
-        /// <param name="scriptExecutionCancellationToken"></param>
-        Task<ScriptStatus?> CompleteScript(CommandContext commandContext, ITentacleClientTaskLog logger, CancellationToken scriptExecutionCancellationToken);
+        /// <param name="requestCancellationToken">Cancels the inflight request</param>
+        Task<ScriptStatus?> CompleteScript(CommandContext commandContext, ITentacleClientTaskLog logger, CancellationToken requestCancellationToken);
```
sburmanoctopus, line 74: Just wondering, what's the consequence of killing a request to cancel?

I would have thought this would result in the tentacle potentially running the script after we have told the user that we are cancelled, right?

Is this going to be an issue? I.e., is the trade off of shutting down Octopus Server quicker worth it?

sburmanoctopus (approved):
This PR does what it says on the tin. So I'll pre-approve it.

I would like to have those questions I raised answered before we merge though, as I want to be sure this is what we want to do.

PR #1105, opened 2025-05-26
Sast/try cancel script on failure

LukeButters:
# Background

We are seeing that short network issues are resulting in non recoverable timeouts when communicating with polling tentacles. This is then resulting in scripts which continue to execute on the tentacle, while Octopus claims the server task is failed. This executing script can prevent further tasks from executing on the tentacle. The executing script can not be killed other than either waiting for the script to complete OR restarting the tentacle (killing the in-flight scripts).

The issue is that the network dies in a way that is only detectable by timeout from the client. Ordinarily this is fine, however unfortunately we are frequently seeing the issue occur at the [5 minute timeout](https://github.com/OctopusDeploy/Halibut/blob/main/source/Halibut/Diagnostics/HalibutTimeoutsAndLimits.cs#L51) when the response is being waited for by the client. The timeout is high so as to allow the client time to execute the request.

Additionally the test utility of interrupting a response message from a polling tentacle is updated to no longer interrupt the set up of a connection between polling tentacle and client. Instead we now only kill/pause the connection once we know the connection from polling tentacle is setup.

# Results

This PR attempts to solve the problem of the script continuing to execute. This is done by sending a non awaited cancel request to the  tentacle, when calls to get status fail.

## Before

Short network issues might result in a failed ServerTask, while the script continues to execute on tentacle preventing further ServerTasks from progressing.

## After

An attempt to cancel the inflight script is made, which for short lived network failures should result in the script being cancelled on tentacle. This will allow further ServerTasks to execute on the tentacle.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionWillAttemptACancelCallWhenGetStatusFails.cs
+++ b/source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionWillAttemptACancelCallWhenGetStatusFails.cs
@@ -5,6 +5,7 @@ using FluentAssertions;
 using Halibut.Diagnostics;
 using NUnit.Framework;
 using Octopus.Tentacle.CommonTestUtils.Diagnostics;
+using Octopus.Tentacle.Contracts;
 using Octopus.Tentacle.Contracts.ClientServices;
 using Octopus.Tentacle.Tests.Integration.Common.Builders.Decorators;
 using Octopus.Tentacle.Tests.Integration.Support;
@@ -22,6 +23,7 @@ namespace Octopus.Tentacle.Tests.Integration
         [TentacleConfigurations]
         public async Task WhenANetworkFailureOccurs_DuringGetStatus_AndTheGetStatusCallDoesNotRecover_AnAttemptToCancelTheScriptIsMade(TentacleConfigurationTestCase tentacleConfigurationTestCase)
         {
+            bool isNetworkDown = true;
             await using var clientTentacle = await tentacleConfigurationTestCase.CreateBuilder()
                 .WithPortForwarderDataLogging()
                 .WithRetryDuration(TimeSpan.FromSeconds(10))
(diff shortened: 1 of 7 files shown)


source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionWillAttemptACancelCallWhenGetStatusFails.cs
```diff
@@ -0,0 +1,87 @@
...
+            // We should find that the cancel request was sent to tentacle.
+            await Wait.For(() => recordedUsages.For(nameof(IAsyncClientScriptServiceV2.CancelScriptAsync)).Completed > 0,
+                TimeSpan.FromSeconds(60),
+                () => throw new Exception("Cancel script was never called."),
+                CancellationToken);
+
+            // Let the script finish.
+            File.WriteAllText(waitForFile, "");
+            await Task.Delay(TimeSpan.FromSeconds(2)); // It's hard to know if the script was killed or not, if we do the assert below too quickly we will
+            // incorrectly see the script as being cancelled.
```
acodrington, line 82: re: the comment, "incorrectly see the script as being cancelled" is confusing. 

I would think that by checking too quickly, the task might still be running when we do the check and therefore it's possible that the task may not have been cancelled properly and will still write the file.

But if we wait 2 seconds, we're giving the task plenty of time to either be cancelled, or completed if something isn't work as expected, _then_ doing the check.

How we can rephrase the comment? I know it's just a comment, but it's a little confusing in an already complex scenario.

source/Octopus.Tentacle.Tests.Integration/Util/TcpTentacleHelpers/PollingResponseMessageTcpKiller.cs
```diff
@@ -50,9 +50,23 @@ public void PauseConnectionOnNextResponse(Action? callBack = null)
...
             {
-                var size = dataFromTentacle.Length;
+                numberOfWritesFromTentacleSeen++;
+
+                if (pauseConnection || killConnection)
+                {
+                    // For polling tentacle to connect it first sends some ssl stuff, then a MX control message to
+                    // register itself and finally a control PROCEED message. This means it is not until the 4th
+                    // write that we could be processing a message.
+                    if (numberOfWritesFromTentacleSeen <= 4)
```
acodrington, line 63: I don't even want to know how/why you had to figure this out 🙈

acodrington (approved):
Looks real good, should solve the issue for Cloud and has a nice test explaining the weird scenario. 🙇🏻

source/Octopus.Tentacle.Tests.Integration/Util/TcpTentacleHelpers/PollingResponseMessageTcpKiller.cs
LukeButters: It is a bit of a pain, and I am honestly not super happy. But I can't think of a better way unless we either just throw fake network exceptions OR we introduce a halibut level proxy to man in the middle the test.

source/Octopus.Tentacle.Tests.Integration/ClientScriptExecutionWillAttemptACancelCallWhenGetStatusFails.cs
acodrington: Yeah that's a big improvement - it is closer to the real world scenario and doesn't involve waiting for some arbitrary amount of time.

Very nicely done.

source/Octopus.Tentacle.Tests.Integration/Util/TcpTentacleHelpers/PollingResponseMessageTcpKiller.cs
acodrington: I think the comment to explain the magic number is good enough for our purposes.

acodrington:
[sc-111052]

PR #1106, opened 2025-05-27
Use existing helper to kill the connection at the right time

LukeButters:
# Background

We already head something that sets up the connection to be killed at just the right time.

Lets use that instead of trying to see when the connection is setup from the port forwarder.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


acodrington (approved):
Nice!

acodrington:
[sc-111052]

PR #1107, opened 2025-05-29
Fix k8s worker running as non-root

liam-mackie:
# Background

When the kubernetes agent runs as a worker, a secret for helm is mounted into the `/root/agent_upgrade/` folder. The `/root` folder has a default ownership of `700`, which means that when running as a non-root user (I.E. in OpenShift), the agent upgrade and health check would fail with `EACCESS` whenever attempting to run helm.

# Results

This fixes this bug, making the `/root` folder accessible by the root _group_ as well.


liam-mackie:
Ignore - this is the wrong image to do this in.

PR #1114, opened 2025-07-30
Indicate disabled tenants in setup dropdown

wlthomson:
# Background

[sc-https://app.shortcut.com/octopusdeploy/story/104616/sev-3-disabled-tenants-appearing-on-tentacle-setup-page-requested-by-lewis-johnson]

# Results

Fixes https://github.com/OctopusDeploy/Issues/issues/9273.

## Before

TODO.

## After

TODO.

# Pre-requisites

### DIFF
--- a/source/Octopus.Manager.Tentacle/TentacleConfiguration/SetupWizard/SetupTentacleWizardModel.cs
+++ b/source/Octopus.Manager.Tentacle/TentacleConfiguration/SetupWizard/SetupTentacleWizardModel.cs
@@ -777,7 +777,7 @@ namespace Octopus.Manager.Tentacle.TentacleConfiguration.SetupWizard
                 AreTenantsSupported = spaceSpecificData.AreTenantsSupported;
                 PotentialTenantTags = spaceSpecificData.TenantTags.SelectMany(tt => tt.Tags.Select(t => t.CanonicalTagName)).ToArray();
                 UpdateSelection(SelectedTenantTags, PotentialTenantTags);
-                PotentialTenants = spaceSpecificData.Tenants.Select(tt => tt.Name).ToArray();
+                PotentialTenants = spaceSpecificData.Tenants.Select(tt => tt.IsDisabled ? $"{tt.Name} (disabled)" : tt.Name).ToArray();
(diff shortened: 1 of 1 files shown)


hnrkndrssn (approved):

source/Octopus.Manager.Tentacle/TentacleConfiguration/SetupWizard/SetupTentacleWizardModel.cs
```diff
@@ -777,7 +777,7 @@ void UpdateTenants()
                 AreTenantsSupported = spaceSpecificData.AreTenantsSupported;
                 PotentialTenantTags = spaceSpecificData.TenantTags.SelectMany(tt => tt.Tags.Select(t => t.CanonicalTagName)).ToArray();
                 UpdateSelection(SelectedTenantTags, PotentialTenantTags);
-                PotentialTenants = spaceSpecificData.Tenants.Select(tt => tt.Name).ToArray();
+                PotentialTenants = spaceSpecificData.Tenants.Select(tt => tt.IsDisabled ? $"{tt.Name} (disabled)" : tt.Name).ToArray();
```
benPearce1, line 780: When you select a disabled tenant does it actually generate the correct value in the script output?

benPearce1: The proper way to do this would be to change all the usages of `PotentialTenants` over to a class that holds the name and disabled flag in different properties, the selected value is still just the name, but the view can render the disabled property seperately

wlthomson: @benPearce1 First suggestion aligns with my thinking, too. Plan is to go with the first approach and lift the "display name" logic into a method on the new class.

wlthomson: @benPearce1 Tested new approach and the script has the correct value now 👍 .

PR #1115, opened 2025-08-04
feat: Add Pod Labels as a configurable option

liam-mackie:
# Background

This PR adds the ability to define extra labels on the script pods, which is required to support [Entra workload identities](https://learn.microsoft.com/en-us/azure/aks/workload-identity-overview?tabs=dotnet#pod-labels).

# Results

Fixes: https://github.com/OctopusDeploy/Issues/issues/9543

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle/Kubernetes/KubernetesScriptPodCreator.cs
+++ b/source/Octopus.Tentacle/Kubernetes/KubernetesScriptPodCreator.cs
@@ -199,11 +199,7 @@ namespace Octopus.Tentacle.Kubernetes
                 {
                     Name = podName,
                     NamespaceProperty = KubernetesConfig.Namespace,
-                    Labels = new Dictionary<string, string>
-                    {
-                        ["octopus.com/serverTaskId"] = command.TaskId,
-                        ["octopus.com/scriptTicketId"] = command.ScriptTicket.TaskId
-                    },
+                    Labels = GetScriptPodLabels(tentacleScriptLog, command),
                     Annotations = ParseScriptPodAnnotations(tentacleScriptLog)
                 },
                 Spec = new V1PodSpec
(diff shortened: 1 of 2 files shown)


eddymoulton (approved):

source/Octopus.Tentacle/Kubernetes/KubernetesScriptPodCreator.cs
```diff
@@ -419,6 +415,25 @@ V1Affinity ParseScriptPodAffinity(InMemoryTentacleScriptLog tentacleScriptLog)
...
+            {
+                ["octopus.com/serverTaskId"] = command.TaskId,
+                ["octopus.com/scriptTicketId"] = command.ScriptTicket.TaskId
+            };
+            var extraLabels = ParseScriptPodJson<Dictionary<string, string>>(
+                tentacleScriptLog,
+                KubernetesConfig.PodLabelsJson,
+                KubernetesConfig.PodLabelsJsonVariableName,
+                "pod labels");
+            labels.AddRange(extraLabels);
```
eddymoulton, line 430: I'm assuming that an empty env var results in an empty list that all just get's ignored?

liam-mackie: Yep - that's what happens in the annotations code at least 😅

APErebus (changes requested):

source/Octopus.Tentacle/Kubernetes/KubernetesScriptPodCreator.cs
liam-mackie: I thought that `ParseScriptPodJson` returned an empty, but instantiated dict. That should end up with just the source, right?

liam-mackie: Never mind, it does return null, but we're safe because our addRange allows for it. For now, I've added a gate for null just in case we ever change that implementation.

PR #1135, opened 2025-08-31
Add telepresence script

liam-mackie:
This is a small script to use telepresence to remote into a cluster and act as if tentacle is running from that cluster. This is great for troubleshooting permission issues and anything to do with service accounts.


APErebus (approved):
LGTM

PR #1141, opened 2025-09-11
Enable multi architecture compatibility between Tentacle and Script pods

octonautcal:
**_Are you a customer of Octopus Deploy? Please contact [our support team](https://octopus.com/support) so we can triage your PR, so that we can make sure it's handled appropriately._**

## Background
The boostrapRunner executable is a Go binary that is built as part of the Tentacle image, and copied over to the script pod at runtime. It is used to wrap the script execution so the pod streams logs in a format that Tentacle can parse. It is not built as part of the worker-tools image because customers are free to specify their own worker-tools image, and we need to ensure the logs are in the correct format.

## Problem
The current bootstrapRunner exe is tightly coupled with the architecture of the Tentacle container. The limitation exists because we only build the bootstrapRunner exe for the arch that the tentacle image is built on.

RUN go build -ldflags "-s -w" -o "bin/bootstrapRunner"

This becomes a problem when the Tentacle and script pods are running on different platform architectures - i.e. when the script pod is spun up on a different node. In such a scenario the script pod would receive a bootstrapRunner exe that it cannot execute.

# Results

## Testing strategy

1. Set up a multi arch node AKS [here](https://portal.azure.com/?ocid=AIDcmmxbrcqs76_SEM__k_CjwKCAiAzZ_NBhAEEiwAMtqKy6Ta-l-wNY6Mo0BCvZcvoaB2U6NyzqIePPgSFWtLaMglheCNv-JBPhoClQAQAvD_BwE_k_#@octopusdeploysandbox.onmicrosoft.com/resource/subscriptions/5d50bb3d-0ab7-4bac-86d6-d53a8756a69f/resourcegroups/md-argocd-in-octopus_group/providers/Microsoft.ContainerService/managedClusters/multi-arch-nodes/overview)

2. Create a worker on https://team-md-argocd-in-octopus.testoctopus.app/app#/Spaces-263/infrastructure/workers/Workers-81/settings. 

3. Test that it normally fails due to incorrect arch

Screenshot 2026-03-05 at 2 27 51 pm
Screenshot 2026-03-05 at 2 41 47 pm

4. Deployed a multiarch image(containing our fixes) to an Azure Container Registry linked it to the Team MD sandbox cluster. 

5. Set updated image to be updated in the helm release

6. Test that it now correctly works regardless of architecture

Screenshot 2026-03-06 at 4 23 47 pm

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


docker/kubernetes-agent-tentacle/Dockerfile
```diff
@@ -9,8 +9,19 @@ ARG TARGETOS
 COPY docker/kubernetes-agent-tentacle/bootstrapRunner/* /bootstrapRunner/
 WORKDIR /bootstrapRunner
 
+# Create bin directory for multiple architectures
+RUN mkdir -p bin
+
+# Build for multiple architectures
 # Note: the given ldflags remove debug symbols
-RUN go build -ldflags "-s -w" -o "bin/bootstrapRunner"
+RUN GOOS=linux GOARCH=amd64 go build -ldflags "-s -w" -o "bin/bootstrapRunner-linux-amd64" \
```
rain-on: change this to a loop?
eg:
```FROM alpine:latest
ARG items="item1 item2 item3"
RUN for item in $items; do \
    echo "Processing $item"; \
    # Your commands here \
done;
```

docker/kubernetes-agent-tentacle/bootstrapRunner/execute-bootstrapRunner.sh
```diff
@@ -0,0 +1,39 @@
...
+        echo "Error: Unsupported architecture: $ARCH" >&2
+        echo "Supported architectures: x86_64, aarch64, arm64, i386, i686, armv7l, armv6l" >&2
+        exit 1
+        ;;
+esac
+
+# Construct binary name
+BINARY_NAME="bootstrapRunner-${OS}-${GO_ARCH}"
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
```
rain-on: There's a BIT of a smell here, which is that this script assumes the bootstrapRunner is in the same directory as this script.

This will work, but only because we _know_ we put the boostrapRunner binaries (via the changes we made to the dockerfile).

It is WAY nicer if we can eliminate implicit assumptions - and codify this relationship instead.

It'll make life easier for future generations.

I'd propose that this script pull the "BootstrapBinaryPath" from an existing environment variable (then maybe fallback to localdirectory if that doesn't exist) - this env-var would be set by the Tentacle, which would _know_ where it had copied the binaries too.

This would remove the requirement that the script and binaries be in the same location.

rain-on (approved):
LGTM

APErebus (approved):
Non-blocking minor things

source/Octopus.Tentacle/Kubernetes/KubernetesScriptPodCreator.cs
```diff
@@ -179,8 +179,12 @@ async Task CreatePod(StartKubernetesScriptCommandV1 command, IScriptWorkspace wo
 
             LogVerboseToBothLogs($"Creating Kubernetes Pod '{podName}'.", tentacleScriptLog);
 
-            workspace.CopyFile(KubernetesConfig.BootstrapRunnerExecutablePath, "bootstrapRunner", true);
-
+            foreach(var file in kubernetesPhysicalFileSystem.EnumerateFiles(KubernetesConfig.BootstrapRunnerExecutablePath))
```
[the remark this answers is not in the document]
octonautcal: updated in [78ecdc6](https://github.com/OctopusDeploy/OctopusTentacle/pull/1141/commits/78ecdc6e85d1ea79167b5a6acd52976c0b5a0470)

PR #1147, opened 2025-09-24
Support retries for long running RPCs. With this We will optionally retry a failed long running file upload/download.

LukeButters:
# Background

Previously we used to attempt file uploads 6 times. This is a problem since for a tentacle that is offline:
* For listening tentacles we might do all 6 retries in under 20s (or something like that).
* For polling tentacles we would wait 12 minutes as we place the request on the queue 6 times with a 2 minute timeout.

This was replaced with retry for up to 2m30s by default.

This has one draw back if we are uploading a file for 5 minutes then have a minor network disruption we would not retry since 5 minutes exceeds 2m30s.

With the Redis Pending Request Queue the likely hood of requests getting interrupted increases (for polling tentacles only). With this PR the Tentacle Client supports the idea of a min number of attempts for interrupted calls:
```
        /// <summary>
        /// This is the minimum number of attempts that will be made to long-running calls,
        /// that are interrupted while executing.
        /// For example, if a long-running file upload is interrupted after ten minutes of transfering,
        /// and this is set to 2, we will make another attempt to upload the file even if the RetryTimeout
        /// is exceeded.
        /// If this is set to 9999, and a polling tentacle did not collect the file upload request from the queue,
        /// we will not continue to make attempts until we have made 9999 attempts. This is because the RPC call
        /// is not considered to be interrupted as the RPC call never started. This means we won't try 9999 times
        /// to send a request to a tentacle that is not connected.
        /// </summary>
        public int? MinimumAttemptsForInterruptedLongRunningCalls { get; }
```

This means if we upload a file for 5 minutes, the transfer is interrupted we will make another attempt to send the file. If tentacle re-connects back in it will re-execute the file upload.

This will only occur if the upload has started. Which means we don't have that "tentacle offline retry for 12min problem as described above".

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Client.Tests/RpcCallRetryHandlerFixture.cs
+++ b/source/Octopus.Tentacle.Client.Tests/RpcCallRetryHandlerFixture.cs
@@ -838,8 +838,6 @@ namespace Octopus.Tentacle.Client.Tests
         public async Task WhenConfiguredToMakeAMinimumNumberOfAttempts_AndTheFirstAttemptExceedsTheRetryDuration_AndTheFailureIsAConnectingFailure_ARetryIsNotMade()
         {
             var callCount = 0;
-            var onRetryActionCalled = false;
-            var onTimeoutActionCalled = false;
 
             // Short timeout to ensure it's exceeded, but minimumAttemptsForInterruptedLongRunningCalls = 2
             var handler = new RpcCallRetryHandler(TimeSpan.FromSeconds(2), minimumAttemptsForInterruptedLongRunningCalls: 9999);
@@ -860,19 +858,15 @@ namespace Octopus.Tentacle.Client.Tests
                     onRetryAction: async (_, _, _, _, _, _) =>
                     {
                         await Task.CompletedTask;
-                        onRetryActionCalled = true;
                     },
                     onTimeoutAction: async (_, _, _, _) =>
                     {
                         await Task.CompletedTask;
-                        onTimeoutActionCalled = true;
                     },
                     CancellationToken.None));
             exception.Should().BeAssignableTo<HalibutClientException>();
             
             callCount.Should().Be(1);
-            onRetryActionCalled.Should().BeTrue();
-            onTimeoutActionCalled.Should().BeTrue();
         }
     }
 }


source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
```diff
@@ -84,7 +84,7 @@ public async Task<T> ExecuteWithRetries<T>(
 
                             onErrorAction?.Invoke(lastException);
 
-                            var remainingDurationInSeconds = (int)(totalRetryDuration - elapsedDuration).TotalSeconds;
+                            var remainingDurationInSeconds = Math.Max((int)(totalRetryDuration - elapsedDuration).TotalSeconds, 0);
```
LukeButters, line 87: This could be negative some time.

source/Octopus.Tentacle.Client/Execution/RpcCallExecutor.cs
```diff
@@ -84,7 +84,7 @@ public async Task<T> ExecuteWithRetries<T>(
...
                             logger.Info($"An error occurred communicating with Tentacle. This action will be retried after {(int)sleepDuration.TotalSeconds} seconds. Retry attempt {retryCount}. Retries will be performed for up to {remainingDurationInSeconds} seconds.");
```
LukeButters, line 88: This logging could be better, we probably should be taking into account the min number of retries.

I am opting not to do it, it would be ideal but the value it provides is to low given the time remaining on the project.

source/Octopus.Tentacle.Client/TentacleClientOptions.cs
```diff
@@ -6,10 +6,13 @@ namespace Octopus.Tentacle.Client
     public class TentacleClientOptions
     {
         public RpcRetrySettings RpcRetrySettings { get; }
+        
+        public int? MinimumAttemptsForInterruptedLongRunningCalls { get; }
```
LukeButters, line 10: We will be able to feature toggle this 👍

source/Octopus.Tentacle.Client.Tests/RpcCallRetryHandlerFixture.cs
```diff
@@ -687,5 +688,194 @@ private TimeSpan GetMinTimeoutDuration(RpcCallRetryHandler handler)
         {
             return handler.RetryTimeout - handler.RetryIfRemainingDurationAtLeast - retryBackoffBuffer;
         }
+
+        [Test]
+        public async Task MinimumAttempts_WhenSetTo1_DoesNotRetryAfterTimeoutExceeded()
```
sburmanoctopus, line 693: Is this test is testing what happens after we timeout, where does it do the timeout? In other tests, we delay in the lambda (that increments `callCount`). Am I missing something?

source/Octopus.Tentacle.Client.Tests/RpcCallRetryHandlerFixture.cs
```diff
@@ -687,5 +688,194 @@ private TimeSpan GetMinTimeoutDuration(RpcCallRetryHandler handler)
...
+            retryCount.Should().Be(2);
+            onTimeoutActionCalled.Should().BeFalse(); // Should succeed before timeout
+        }
+        
+        /// <summary>
+        /// Connection error means the tentacle did not get the request, which means the tentacle is considered to be offline.
+        /// This shows that we won't exceed the retry duration attempting to connect to an offline tentacle to meet the
+        /// minimumAttemptsForInterruptedLongRunningCalls count.
+        /// </summary>
+        /// <exception cref="HalibutClientException"></exception>
```
sburmanoctopus, line 833: Not: why the exception tag? Remove?

source/Octopus.Tentacle.Client.Tests/RpcCallRetryHandlerFixture.cs
```diff
@@ -687,5 +688,194 @@ private TimeSpan GetMinTimeoutDuration(RpcCallRetryHandler handler)
...
+            var handler = new RpcCallRetryHandler(TimeSpan.FromSeconds(5), minimumAttemptsForInterruptedLongRunningCalls: 9999);
+
+            try
+            {
+                await handler.ExecuteWithRetries(
+                async ct =>
+                {
+                    callCount++;
+                    
+                    // Delay 2 seconds to ensure the ct doesn't get canceled.
```
sburmanoctopus, line 851: This comment says 2 seconds when it's 1 in the code.

source/Octopus.Tentacle.Client.Tests/RpcCallRetryHandlerFixture.cs
LukeButters: fixed sleeping in the lambda

PR #1174, opened 2026-01-06
Skip disk space check when getting workspace for cleanup

LukeButters:
## Summary

We often see failures because we are unable to clean up after a script because the "disk space check" has failed. Meaning we don't cleanup because the tentacle is low on disk space.

Since that is insane, this PR changes tentacle so that it can cleanup and check status of scripts without checking if the tentacle is low on disk space.

Also this allows a script to be cancelled and cleaned up when it is low on disk space.

See: https://octopusdeploy.slack.com/archives/CG4JP65N2/p1762406103092919?thread_ts=1762404506.174969&cid=CG4JP65N2

## AI Summary
- Fixed catch-22 where low disk space prevented cleanup operations that would free up space
- Added `WorkspaceReadinessCheck` enum with explicit `Perform`/`Skip` values
- Updated `IScriptWorkspaceFactory.GetWorkspace()` to require readiness check specification
- Updated all call sites to explicitly choose whether disk space validation is needed

## Problem
When completing/cleaning up a script, the system was calling `GetWorkspace()` which always performed a disk space check via `CheckReadiness()`. When disk space was low, this check would throw an `IOException` preventing:
1. Cleanup operations that would actually **free up the needed space**
2. Status checks from reading logs to report progress
3. Cancel operations from accessing workspace to report cancellation status
4. Any read-only operations that don't require disk space

## Solution
Replaced implicit disk space checking with explicit control via a `WorkspaceReadinessCheck` enum:
- **`WorkspaceReadinessCheck.Skip`**: For cleanup operations (freeing space) and read-only operations (reading logs, checking state)
- **`WorkspaceReadinessCheck.Perform`**: For write operations that require disk space (starting scripts, writing files)

The enum parameter is **required** (not optional), making it explicit at every call site whether space validation is needed.

## Changes
- **New file**: `WorkspaceReadinessCheck.cs` - Enum with `Perform` and `Skip` values
- `IScriptWorkspaceFactory.cs` - Changed signature to require `WorkspaceReadinessCheck` parameter
- `ScriptWorkspaceFactory.cs` - Updated implementation to conditionally check readiness based on enum
- `ScriptService.cs` - Skip check in `CompleteScriptAsync` and `GetResponse` (V1)
- `ScriptServiceV2.cs` - Skip check in `CompleteScriptAsync` and `GetResponse`; perform check when starting scripts (V2)
- `KubernetesScriptServiceV1.cs` - Skip check in `CompleteScriptAsync` (Kubernetes)
- `ScriptPodLogEncryptionKeyProvider.cs` - Perform check when writing keys, skip when reading
- `ScriptPodSinceTimeStore.cs` - Perform check (used for both read and write)
- Test fixtures updated to explicitly specify readiness check behavior

## Test plan
- [ ] Verify cleanup succeeds when disk space is below threshold
- [ ] Verify status checks work when disk space is below threshold
- [ ] Verify cancel operations work when disk space is below threshold
- [ ] Verify normal workspace operations (create/prepare) still perform space checks
- [ ] Test all three script service variants (V1, V2, Kubernetes)
- [ ] Confirm workspace cleanup frees up space successfully
- [ ] Verify read-only operations (logs, status) work under low disk space

### DIFF
--- a/source/Octopus.Tentacle.Core/Services/Scripts/WorkSpace/ScriptWorkspaceFactory.cs
+++ b/source/Octopus.Tentacle.Core/Services/Scripts/WorkSpace/ScriptWorkspaceFactory.cs
@@ -56,7 +56,7 @@ namespace Octopus.Tentacle.Scripts
             List<ScriptFile> files,
             CancellationToken cancellationToken)
         {
-            var workspace = GetWorkspace(ticket);
+            var workspace = GetWorkspace(ticket, WorkspaceReadinessCheck.Perform);
             workspace.IsolationLevel = isolationLevel;
             workspace.ScriptMutexAcquireTimeout = scriptMutexAcquireTimeout;
             workspace.ScriptArguments = scriptArguments;


LukeButters:
I am not sure we why do a disk space check on new scripts maybe a good reason exists, so I left that.

If we try catch then we wont have a workspace to call delete on so we can't clean it with that approach.

PR #1175, opened 2026-01-09
Make running K8S Agent locally a little easier

zentron:
When running the agent locally, ill often have any number of different kubectl contexts and namespaces that I want the agent to connect with.

This PR adds a little more configuration support based on env variables

### DIFF
--- a/source/Octopus.Tentacle/Kubernetes/LocalMachineKubernetesClientConfigProvider.cs
+++ b/source/Octopus.Tentacle/Kubernetes/LocalMachineKubernetesClientConfigProvider.cs
@@ -19,25 +19,38 @@ namespace Octopus.Tentacle.Kubernetes
             {
                 return GetTelepresenceConfig(telepresenceRoot);
             }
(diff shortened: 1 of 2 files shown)


source/Octopus.Tentacle/Kubernetes/LocalMachineKubernetesClientConfigProvider.cs
```diff
@@ -70,7 +83,7 @@ KubernetesClientConfiguration GetTelepresenceConfig(string telepresenceRoot)
                 SslCaCerts = certificates,
             };
             
-            var namespaceVar = Environment.GetEnvironmentVariable("OCTOPUS__K8STENTACLE__NAMESPACE");
+            var namespaceVar = Environment.GetEnvironmentVariable(KubernetesConfig.NamespaceVariableName);
```
flin-8: I think only the namespace configuration is duplicated across the Telepresence and normal configurations. When/if we have more duplication then it's worth extracting out the common stuff

PR #1183, opened 2026-02-11
Remove unused method, expose tentacle environment variables into Tentacle Core, and add a method for defining the tentacle home env var.

LukeButters:
# Background

Tentacle Simulator and probably the in process Tentacle have a bug where the `TentacleHome` directory does not match what is returned by `IHomeDirectoryProvider`. This causes an issue where package deltas are never sent.

This PR aims to improve the situation by making it easier to setup an in process tentacle it does this by:
* Removing a method that is never used, and so doesn't need to be set in Tentacle Simulator or the in process Tentacle.
* Moving the environment variables to Tentacle Core, which will allow Simulator/in-proc Tentacle to reference those rather than hard code.
* Creating a method that defines how the `TentacleHome` environment variable is defined, preventing the issue seen in Tentacle Simulator. Specifically where the value of `TentacleHome` env var did not match what is returned by `IHomeDirectoryProvider.HomeDirectory`

Note: Tentacle Core is the place we have been placing the code which can be used by the In process tentacle for integration tests as well as for use by Tentacle Simulator for development. Ideally core would be enough to get a sufficiently real Tentacle.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Manager.Tentacle/TentacleConfiguration/InstanceConfigurationLoader.cs
+++ b/source/Octopus.Manager.Tentacle/TentacleConfiguration/InstanceConfigurationLoader.cs
@@ -60,7 +60,7 @@ namespace Octopus.Manager.Tentacle.TentacleConfiguration
                 systemLog,
                 applicationInstanceRecord.ConfigurationFilePath);
 
-            var homeConfig = new HomeConfiguration(ApplicationName.Tentacle, selector);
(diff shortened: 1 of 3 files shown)


flin-8 (approved):
✅ for our files - USINGS changes only

rhysparry (approved):
💚 LGTM

PR #1186, opened 2026-02-15
Update Halibut to 8.1.1860

LukeButters:
## Summary
- Updates Halibut package version from 8.1.1707 to 8.1.1860

No change to Tentacle, brings in the latest Halibut which has support for RPCs over the queue e.g. RPC over redis. 

## Test plan
- Run the tests, because we have confidence in the tests.


julietrb1 (approved):
# 🍒

Approving in the understanding that this is an innocuous change 🙃

PR #1189, opened 2026-02-23
Fix flaky test: k8s script v1 integration test

todthomson:
# Background

Because of an historically intermittent (flaky) test run failure, which triggers alerts to @OctopusDeploy/team-executions-foundations in test:

```
Octopus.Tentacle.Kubernetes.Tests.Integration
.KubernetesAgent.KubernetesScriptServiceV1IntegrationTest.RunSimpleScript()
```

caused by:

```
Server exception: 
System.IO.EndOfStreamException: Attempted to read past the end of the stream.
at Halibut.Transport.Protocol.MessageSerializer
.ReadMessageAsync[T](RewindableBufferStream stream, CancellationToken cancellationToken)
```

which results in:

```
An error occurred communicating with Tentacle.
This action will be retried after 1 seconds.
Retry attempt 1.
Retries will be performed for up to 89 seconds.
```

which subsequently results in an assertion failure:

```
Started to be 1L, but found 2L.
```

caused by the following assertion:

```
recordedMethodUsages.For(nameof(IAsyncClientKubernetesScriptServiceV1.StartScriptAsync))
.Started.Should().Be(1);
=> 2
```

but I think that's perfectly reasonable (i.e. it's reasonable to restart when loosing comms with Tentacle).

# Results

Thus, I've updated the assertion to be "at least once":

```
recordedMethodUsages.For(nameof(IAsyncClientKubernetesScriptServiceV1.StartScriptAsync))
.Started.Should().BeGreaterThanOrEqualTo(1);
```

## Before

Flaky test e.g.
https://build.octopushq.com/buildConfiguration/TeamFireAndMotion_OctopusTentacle_TentacleVLatest_net80_ChainFullChainNightly/21130454

## After

No more flaky test.

# How to review this PR

This is how Octopus behaves when the connection to Tentacle is interrupted.

Do you agree that, therefore, the assertion in the original test "as written" was wrong?

**If so, please approve.**

**If not, please talk to me and explain why (the assertion should remain as it is)?**

# Pre-requisites

Yes, this is a flaky test, I don't think it needs a GitHub issue.

Consultation plan:
1. Talk internally within my team (to see if I'm on the right track).
1. Talk to the original test author @APErebus (to confirm that I'm on the right track).

My change IS a test, and, it will run in the CI environment.

### DIFF
--- a/source/Octopus.Tentacle.Kubernetes.Tests.Integration/KubernetesAgent/KubernetesScriptServiceV1IntegrationTest.cs
+++ b/source/Octopus.Tentacle.Kubernetes.Tests.Integration/KubernetesAgent/KubernetesScriptServiceV1IntegrationTest.cs
@@ -56,14 +56,8 @@ public class KubernetesScriptServiceV1IntegrationTest : KubernetesAgentIntegrati
         result.ExitCode.Should().Be(0);
         result.State.Should().Be(ProcessState.Complete);
 
-        // re: flaky: "An error occurred communicating with Tentacle.
-        //            This action will be retried after 1 second.
-        //            Retry attempt 1.
-        //            Retries will be performed for up to 89 seconds."
-        // Note: Occasionally we see that communication with Tentacle gets interrupted,
-        //       and Tentacle does what it _should_ do (I think) and reconnects and restarts the script.
-        // Action: This seems like what you would want to happen,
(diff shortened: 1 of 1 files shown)


LukeButters (approved):

Jtango18 (approved):
Approved by MD - the logic behind it makes sense

@APErebus - if you find this abhorent, we can always reverse down the track.

PR #1194, opened 2026-03-02
Add pod affinity if ReadWriteOnce is set

APErebus:
# Background

When the storage access mode for the Kubernetes agent is set to `ReadWriteOnce` (RWO), we need to force the created script pods to be scheduled onto the same node as the tentacle pod (as they will be sharing the node's storage)

# Results

Reads the access modes from an environment variable and when it's a single value of `ReadWRiteOnce`, updates the script pod affinity to include a new `podAffinity` term that matches on the tentacle's `app.kubernetes.io/instance` and `agent.octopus.com/component` labels.

The `topologyKey` of `kubernetes.io/hostname` forces the same node

image

image

Fixes MD-1564

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


eddymoulton (approved):
The changes look good and match how you described the problem earlier.

I think it would be good to get @liam-mackie to run over this too with his context.

PR #1207, opened 2026-04-08
Fix tentacle hostname resolution on non-domain-joined machines

wlthomson:
# Background

A customer running the Windows Tentacle container in ECS on EC2 reported that their Tentacle was being registered in
Octopus with an invalid URL. They had configured `PublicHostNameConfiguration=FQDN` so the container would register
using its internal EC2 DNS hostname (e.g. `ip-xxx-xx-x-x.ec2.internal`). This seems to be the right configuration when the Octopus server and the ECS task are in the same VPC, so should be supported.

Fixes HPY-1292

# Results

The `configure-tentacle.ps1` script constructs the FQDN by concatenating `DNSHostName` and `Domain`. On a domain-joined machine this is correct (`DNSHostName` is the short name and `Domain `is the DNS suffix). However, on a non-domain-joined machine like an EC2 instance, `DNSHostName` is already fully qualified and `Domain` is empty.

This PR adds a conditional to check if `DNSHostName `already contains a dot or `Domain` is empty before attempting to join them.

# How to review this PR

- Main risk here is regressions. The conditional is intentionally restrictive to reduce the risk of changing desired existing behaviour, but the risk is still there.

Quality :heavy_check_mark:

- Unit testing this area is difficult, and adding an integration test for this use-case feels a bit disproportionate.
- If there are concerns, I can do some manual testing against a real EC2 instance (but again, time cost is high vs. the value of this change).

# Pre-requisites

### DIFF
--- a/docker/windows/Scripts/configure-tentacle.ps1
+++ b/docker/windows/Scripts/configure-tentacle.ps1
@@ -60,7 +60,14 @@ function Get-PublicHostName
   elseif ($publicHostNameConfiguration -eq "FQDN")
   {
     $computer = Get-CimInstance win32_computersystem
-    $publicHostName = "$($computer.DNSHostName).$($computer.Domain)"
+    # On domain-joined machines, DNSHostName is the short name and Domain is the DNS suffix.
+    # On non-domain-joined machines (e.g. EC2), DNSHostName is already fully qualified and
+    # Domain is empty or a non-DNS workgroup name, so use DNSHostName as-is.
+    if ($computer.DNSHostName.Contains('.') -or [string]::IsNullOrEmpty($computer.Domain)) {
+      $publicHostName = $computer.DNSHostName
+    } else {
+      $publicHostName = "$($computer.DNSHostName).$($computer.Domain)"
+    }
   }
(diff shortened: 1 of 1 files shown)


docker/windows/Scripts/configure-tentacle.ps1
```diff
@@ -60,7 +60,14 @@ function Get-PublicHostName
   elseif ($publicHostNameConfiguration -eq "FQDN")
   {
     $computer = Get-CimInstance win32_computersystem
-    $publicHostName = "$($computer.DNSHostName).$($computer.Domain)"
+    # On domain-joined machines, DNSHostName is the short name and Domain is the DNS suffix.
```
hnrkndrssn, line 63: Do we also need to make sure the Linux equivalent script functions correctly?

Currently it executes `hostname --fqdn` to get the FQDN host name.

wlthomson: From my understanding, linux should be fine as `hostname` returns the complete FQDN. This problem arises from having to manually construct it on Windows from concatenation of two separate fields (from `Win32_ComputerSystem`).

hnrkndrssn (approved):

PR #1209, opened 2026-04-09
Reduce memory consumption on logs decryption

oleksandr-codefresh:
# Background

PodLogEncryptionProvider itself is recreated on every GetLogs() call, which is called on a polling loop for each running script pod. So for a script producing lots of output, you get: N_polls × M_lines_per_poll cipher object allocations.

This lead to excessive amount of EasEngine & cipher instances, that can get into Gen 2 GC. Which potentially creates a "memory leak" during couple of days on running pod.

# Results

Fixes [YOS-30](https://linear.app/octopus/issue/YOS-30/sev-3-low-or-possible-memory-leak-in-tentaclek8s-agent)

## Before
telegram-cloud-photo-size-2-5375229125628467206-y

## After
telegram-cloud-photo-size-2-5375229125628467213-y

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


APErebus (approved):
Yeah awesome! Thanks @oleksandr-codefresh

PR #1215, opened 2026-04-30
If the test process wont die, don't fail the build just ignore it and proceed

LukeButters:
# Background

Fixes: EFT-3259
**AI summary:**
  The [ShutdownDotNetAfterServerBuildAttribute] runs dotnet build-server shutdown after every build. On the Windows 2012 build agent this process hangs, causing Nuke to throw an unhandled
  exception and exit with code 255 — even when all tests pass.

  macOS build agents are persistent (non-ephemeral) so the shutdown is still needed there to prevent compiler server processes accumulating between builds. Windows and Linux agents are
  ephemeral so processes are cleaned up when the agent tears down.

  Replaces the attribute with an OnBuildFinished override that calls DotNetBuildServerShutdown() on macOS only.

This should fix the issue seen at:

https://build.octopushq.com/buildConfiguration/TeamFireAndMotion_OctopusTentacle_TentacleVLatest_net80_IntegrationTestNet80onWindows2012net48service/22113648?buildTab=tests

where all the tests pass but we could not shutdown correctly.

# Results

Fixes https://github.com/OctopusDeploy/OctopusTentacle/issues/... _(optional public issue)_

## After

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -19,6 +19,10 @@ using Nuke.Common.Utilities.Collections;
 using Serilog;
 using static Nuke.Common.Tools.DotNet.DotNetTasks;
 
+// macOS build agents are persistent (non-ephemeral), so compiler server processes must be cleaned up between builds.
+#if MACOS
+[ShutdownDotNetAfterServerBuild]
+#endif
 partial class Build : NukeBuild
 {
     /// Support plugins are available for:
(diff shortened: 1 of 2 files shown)


LukeButters:
Will verify that the build servers are ephemeral, if they are not lets make them ephemeral then merge.

build/Build.cs
```diff
@@ -19,7 +19,10 @@
 using Serilog;
 using static Nuke.Common.Tools.DotNet.DotNetTasks;
 
+// macOS build agents are persistent (non-ephemeral), so compiler server processes may need to be cleaned up between builds.
```
LukeButters, line 22: Other agents are ephemeral so we wont need this on them.

Also we don't do this on octopus.

xwipeoutx (approved):

PR #1223, opened 2026-05-14
Fix an issue where cancelled scripts would wait for the STDOUT/STDERR pipes to be closed.

### LINKED ISSUE #1225: Tentacle is unable to cancel scripts, which have children that inherit and do not close STDOUT/STDERR when the parent script completes.
### Team

- [x] I've assigned a team label to this issue

### What happened?

When:
1. Tentacle starts a child script which starts a grandchild that inherits the STDOUT/STDERR pipes.
2. The child process completes (the child does not wait for the grandchild), and the grandchild continues on.
3. Cancellation occurs, since the parent is already completed we don't find the grandchild and so never Kill it.
4. **Tentacle waits forever** for the STDOUT/STDERR pipes to be closed. This will only happen when the grandchild dies.

### Reproduction

Reproduced in tests in the fix.

### Error and Stacktrace

```shell

```

### More Information

_No response_

### Workaround

_No response_

LukeButters:
Fixes: EFT-3354
Fixes: https://github.com/OctopusDeploy/OctopusTentacle/issues/1225

# Background

Fixes an issue where cancellation would hang forever waiting for a grandchild process that is holding open STDOUT/STDERR, but will never be killed since its parent was already completed.

The problem occurs when:
1. Tentacle starts a child which starts a grandchild that inherits the STDOUT/STDERR pipes.
2. The child process completes (the child does not wait for the grandchild), and the grandchild continues on.
3. Cancellation occurs, since the parent is already completed we don't find the grandchild and so never Kill it.
4. Tentacle waits forever for the STDOUT/STDERR pipes to be closed. This will only happen when the grandchild dies.

The Fix is: When cancellation occurs after we attempt to kill, we close the pipes.

# Results

## Before

Tests added fail.

## After

Tests added pass.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Core/Util/CommandLine/SilentProcessRunner.cs
+++ b/source/Octopus.Tentacle.Core/Util/CommandLine/SilentProcessRunner.cs
@@ -221,6 +221,21 @@ namespace Octopus.Tentacle.Util
                     error($"Failed to kill the launched process: {killProcessException}");
                 }
             }
+            finally
+            {
+                try
+                {
+                    // When cancelling, close the file handles.
+                    // If the child finishes, but the grandchild holds stdout/stderr and never completes,
+                    // THEN we won't issue a kill to the grandchild but will wait for the grandchild to
+                    // close the handles.
+                    process.Close();
+                }
+                catch (Exception ex)
+                {
+                    error($"Failed to close process resources: {ex.Message}");
+                }
+            }
         }
 
         [DllImport("kernel32.dll", SetLastError = true)]
--- a/source/Octopus.Tentacle.Tests.Integration/Util/SilentProcessRunnerFixture.cs
+++ b/source/Octopus.Tentacle.Tests.Integration/Util/SilentProcessRunnerFixture.cs
@@ -1,11 +1,16 @@
 ﻿using System;
+using System.Diagnostics;
+using System.IO;
 using System.Text;
 using System.Threading;
+using System.Threading.Tasks;
 using FluentAssertions;
 using NUnit.Framework;
+using Octopus.Tentacle.CommonTestUtils;
 using Octopus.Tentacle.Tests.Integration.Support;
 using Octopus.Tentacle.Tests.Integration.Support.TestAttributes;
 using Octopus.Tentacle.Util;
+using PlatformDetection = Octopus.Tentacle.Util.PlatformDetection;
 
 namespace Octopus.Tentacle.Tests.Integration.Util
 {
@@ -49,11 +54,11 @@ namespace Octopus.Tentacle.Tests.Integration.Util
                     cts.Token);
 
                 exitCode.Should().Be(99, "our custom exit code should be reflected");
-                
+
                 // It seems the encoding can change during the lifetime of the OS so don't expect it wont change in tests.
                 debugMessages.ToString().Should().ContainEquivalentOf($"Starting {command} in working directory '' using '");
                 debugMessages.ToString().Should().ContainEquivalentOf($"' encoding running as '{TestEnvironmentHelper.CurrentUserName}'");
-                
+
                 errorMessages.ToString().Should().BeEmpty("no messages should be written to stderr");
                 infoMessages.ToString().Should().BeEmpty("no messages should be written to stdout");
             }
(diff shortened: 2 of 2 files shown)


source/Octopus.Tentacle.Core/Util/CommandLine/SilentProcessRunner.cs
```diff
@@ -128,17 +128,30 @@ void WriteData(Action<string> action, ManualResetEventSlim resetEvent, DataRecei
...
+                               if (Volatile.Read(ref asyncReadsStarted))
+                                   try { process.Close(); } catch (Exception ex) { error($"Failed to close process resources: {ex.Message}"); }
                            }))
                     {
                         if (cancel.IsCancellationRequested)
                             DoOurBestToCleanUp(process,  error);
 
                         process.BeginOutputReadLine();
                         process.BeginErrorReadLine();
+                        Volatile.Write(ref asyncReadsStarted, true);
```
LukeButters: The AI says:

```
● asyncReadsStarted is written by one thread (the thread running ExecuteCommand) and read by another (whoever fires cancel.Cancel(), which invokes the Register callback synchronously on the cancelling thread). It's a textbook cross-thread bool flag.

  Without any synchronisation:

  - JIT can hoist the read. The compiler could cache the captured local in a register and not re-fetch it. Less likely in a one-shot lambda, but not guaranteed.
  - CPU can reorder. On weak memory models (ARM — the project's RIDs include linux-arm64/linux-arm), a plain read might not observe a prior write from another core without a memory fence.

  Volatile.Write does a release fence after writing true; Volatile.Read does an acquire fence before reading. That pairs to give you the happens-before guarantee you need: if the writer set it before the cancel fired, the callback is guaranteed to see true.

  Why not other options:

  - volatile keyword — only allowed on fields, not locals. asyncReadsStarted is a captured local, so it's lifted into a compiler-generated closure class field that you can't decorate.
  - Interlocked — doesn't support bool. You'd have to widen to int, which is uglier for no real benefit here.
  - lock — overkill for a one-way flag.

  Honest caveat: in this specific case it's belt-and-braces. CancellationTokenSource.Cancel() internally uses interlocked operations that act as a full fence, so if Thread A's write happens before Cancel(), other threads observing the cancelled state will also see the write
  transitively. So a plain bool would probably work in practice. But Volatile.Read/Write makes the cross-thread intent explicit, is cheap, and is the correct portable answer that doesn't rely on knowing CTS internals — worth the two lines.
```

source/Octopus.Tentacle.Tests.Integration/Util/SilentProcessRunnerFixture.cs
```diff
@@ -118,6 +121,216 @@ public void CancellationToken_ShouldForceKillTheProcess()
...
+            //      handles, then immediately exits.
+            //   3. PowerShell waits for cmd to exit, looks up ping's PID via WMI
+            //      (filtered by ParentProcessId=cmd-PID), writes it for cleanup, exits.
+            //
+            // Result on the WaitForExit thread: PowerShell is dead, cmd is dead, ping
+            // is alive holding the write-ends of the pipes. Process.Kill(true) walks
+            // descendants by ParentProcessId and finds NOTHING (cmd's PID is gone, so
+            // ping no longer appears as a descendant). Without the fix that closes
+            // StandardOutput/StandardError after Kill, WaitForExit() hangs forever.
+            var psScript = @"
```
LukeButters, line 156: yep AI generated.
The important thing is that the test failed before hand, which the AI tells me it did :D

xwipeoutx:
Ensure you make a public issue on this one

source/Octopus.Tentacle.Core/Util/CommandLine/SilentProcessRunner.cs
```diff
@@ -128,17 +128,30 @@ void WriteData(Action<string> action, ManualResetEventSlim resetEvent, DataRecei
...
+                               DoOurBestToCleanUp(process, error);
+                               // Close our end of the redirected stdout/stderr pipes
+                               // so the parameterless Process.WaitForExit() — which
+                               // waits for the async readers to drain — can return
+                               // even when a grandchild process inherited the
+                               // write-ends and outlives Kill. Without this, the
+                               // active ReadAsync calls never see EOF and the call
+                               // hangs forever. Only valid once BeginOutputReadLine
+                               // has wired up the AsyncStreamReaders.
+                               if (Volatile.Read(ref asyncReadsStarted))
```
xwipeoutx: [formatting] Can you improve the formatting here? Jamming a whole try/catch onto a single line to avoid a brace is ... a decision.

source/Octopus.Tentacle.Core/Util/CommandLine/SilentProcessRunner.cs
```diff
@@ -128,17 +128,30 @@ void WriteData(Action<string> action, ManualResetEventSlim resetEvent, DataRecei
                     process.Start();
 
                     var running = true;
```
xwipeoutx, line 130: I ran this through Claude and I think `running` has similar problems... can `running` get hoisted and cause issues? 

Honestly I'm struggling to follow along here, and this is still in draft so don't want to spend much time on it.

source/Octopus.Tentacle.Core/Util/CommandLine/SilentProcessRunner.cs
```diff
@@ -128,17 +128,30 @@ void WriteData(Action<string> action, ManualResetEventSlim resetEvent, DataRecei
...
+                               // hangs forever. Only valid once BeginOutputReadLine
+                               // has wired up the AsyncStreamReaders.
+                               if (Volatile.Read(ref asyncReadsStarted))
+                                   try { process.Close(); } catch (Exception ex) { error($"Failed to close process resources: {ex.Message}"); }
                            }))
                     {
                         if (cancel.IsCancellationRequested)
                             DoOurBestToCleanUp(process,  error);
 
                         process.BeginOutputReadLine();
```
xwipeoutx, line 140: There's still race conditions here - we can cancel while we're beginning still.  Again dunno if this is a problem... I think this reduces the window, but I'm not convinced the volatile gets us much.

xwipeoutx (commented):
I feel like we're solving the wrong problem... the control flow itself is kinda borked, right? There's chances for stale bools all over it.

Like, should `DoOurBestToCleanUp` call `Close()` after killing it? It'll cause the `BeginOutputReadLine` to throw but... who cares? We're dead anyway?

I think I'd need to dive in further to actually understand it

xwipeoutx (commented):
So much simpler!

xwipeoutx (approved):

PR #1236, opened 2026-05-25
Make `SilentProcessRunner.Execute` async

jimmyp:
## Why

Tentacle needs to release the script-isolation mutex even when a script can't be cancelled. Philips is running into situations where CrowdStrike and Rapid7 contention freezes the script running on the Tentacle and our cancel signal can't shake it loose. The only recovery today is restart tentacle.

To make this so in a follow up PR `SilentProcessRunner.ExecuteCommand` becomes async. Which makes everything that calls it async too.

This PR just adds the changes to method signatures that need to become async, to make the next riskier PR easier to review.

## What changed

Replaces `SilentProcessRunner.ExecuteCommand` with `ExecuteCommandAsync`, plus the matching interfaces (`ISilentProcessRunner`, `ICommandLineRunner`) and their wrappers. All ~20 callers in production, integration tests, and test scaffolding.

There are 5 places we synchronously call these new async call paths, each documented well in code:

- `KubernetesDirectoryInformationProvider.GetPathUsedBytes` (only called via an `IOctopusFileSystem` sync override)
- `PowerShellPrerequisite.Check` (`IPrerequisite` is sync, used by the WPF installer)
- `CommandLineRunner.Execute` (sync wrapper for the WPF installer flow)
- `LinuxServiceConfigurator` and `WindowsServiceConfigurator` (Topshelf's `ServiceCommand → AbstractCommand → ICommand.Start` chain is sync top to bottom)

## Risks & mitigations

The classic sync-over-async deadlock is the obvious worry here. `.GetAwaiter().GetResult()` hangs when the calling thread has captured a `SynchronizationContext`. See each call site for why it is safe.

Mock vs. production drift in tests was a surprising issue I ran intodue to places we mock `Execute` we needed to switch this to  `ExecuteAsync`. We hit this on `WhenSettingUpPollingTentacle_TelemetryEventShouldBeSent` and the wizard model builder now stubs both.

## For Kubernetes agent owners

This PR touches your ownership area but should cause no behaviour changes.

- `KubernetesDirectoryInformationProvider`: exposes sync (existing) and async (new) ways to read path usage. Same `du` invocation, same 30-second cache; we need both because one downstream is sync and another is async.
- `KubernetesPhysicalFileSystem`: same shape. Sync `GetStorageInformation` stays, new async `GetStorageInformationAsync` added. Both go through one private helper so the two never drift.
- `KubernetesScriptPodCreator.CreateScriptContainer`: previously blocked on async work indirectly. Now `await`s the new async sibling directly. No behaviour change, just one fewer sync-over-async hop on your script-pod creation path.
- `KubernetesAgentInstaller.Dispose` (integration test scaffolding only): still blocks on async because `Dispose` is sync.

What we'd love your eyes on: there's one remaining sync-over-async hop in the K8s agent at `EnsureDiskHasEnoughFreeSpace`, because it implements a sync filesystem interface. It blocks on the async `du` path. This is safe in a normal .NET console app because there's no `SynchronizationContext` for the awaited continuation to come back to. If you know of anything in the K8s agent's host startup (pod watchers, bootstrap, a custom task scheduler) that would install a `SynchronizationContext` and break that assumption, the disk-space check is where the deadlock would surface. Worth a quick sanity check against your mental model of how the agent boots up.

## How to review

- Is this change safe?
- Is there anything I you would do to de-risk it further beyond a green build?

### DIFF
--- a/source/Octopus.Manager.Tentacle/PreReq/PowerShellPrerequisite.cs
+++ b/source/Octopus.Manager.Tentacle/PreReq/PowerShellPrerequisite.cs
@@ -1,6 +1,7 @@
 using System;
 using System.IO;
 using System.Threading;
+using System.Threading.Tasks;
 using Octopus.Tentacle.Diagnostics;
 using Octopus.Tentacle.Util;
 
@@ -10,9 +11,18 @@ namespace Octopus.Manager.Tentacle.PreReq
     {
         public string StatusMessage => "Checking that Windows PowerShell 2.0 is installed...";
 
+        // We're at the WPF installer prerequisite boundary. IPrerequisite.Check() must return
+        // synchronously, so this is the sync-over-async bridge: a one-line wrapper over the
+        // private async implementation. Safe because the installer dispatches us on a plain
+        // thread-pool worker. No captured SynchronizationContext, so no deadlock.
+        // See https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html
         public PrerequisiteCheckResult Check()
+            => CheckAsync().GetAwaiter().GetResult();
+
+        async Task<PrerequisiteCheckResult> CheckAsync()
         {
-            return CheckPowerShellIsInstalled(out var commandLineOutput)
+            var (installed, commandLineOutput) = await CheckPowerShellIsInstalledAsync();
+            return installed
                 ? PrerequisiteCheckResult.Successful()
                 : PrerequisiteCheckResult.Failed("Windows PowerShell 2.0 or above does not appear to be installed and on the System Path on this machine. Please install Windows PowerShell or add it to the System Path then re-run this setup tool.",
                     helpLink: "http://g.octopushq.com/HowDoIInstallPowerShell",
@@ -20,44 +30,39 @@ namespace Octopus.Manager.Tentacle.PreReq
                     commandLineOutput: commandLineOutput);
         }
 
-        static bool CheckPowerShellIsInstalled(out string commandLineOutput)
+        static async Task<(bool installed, string commandLineOutput)> CheckPowerShellIsInstalledAsync()
         {
             var stdOut = new StringWriter();
             var stdErr = new StringWriter();
 
             const string powerShellExe = "powershell.exe";
             const string arguments = "-NonInteractive -NoProfile -Command \"Write-Output $PSVersionTable.PSVersion\"";
-            commandLineOutput = $"{powerShellExe} {arguments}";
+            var commandLineOutput = $"{powerShellExe} {arguments}";
 
             // Despite our old check conforming to Microsoft's recommendations
             // for PS version checking around the 1.0/2.0 era, and extending
             // to detect 3.0, it failed to detect 4. Going the direct route:
             try
             {
-                // We're in the WPF installer prerequisite check. IPrerequisite.Check() must return
-                // synchronously, so we block on the async call with .GetAwaiter().GetResult().
-                // This is sync-over-async but is safe because the installer dispatches us on a
-                // plain thread-pool worker. No captured SynchronizationContext, so no deadlock.
-                // See https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html
-                SilentProcessRunnerExtended.ExecuteCommandAsync(
+                await SilentProcessRunnerExtended.ExecuteCommandAsync(
                     powerShellExe,
                     arguments,
                     ".",
                     stdOut.WriteLine,
                     s => stdErr.WriteLine($"ERR: {s}"),
-                    cancel: CancellationToken.None).GetAwaiter().GetResult();
+                    cancel: CancellationToken.None);
 
                 var outputText = stdOut.ToString();
                 new SystemLog().Verbose("PowerShell prerequisite check output: " + outputText);
 
                 commandLineOutput = $"{commandLineOutput}{Environment.NewLine}{stdOut}{Environment.NewLine}{stdErr}";
 
-                return outputText.Contains("Major");
+                return (outputText.Contains("Major"), commandLineOutput);
             }
             catch (Exception ex)
             {
                 commandLineOutput = $"{commandLineOutput}{Environment.NewLine}{ex}";
-                return false;
+                return (false, commandLineOutput);
             }
         }
     }
--- a/source/Octopus.Manager.Tentacle/PreReq/PowerShellPrerequisite.cs
+++ b/source/Octopus.Manager.Tentacle/PreReq/PowerShellPrerequisite.cs
@@ -11,10 +11,16 @@ namespace Octopus.Manager.Tentacle.PreReq
     {
         public string StatusMessage => "Checking that Windows PowerShell 2.0 is installed...";
 
-        // We're at the WPF installer prerequisite boundary. IPrerequisite.Check() must return
-        // synchronously, so this is the sync-over-async bridge: a one-line wrapper over the
-        // private async implementation. Safe because the installer dispatches us on a plain
-        // thread-pool worker. No captured SynchronizationContext, so no deadlock.
+        // Why this is sync: IPrerequisite.Check() is part of a sync interface used by
+        // the WPF installer's prerequisite plumbing. Making it async would mean
+        // converting the whole IPrerequisite chain, which is a wider refactor than
+        // this PR.
+        //
+        // Why blocking on the async call is safe: PreReqWindow.Start dispatches each
+        // prerequisite via DispatchHelper.Background, which queues us via
+        // ThreadPool.QueueUserWorkItem. That's a plain thread-pool worker with no
+        // SynchronizationContext, so there's nothing for the awaited continuation
+        // to wait on.
         // See https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html
         public PrerequisiteCheckResult Check()
             => CheckAsync().GetAwaiter().GetResult();
--- a/source/Octopus.Manager.Tentacle/PreReq/PowerShellPrerequisite.cs
+++ b/source/Octopus.Manager.Tentacle/PreReq/PowerShellPrerequisite.cs
@@ -1,5 +1,6 @@
 using System;
 using System.IO;
+using System.Threading;
 using Octopus.Tentacle.Diagnostics;
 using Octopus.Tentacle.Util;
 
@@ -33,12 +34,21 @@ namespace Octopus.Manager.Tentacle.PreReq
             // to detect 3.0, it failed to detect 4. Going the direct route:
             try
             {
-                SilentProcessRunnerExtended.ExecuteCommand(
+                // We're in the WPF installer prerequisite check. IPrerequisite.Check() must return
+                // synchronously — there's no async version of the interface — so we block on the async
+                // call with .GetAwaiter().GetResult().
+                // This is safe because we're on a plain thread-pool worker. The risk with blocking on
+                // async is a deadlock: if the async work needs to resume on the same thread that's
+                // blocked waiting for it, neither can make progress. Thread-pool workers don't have
+                // that constraint — when the async work finishes it can pick up on any free thread,
+                // not specifically this one, so the block resolves normally.
+                SilentProcessRunnerExtended.ExecuteCommandAsync(
                     powerShellExe,
                     arguments,
                     ".",
                     stdOut.WriteLine,
-                    s => stdErr.WriteLine($"ERR: {s}"));
+                    s => stdErr.WriteLine($"ERR: {s}"),
+                    cancel: CancellationToken.None).GetAwaiter().GetResult();
 
                 var outputText = stdOut.ToString();
                 new SystemLog().Verbose("PowerShell prerequisite check output: " + outputText);
--- a/source/Octopus.Manager.Tentacle/PreReq/PowerShellPrerequisite.cs
+++ b/source/Octopus.Manager.Tentacle/PreReq/PowerShellPrerequisite.cs
@@ -35,13 +35,10 @@ namespace Octopus.Manager.Tentacle.PreReq
             try
             {
                 // We're in the WPF installer prerequisite check. IPrerequisite.Check() must return
-                // synchronously — there's no async version of the interface — so we block on the async
-                // call with .GetAwaiter().GetResult().
-                // This is safe because we're on a plain thread-pool worker. The risk with blocking on
-                // async is a deadlock: if the async work needs to resume on the same thread that's
-                // blocked waiting for it, neither can make progress. Thread-pool workers don't have
-                // that constraint — when the async work finishes it can pick up on any free thread,
-                // not specifically this one, so the block resolves normally.
+                // synchronously, so we block on the async call with .GetAwaiter().GetResult().
+                // This is sync-over-async but is safe because the installer dispatches us on a
+                // plain thread-pool worker. No captured SynchronizationContext, so no deadlock.
+                // See https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html
                 SilentProcessRunnerExtended.ExecuteCommandAsync(
                     powerShellExe,
                     arguments,
--- a/source/Octopus.Tentacle.Core/Util/CommandLine/SilentProcessRunner.cs
+++ b/source/Octopus.Tentacle.Core/Util/CommandLine/SilentProcessRunner.cs
@@ -27,7 +27,7 @@ namespace Octopus.Tentacle.Util
             return ExecuteCommandAsync(executable, arguments, workingDirectory, debug, info, error, customEnvironmentVariables: null, cancel: cancel);
         }
 
-        public static async Task<int> ExecuteCommandAsync(
+        public static Task<int> ExecuteCommandAsync(
             string executable,
             string arguments,
             string workingDirectory,
@@ -126,7 +126,6 @@ namespace Octopus.Tentacle.Util
                         WriteData(error, errorResetEvent, e);
                     };
 
-                    process.EnableRaisingEvents = true;
                     process.Start();
 
                     var running = true;
@@ -142,11 +141,7 @@ namespace Octopus.Tentacle.Util
                         process.BeginOutputReadLine();
                         process.BeginErrorReadLine();
 
-#if NETFRAMEWORK
-                        await WaitForExitAsyncNetFramework(process, cancel).ConfigureAwait(false);
-#else
-                        await process.WaitForExitAsync(cancel).ConfigureAwait(false);
-#endif
+                        process.WaitForExit();
 
                         SafelyCancelRead(process.CancelErrorRead, debug);
                         SafelyCancelRead(process.CancelOutputRead, debug);
@@ -158,7 +153,7 @@ namespace Octopus.Tentacle.Util
                         debug($"Process {exeFileNameOrFullPath} in {workingDirectory} exited with code {exitCode}");
 
                         running = false;
-                        return exitCode;
+                        return Task.FromResult(exitCode);
                     }
                 }
             }
@@ -208,30 +203,6 @@ namespace Octopus.Tentacle.Util
                 debug($"Swallowing {ex.GetType().Name} calling {action.Method.Name}.");
             }
         }
-
-#if NETFRAMEWORK
-        static Task WaitForExitAsyncNetFramework(Process process, CancellationToken cancellationToken)
-        {
-            var tcs = new TaskCompletionSource<object?>(TaskCreationOptions.RunContinuationsAsynchronously);
-            CancellationTokenRegistration registration = default;
-            void OnExited(object? sender, EventArgs e)
-            {
-                registration.Dispose();
-                tcs.TrySetResult(null);
-            }
-            process.Exited += OnExited;
-            if (process.HasExited) tcs.TrySetResult(null);
-            if (cancellationToken.CanBeCanceled)
-            {
-                registration = cancellationToken.Register(() =>
-                {
-                    process.Exited -= OnExited;
-                    tcs.TrySetCanceled(cancellationToken);
-                });
-            }
-            return tcs.Task;
-        }
-#endif
 
         static void DoOurBestToCleanUp(Process process, Action<string> error)
         {
--- a/source/Octopus.Tentacle.Core/Util/CommandLine/SilentProcessRunner.cs
+++ b/source/Octopus.Tentacle.Core/Util/CommandLine/SilentProcessRunner.cs
@@ -8,13 +8,14 @@ using System.Runtime.InteropServices;
 using System.Runtime.Versioning;
 using System.Text;
 using System.Threading;
+using System.Threading.Tasks;
 using Octopus.Tentacle.Core.Diagnostics;
 
 namespace Octopus.Tentacle.Util
 {
     public static class SilentProcessRunner
     {
-        public static int ExecuteCommand(
+        public static Task<int> ExecuteCommandAsync(
             string executable,
             string arguments,
             string workingDirectory,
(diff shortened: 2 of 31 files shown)


source/Octopus.Tentacle.Core/Util/CommandLine/SilentProcessRunner.cs
```diff
@@ -140,7 +142,11 @@ void WriteData(Action<string> action, ManualResetEventSlim resetEvent, DataRecei
                         process.BeginOutputReadLine();
                         process.BeginErrorReadLine();
 
-                        process.WaitForExit();
+#if NETFRAMEWORK
```
jimmyp: Remove all changes from this method implementation, this should be a signature change only in this class

source/Octopus.Tentacle.Core/Util/CommandLine/SilentProcessRunner.cs
```diff
@@ -203,6 +209,30 @@ static void SafelyCancelRead(Action action, Action<string> debug)
             }
         }
 
+#if NETFRAMEWORK
```
jimmyp: Remove this and put it in the next PR

source/Octopus.Tentacle.Kubernetes.Tests.Integration/Setup/KubernetesAgentInstaller.cs
```diff
@@ -219,14 +219,16 @@ public void Dispose()
                 NamespaceFlag,
                 AgentName);
 
-            var exitCode = SilentProcessRunner.ExecuteCommand(
+            // Dispose() cannot be made async; .GetAwaiter().GetResult() is safe here
+            // because this runs in test teardown (not inside an async context with a sync-blocking SynchronizationContext).
```
jimmyp: This doesn't match our lighthouse example comment on the wpf installer. Specifically where it calls out the pattern and links the blog post

source/Octopus.Tentacle.Kubernetes.Tests.Integration/Setup/KubernetesClusterInstaller.cs
```diff
@@ -174,15 +174,17 @@ string BuildNfsCsiDriverInstallArguments()
 
     public void Dispose()
     {
-        var exitCode = SilentProcessRunner.ExecuteCommand(
+        // Dispose() cannot be made async; .GetAwaiter().GetResult() is safe here
```
jimmyp: This also doesn't follow our lighthouse comment example

source/Octopus.Tentacle.Tests.Integration/Startup/LinuxConfigureServiceHelperFixture.cs
```diff
@@ -47,7 +47,8 @@ public void CannotWriteToServiceFileAsUser()
             WriteUnixFile(scriptPath);
 
             var chmodCmd = new CommandLineInvocation("/bin/bash", $"-c \"chmod 777 {scriptPath}\"");
-            chmodCmd.ExecuteCommand();
+            // Safe: sync test helper, no synchronisation context.
```
jimmyp: Does not follow our lighthouse comment example

jimmyp: Also there are many instances of this in this class. Can we create a helper method to dry it up or push this sync to async transition higher?

source/Octopus.Tentacle.Tests.Integration/Support/TentacleFetchers/LinuxTentacleFetcher.cs
```diff
@@ -62,14 +62,16 @@ public static void ExtractTarGzip(string gzArchiveName, string destFolder, ILogg
...
+            var exitCode = SilentProcessRunner.ExecuteCommandAsync(
                 "tar",
                 $"xzvf \"{gzArchiveName}\" -C \"{destFolder}\"",
                 tmp.DirectoryPath,
                 log,
                 log,
                 log,
-                CancellationToken.None);
+                cancel: CancellationToken.None)
+                // Safe: static void helper, no synchronisation context.
```
jimmyp: Does not follow our lighthouse comment example

source/Octopus.Tentacle.Tests.Integration/Util/LinuxTestUserPrincipal.cs
```diff
@@ -22,7 +22,8 @@ public LinuxTestUserPrincipal(string username)
         static void RunCommand(string arguments, bool failOnNonZeroExitCode = true)
         {
             var commandLineInvocation = new CommandLineInvocation("/bin/bash", arguments);
-            var result = commandLineInvocation.ExecuteCommand();
+            // Safe: constructor-time helper, no synchronisation context.
```
jimmyp: Does not follow our lighthouse comment example

source/Octopus.Tentacle.Tests.Integration/Util/SilentProcessRunnerFixture.cs
```diff
@@ -406,6 +406,8 @@ static string EchoEnvironmentVariable(string varName)
             return $"${varName}";
         }
 
+        // Sync-over-async is safe here: NUnit runs tests on a plain ThreadPool thread with no
+        // synchronisation context, so there is no risk of deadlock.
```
jimmyp: Does not follow our lighthouse comment example and is not at the actual call site

source/Octopus.Tentacle/Startup/LinuxServiceConfigurator.cs
```diff
@@ -194,7 +194,13 @@ void WriteUnitFile(string path, string contents)
             File.WriteAllText(path, contents);
 
             var commandLineInvocation = new CommandLineInvocation("/bin/bash", $"-c \"chmod 644 {path}\"");
-            var result = commandLineInvocation.ExecuteCommand();
+            // We're in WriteUnitFile, called from IServiceConfigurator.ConfigureService implementations
+            // which are sync (called from the Tentacle service-management CLI), so we block on the
+            // async call with .GetAwaiter().GetResult().
```
jimmyp: ITs not clear why these things have to be sync?

source/Octopus.Tentacle/Startup/LinuxServiceConfigurator.cs
```diff
@@ -194,7 +194,13 @@ void WriteUnitFile(string path, string contents)
...
+            // This is sync-over-async but is safe because the CLI dispatches us on a plain
```
jimmyp: How do we know the cli dispatches us on a thread pool worker?

source/Octopus.Tentacle/Startup/LinuxServiceConfigurator.cs
```diff
@@ -219,14 +225,17 @@ void CheckSystemPrerequisites()
         bool IsSystemdInstalled()
         {
             var commandLineInvocation = new CommandLineInvocation("/bin/bash", "-c \"command -v systemctl >/dev/null\"");
-            var result = commandLineInvocation.ExecuteCommand();
+            // Same sync-over-async boundary as WriteUnitFile: CheckSystemPrerequisites is called
+            // from the sync ConfigureService path, on a plain thread-pool worker.
```
jimmyp: Does not follow our lighthouse comment example

source/Octopus.Tentacle/Startup/LinuxServiceConfigurator.cs
```diff
@@ -219,14 +225,17 @@ void CheckSystemPrerequisites()
...
+            var result = commandLineInvocation.ExecuteCommandAsync().GetAwaiter().GetResult();
             return result.ExitCode == 0;
         }
 
         bool HaveSudoPrivileges()
         {
             var commandLineInvocation = new CommandLineInvocation("/bin/bash", "-c \"sudo -vn 2> /dev/null\"");
-            var result = commandLineInvocation.ExecuteCommand();
+            // Same sync-over-async boundary as IsSystemdInstalled.
```
jimmyp: Does not follow our lighthouse comment example

source/Octopus.Manager.Tentacle/PreReq/PowerShellPrerequisite.cs
```diff
@@ -33,12 +34,18 @@ static bool CheckPowerShellIsInstalled(out string commandLineOutput)
             // to detect 3.0, it failed to detect 4. Going the direct route:
             try
             {
-                SilentProcessRunnerExtended.ExecuteCommand(
+                // We're in the WPF installer prerequisite check. IPrerequisite.Check() must return
+                // synchronously, so we block on the async call with .GetAwaiter().GetResult().
+                // This is sync-over-async but is safe because the installer dispatches us on a
+                // plain thread-pool worker. No captured SynchronizationContext, so no deadlock.
+                // See https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html
```
jimmyp: @LukeButters do you have enough knowledge of this repo to verify Claude claims on each of these? Or shall I organise us a sync to go through them one by one

source/Octopus.Tentacle/Kubernetes/KubernetesDirectoryInformationProvider.cs
```diff
@@ -55,7 +55,14 @@ public KubernetesDirectoryInformationProvider(ISystemLog log, ISilentProcessRunn
...
             var stdOut = new List<string>();
             var stdErr = new List<string>();
-            var exitCode = silentProcessRunner.ExecuteCommand("du", $"-s -B 1 {directoryPath}", "/", stdOut.Add, stdErr.Add);
+            // We're in the IMemoryCache.GetOrCreate factory that populates the disk-space cache entry.
+            // The cache factory delegate is synchronous (Func<ICacheEntry, T>), so we block on the
+            // async call with .GetAwaiter().GetResult().
+            // This is sync-over-async but is safe because the cache factory runs on a plain
+            // thread-pool worker. No captured SynchronizationContext, so no deadlock.
+            // See https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html
+            var exitCode = silentProcessRunner.ExecuteCommandAsync("du", $"-s -B 1 {directoryPath}", "/", stdOut.Add, stdErr.Add)
```
LukeButters: @jimmyp Might want check with modern deployments on this one or push the async.

source/Octopus.Manager.Tentacle/PreReq/PowerShellPrerequisite.cs
LukeButters: Maybe @sburmanoctopus knows about this, he has done UI programming.

sburmanoctopus: From my WPF days, blocking would cause the UI to freeze, which wasn't great.

When async/await came along, it was amazing, because you could do async stuff as long as you added `.ConfigureAwait(true)` to the end. This would ensure the logic returned to the calling thread, which was the UI thread. This freed up the UI thread for other things while async code was running, and didn't block the UI thread anymore!

So by default, I would suggest using `.ConfigureAwait(true)`. 

However, in saying that, I have no context on where this code is being called, how it's being called etc. It's possible something else is dealing with the thread locking issue, or this is being called from a place where we must block the UI thread.

source/Octopus.Tentacle/Kubernetes/KubernetesDirectoryInformationProvider.cs
jimmyp: Took the second option @LukeButters: pushed the async sibling up so the in-process async caller awaits directly.

The two consumers of this method had different async-ness:

- **`KubernetesScriptPodCreator.CreateScriptContainer`** is `async` end-to-end. It used to call sync `GetStorageInformation()` → sync `GetPathUsedBytes()` → block on async. That hop is gone — `CreateScriptContainer` now `await`s a new `GetStorageInformationAsync()` / `GetPathUsedBytesAsync()` pair.
- **`KubernetesPhysicalFileSystem.EnsureDiskHasEnoughFreeSpace`** is the `IOctopusFileSystem` override and is sync because the interface is sync. The sync `GetPathUsedBytes` still exists for this one consumer.

So we still have one sync-over-async on the file-system override path. Hoping you can do a sanity check on it @liam-mackie since you wrote the original — the relevant safety claim is:

> The Kubernetes agent is a console process (`static int Main`), so .NET doesn't install a default `SynchronizationContext`. I've also confirmed by binary inspection that **Halibut 8.1.1633** (the version Tentacle uses) contains no references to `SynchronizationContext` or `SetSynchronizationContext` — its request-handler threads don't install one either. As long as nothing in the K8s-agent host startup (pod watcher, bootstrap wrappers, anything you'd recognise) installs one, the remaining `.GetAwaiter().GetResult()` on the `EnsureDiskHasEnoughFreeSpace` path is deadlock-safe.

An example of what _would_ make it unsafe (so you know what to look for): something like `SynchronizationContext.SetSynchronizationContext(new SomeCustomContext())` early in `Main`, or a UI-style context being installed by a library before the file-system code runs. Anything custom there would mean `await` inside `GetPathUsedBytesAsync` could try to resume on a thread that's blocked on `.GetResult()` — classic deadlock.

Does anything in the agent startup do that? If not, I'll leave the sync wrapper as-is.

jimmyp: Sorry @liam-mackie, Claude got enthusiastic and posted this before I was really ready to reach out to you. Are you the right person to talk to? If so great, but let me shape this up a bit better before bothering to engage with it

liam-mackie: I've got context here, and I'm probably the best person in this particular instance, but it's likely worth reaching out to the k8s requests channel for more information.

source/Octopus.Manager.Tentacle/PreReq/PowerShellPrerequisite.cs
jimmyp: Its being called from `Winow.Start` via `DispatchHelper.Background` [here](https://github.com/OctopusDeploy/OctopusTentacle/blob/91c254c46ffe4d3e1058675cf6070e5b89cce8b1/source/Octopus.Manager.Tentacle/PreReq/PreReqWindow.cs#L44) which queues it via `Threadpool.QueueUserWorkItem` which I think means that it is not on the UI thread and because `Threadpool.QueueUserWorkItem` isn't adding a synchronisation context we aren't suceptible to blocking?

I'd love it if someone could double check my thinking here. Claude can you give your 2c as well?

source/Octopus.Manager.Tentacle/PreReq/PowerShellPrerequisite.cs
```diff
@@ -9,48 +11,58 @@ public class PowerShellPrerequisite : IPrerequisite
     {
         public string StatusMessage => "Checking that Windows PowerShell 2.0 is installed...";
 
+        // We're at the WPF installer prerequisite boundary. IPrerequisite.Check() must return
+        // synchronously, so this is the sync-over-async bridge: a one-line wrapper over the
+        // private async implementation. Safe because the installer dispatches us on a plain
+        // thread-pool worker. No captured SynchronizationContext, so no deadlock.
+        // See https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html
```
jimmyp, line 24: Lets update this with a bit more commentary from the comment on the outdated comment on this seciton

source/Octopus.Tentacle.Kubernetes.Tests.Integration/Setup/KubernetesAgentInstaller.cs
```diff
@@ -219,14 +219,19 @@ public void Dispose()
                 NamespaceFlag,
                 AgentName);
 
-            var exitCode = SilentProcessRunner.ExecuteCommand(
+            // We're in IDisposable.Dispose(). Dispose() must return synchronously, so we
+            // block on the async call with .GetAwaiter().GetResult(). This is sync-over-async
+            // but is safe because the NUnit test runner dispatches us on a worker thread
+            // without a captured SynchronizationContext, so no deadlock.
+            // See https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html
```
jimmyp, line 229: And also because this is a test the impact of getting this wrong is low. Add this to all of our test justifications

source/Octopus.Tentacle.Tests.Integration/Support/TentacleFetchers/LinuxTentacleFetcher.cs
```diff
@@ -62,14 +62,20 @@ public static void ExtractTarGzip(string gzArchiveName, string destFolder, ILogg
             using var tmp = new TemporaryDirectory();
 
             Action<string> log = s => logger.Information(s);
-            var exitCode = SilentProcessRunner.ExecuteCommand(
+            // We're in a synchronous public static helper (ExtractTarGzip). The method
```
jimmyp: We need to talk about where this is called from and why this method needs to be sync

source/Octopus.Tentacle.Tests.Integration/Util/SilentProcessRunnerFixture.cs
```diff
@@ -418,7 +418,15 @@ static int Execute(
             var debug = new StringBuilder();
             var info = new StringBuilder();
             var error = new StringBuilder();
-            var exitCode = SilentProcessRunner.ExecuteCommand(
+
+            // We're in a synchronous test helper (Execute) that exposes a sync int
+            // return and out parameters. The method must return synchronously, so we
```
jimmyp: Why must this method return synchronously?

source/Octopus.Tentacle.Tests/Util/LinuxTestUserPrincipal.cs
```diff
@@ -22,7 +22,7 @@ public LinuxTestUserPrincipal(string username)
         static void RunCommand(string arguments, bool failOnNonZeroExitCode = true)
         {
             var commandLineInvocation = new CommandLineInvocation("/bin/bash", arguments);
-            var result = commandLineInvocation.ExecuteCommand();
+            var result = commandLineInvocation.ExecuteCommandAsync().GetAwaiter().GetResult();
```
jimmyp, line 34: No justification here

source/Octopus.Tentacle/Kubernetes/KubernetesDirectoryInformationProvider.cs
```diff
@@ -36,26 +38,36 @@ public KubernetesDirectoryInformationProvider(ISystemLog log, ISilentProcessRunn
             this.directoryInformationCache = directoryInformationCache;
         }
 
+        // Sync-over-async bridge for the one remaining sync caller: KubernetesPhysicalFileSystem
```
jimmyp: This requires context of our conversation to understand. Make this match the same format as the other justifications and make it clear and succinct

source/Octopus.Tentacle/Startup/LinuxServiceConfigurator.cs
```diff
@@ -47,22 +48,35 @@ public void ConfigureServiceByConfigPath(string thisServiceName,
...
+        // We're at the IServiceConfigurator boundary. IServiceConfigurator is consumed by
+        // ServiceCommand (an AbstractCommand), and AbstractCommand.Start() is sync because
+        // ICommand.Start() is sync (Topshelf's runtime callback API is sync). So
+        // ConfigureService must return synchronously. This is the single sync-over-async
+        // bridge for the Linux service-configuration code path: a one-line wrapper over the
+        // private async implementation. Safe because no SynchronizationContext is captured on
+        // this call stack: the console-app main thread has none by default, and Topshelf's
+        // OnStart callback runs on a `new Thread(...)` worker that also has none. Either way,
+        // no captured context means no deadlock.
+        // See https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html
```
jimmyp, line 63: This is close but I dont think people will know what the "Linux service-configuration code path" is describe what it is used for not what it is. Is bridge the right term too? Seems jargony and hard to understand

jimmyp: Also add the context that we use topshelf when we run abstract commands from the windows host and this also implements abstract command

source/Octopus.Tentacle/Startup/WindowsServiceConfigurator.cs
```diff
@@ -58,12 +59,30 @@ public void ConfigureServiceByConfigPath(string thisServiceName,
                 serviceConfigurationState);
         }
 
+        // We're at the IServiceConfigurator boundary. IServiceConfigurator is consumed by
```
jimmyp: Apply the same feedback here as we gave on the linux service configurator

source/Octopus.Manager.Tentacle/PreReq/PowerShellPrerequisite.cs
sburmanoctopus: Ah yes. 

Yeah, the `Start` method is being called by the `Loaded` event, which is on the UI thread. But like you said, since the `Start` method calls `ThreadPool.QueueUserWorkItem`, any work done there is on a thread from the thread pool (so not the UI thread).

In which case, all we would be doing here is blocking the thread from the thread pool. If there aren't many processes being run on the thread pool, I don't think this is an issue. Especially for a manager UI.

It would be nice to see if we could make that whole call chain async, but I'm not sure that's worth the effort.

jimmyp: Yeah I tried. It runs into an issue when another host doesn't have an async main entry point. I could push it to there. But I think I'm happy with where I've gotten to so far

source/Octopus.Tentacle/Kubernetes/KubernetesPhysicalFileSystem.cs
```diff
@@ -54,5 +55,17 @@ public override void EnsureDiskHasEnoughFreeSpace(string directoryPath, long req
 
             return null;
         }
+
+        public async Task<(ulong freeSpaceBytes, ulong totalSpaceBytes)?> GetStorageInformationAsync()
+        {
+            var bytesUsed = await directoryInformationProvider.GetPathUsedBytesAsync(HomeDir);
+            var bytesTotal = directoryInformationProvider.GetPathTotalBytes();
```
jimmyp, line 65: Lets pull this out as a shared sync method with a comment that this is to preserve dry and the async call chain we need

source/Octopus.Tentacle/Kubernetes/KubernetesDirectoryInformationProvider.cs
```diff
@@ -36,26 +38,39 @@ public KubernetesDirectoryInformationProvider(ISystemLog log, ISilentProcessRunn
...
+        // Why this is sync: the only caller is EnsureDiskHasEnoughFreeSpace, which
+        // overrides a sync method on the IOctopusFileSystem chain. Making that
+        // whole chain async is a wider refactor than this PR. New code should
+        // call GetPathUsedBytesAsync directly instead of going through here.
+        //
+        // Why blocking on the async call is safe: .GetAwaiter().GetResult() can
+        // deadlock when the calling thread has a SynchronizationContext. The
+        // Kubernetes agent is a console app and doesn't set one up, so there's
+        // nothing for the awaited continuation to wait on.
+        // See https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html
```
jimmyp, line 50: @liam-mackie I think this is the only place I need your eyes.

This change is because we are making something async lower down the call stack to help make it easier to cancel/abandon when a script it not behaving. Most tentacle code is sync so we end up blocking on async code which can cause issues. I think the comment above justifies that this instance below wont cause any issues, but let me know if you think the rationale misses something?

liam-mackie: That sounds about right to me - I don't think the agent has anything that would add a `SynchronisationContext` - @APErebus just double checking you don't have anything to add here, since you have a bunch of experience here

LukeButters (approved):

PR #1239, opened 2026-05-27
remove `claude.yml`, as we're not currently using the in-GitHub Claude Code PR review feature

todthomson:
We've disabled the Claude Code code reviews feature (for now) as there is a significant expense involved in each review.

This `claude.yml` GitHub Actions workflow isn't required in the absence of that feature, so we're removing it for now.


OctoJames (approved):
Looks good to me

PR #1240, opened 2026-05-27
feat: add liveness probe support for polling Tentacles

ATGardner:
# Background

Polling Kubernetes Tentacles need a liveness probe so Kubernetes can detect and recover from cases where a pod remains `Running` but the Tentacle process is stuck (deadlocked, frozen by EDR contention, etc.). Readiness is already covered by an exec-based command, so this PR closes the remaining gap.

Polling Tentacles run with `NoListen=true` and therefore do not expose an HTTP endpoint, so a Helm-chart-only change isn't viable — the signal has to come from the Tentacle process itself.

# Results

Adds a `tentacle livez` exec command suitable for use as a Kubernetes `livenessProbe`.

Two pieces work together:

- **`LivenessHeartbeatTask`** — a `BackgroundTask` registered in `MaintenanceModule` that touches `<HomeDirectory>/livez.heartbeat` every 10 seconds while the agent is running. Failure to write is logged but does not crash the task — the heartbeat going stale is the signal.
- **`LivezCommand`** — `tentacle livez [--max-age=<seconds>]` (default 60s). Resolves the heartbeat path from the selected instance's `IHomeConfiguration`, then fails with a `ControlledFailureException` (non-zero exit) if the file is missing or its `LastWriteTimeUtc` is older than the threshold. Time comes from `IClock` so the behaviour is testable.

This pairs with the matching change in [OctopusDeploy/helm-charts#608](https://github.com/OctopusDeploy/helm-charts/pull/608), which wires the new command into the polling-Tentacle pod spec as a `livenessProbe`.

## Before

Polling Tentacles had no liveness signal. A stuck agent in a `Running` pod would stay stuck until something external (operator, manual `kubectl delete pod`) restarted it.

## After

```
$ tentacle livez --instance=Tentacle
# exit 0 while the heartbeat is fresh

$ tentacle livez --instance=Tentacle --max-age=30
Liveness heartbeat at /etc/octopus/.../livez.heartbeat is stale: last written 47s ago (threshold 30s).
# exit non-zero — kubelet restarts the pod
```

# How to review this PR

Quality :heavy_check_mark:

Two areas worth a careful look:

1. **Threshold choice.** Heartbeat ticks every 10s, default `--max-age` is 60s — a 6× margin against transient I/O blips. Open to feedback on whether that's the right ratio for the Helm chart's `periodSeconds` / `failureThreshold` defaults.
2. **Failure modes of the heartbeat task itself.** `IOException` and `UnauthorizedAccessException` on the touch are logged and swallowed so a brief filesystem hiccup doesn't take down the task entirely. Anything else bubbles up and ends the task — which is intentional, because a permanently broken heartbeat *should* trip the probe.

Test coverage in `LivezCommandFixture` and `LivenessHeartbeatTaskFixture` (fresh / stale / missing / invalid `--max-age` / heartbeat file gets written and refreshed).

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle/Commands/LivezCommand.cs
+++ b/source/Octopus.Tentacle/Commands/LivezCommand.cs
@@ -1,0 +1,61 @@
+using System;
+using System.IO;
+using Octopus.Tentacle.Configuration;
+using Octopus.Tentacle.Configuration.Instances;
+using Octopus.Tentacle.Core.Diagnostics;
+using Octopus.Tentacle.Diagnostics;
+using Octopus.Tentacle.Startup;
+using Octopus.Time;
+
+namespace Octopus.Tentacle.Commands
+{
+    public class LivezCommand : AbstractStandardCommand
+    {
+        const int DefaultMaxAgeSeconds = 60;
+
+        readonly Lazy<IHomeConfiguration> homeConfiguration;
+        readonly IClock clock;
+
+        int maxAgeSeconds = DefaultMaxAgeSeconds;
+
+        public LivezCommand(
+            Lazy<IHomeConfiguration> homeConfiguration,
+            IClock clock,
+            IApplicationInstanceSelector instanceSelector,
+            ISystemLog log,
+            ILogFileOnlyLogger logFileOnlyLogger)
+            : base(instanceSelector, log, logFileOnlyLogger)
+        {
+            this.homeConfiguration = homeConfiguration;
+            this.clock = clock;
+
+            Options.Add(
+                "max-age=",
+                $"Maximum allowed age of the liveness heartbeat in seconds (default {DefaultMaxAgeSeconds}).",
+                v => maxAgeSeconds = int.Parse(v));
+        }
+
+        protected override void Start()
+        {
+            base.Start();
+
+            if (maxAgeSeconds <= 0)
+                throw new ControlledFailureException("--max-age must be a positive number of seconds.");
+
+            var home = homeConfiguration.Value.HomeDirectory;
+            if (string.IsNullOrWhiteSpace(home))
+                throw new ControlledFailureException("Could not determine the Tentacle home directory for the selected instance.");
+
+            var heartbeatPath = Path.Combine(home!, LivenessHeartbeatTask.HeartbeatFileName);
+
+            if (!File.Exists(heartbeatPath))
+                throw new ControlledFailureException($"Liveness heartbeat file not found at {heartbeatPath}; the Tentacle agent may not be running.");
+
+            var lastWriteUtc = File.GetLastWriteTimeUtc(heartbeatPath);
+            var age = clock.GetUtcTime().DateTime - lastWriteUtc;
+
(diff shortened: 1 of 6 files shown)


gb-8 (approved):
EFT Code owner review: ✅ Looks good to me.

Worth getting some 👀 from Modern Deployments team as per [Luke's comment](https://octopusdeploy.slack.com/archives/C04GAH78491/p1779918674610279?thread_ts=1779881911.609279&cid=C04GAH78491).

source/Octopus.Tentacle/Commands/LivezCommand.cs
```diff
@@ -0,0 +1,61 @@
...
+            if (string.IsNullOrWhiteSpace(home))
+                throw new ControlledFailureException("Could not determine the Tentacle home directory for the selected instance.");
+
+            var heartbeatPath = Path.Combine(home!, LivenessHeartbeatTask.HeartbeatFileName);
+
+            if (!File.Exists(heartbeatPath))
+                throw new ControlledFailureException($"Liveness heartbeat file not found at {heartbeatPath}; the Tentacle agent may not be running.");
+
+            var lastWriteUtc = File.GetLastWriteTimeUtc(heartbeatPath);
+            var age = clock.GetUtcTime().DateTime - lastWriteUtc;
```
zentron, line 55: Minor, this _generally_ will work fine, but only due to convention the system clock returning the LocalDateTime in Offset=0.

```suggestion
            var age = clock.GetUtcTime().UtcDateTime - lastWriteUtc;
```

source/Octopus.Tentacle/Diagnostics/LivenessHeartbeatTask.cs
```diff
@@ -0,0 +1,82 @@
...
+            this.homeConfiguration = homeConfiguration;
+            this.tickInterval = tickInterval;
+        }
+
+        protected override async Task RunTask(CancellationToken cancellationToken)
+        {
+            var heartbeatPath = ResolveHeartbeatPath();
+            if (heartbeatPath is null)
+            {
+                Log.Warn("Liveness heartbeat task could not resolve a home directory; heartbeat file will not be written.");
```
zentron, line 40: I'd probably log this as an error, to really get attention at the startup (as it only runs this check once on startup the gets lost in a sea of logs) or conversely let it keep ticking and logging warnings, perhaps the path resolves properly later?

zentron (commented):
Few minor changes suggested

source/Octopus.Tentacle/Commands/LivezCommand.cs
ATGardner: fixed in https://github.com/OctopusDeploy/OctopusTentacle/pull/1242

source/Octopus.Tentacle/Diagnostics/LivenessHeartbeatTask.cs
ATGardner: fixed in https://github.com/OctopusDeploy/OctopusTentacle/pull/1242

PR #1242, opened 2026-05-28
fix: correct liveness probe time calculation and improve logging

ATGardner:
just fixed two comments i missed from https://github.com/OctopusDeploy/OctopusTentacle/pull/1240


gb-8 (approved):
EFT Code-owner review: ✅ Approved.

Changes match feedback on previous PR.

PR #1245, opened 2026-06-01
Run Windows-only tests on a GitHub Actions runner (on demand)

jimmyp:
## Background

Some of our integration tests are Windows-only (`[WindowsTest]`) which me and Claude obviously cant run my Mac. The only option currently is to push to a PR and let CI run, but thats a long feedback loop, this PR adds a new github actions based mechanism to shorten the loop for agents.

## Results

Adds `.github/workflows/windows-test.yml`: a `workflow_dispatch`-only job on `windows-latest` that takes a required `filter` and runs `dotnet test` against `Octopus.Tentacle.Tests.Integration` (`--framework net8.0 --filter <filter>`). No `push` or `pull_request` trigger, so it only runs when someone asks.

Also adds a `run-windows-tests` skill under `.claude/skills/` that dispatches the workflow and watches it to a pass/fail.

I've run it end to end: [`CancellationToken_WhenGrandchildHoldsRedirectedPipes_ShouldNotHang` passed on the runner in 543ms, about 5 minutes start to finish.](https://github.com/OctopusDeploy/OctopusTentacle/actions/runs/26786771691)

## How to review this PR

- 🤷 its a brave new agentic world?

## Reducing risk

This is dev tooling only, I cant see any real risk

_[JIM_BOT.EXE v2.13]_


.claude/skills/run-windows-tests/run.sh
```diff
@@ -0,0 +1,25 @@
...
+set -euo pipefail
+
+FILTER="${1:?usage: run.sh \"<filter>\"  e.g. Name~WhenGrandchildHoldsRedirectedPipes}"
+REF="$(git rev-parse --abbrev-ref HEAD)"
+
+echo "Dispatching windows-test.yml on $REF with filter: $FILTER"
+gh workflow run windows-test.yml -f filter="$FILTER" --ref "$REF"
+
+# Give GitHub a moment to register the run, then find and watch it.
+sleep 5
```
rhysparry: I'm curious whether this will be flaky in practice, but can be addressed if it proves to be an issue.

LukeButters (approved):

.claude/skills/run-windows-tests/SKILL.md
```diff
@@ -0,0 +1,53 @@
...
+## Overview
+
+Windows-only tests (`[WindowsTest]`) compile on macOS but **skip at runtime** — they need
+a real Windows host (they spawn `cmd.exe`, `ping.exe`, query WMI). This skill runs them on
+a **GitHub Actions `windows-latest`** runner and streams the result back. `github.com` is
+reachable from the sandbox, so I can trigger and watch the run to completion myself — no
+local VM, no handing it back.
+
+(A local QEMU VM was tried and abandoned: brew's edk2 firmware on Apple Silicon only
+enumerates USB storage and won't boot the Windows installer. See the project memory.)
```
rhysparry: Not sure how helpful this is to the agent. I don't think memory is shared in the repo, right?

.claude/skills/run-windows-tests/SKILL.md
```diff
@@ -0,0 +1,53 @@
...
+## When to use
+
+- A test is `[WindowsTest]` / shows as skipped locally and you need its real result.
+- You're iterating on Windows process/abandonment behavior in `SilentProcessRunner`.
+- NOT for tests that run on macOS — run those with `dotnet test` directly.
+
+## How to run
+
+```bash
+.claude/skills/run-windows-tests/run.sh "Name~WhenGrandchildHoldsRedirectedPipes"
```
rhysparry: Is it clear that this is an example?

.github/workflows/windows-test.yml
```diff
@@ -0,0 +1,32 @@
...
+    inputs:
+      filter:
+        description: "Test filter, e.g. Name~WhenGrandchildHoldsRedirectedPipes"
+        required: true
+
+jobs:
+  windows-test:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v4
```
rhysparry: current checkout version is 6.
```suggestion
      - uses: actions/checkout@v6
```

.github/workflows/windows-test.yml
```diff
@@ -0,0 +1,32 @@
...
+
+      - uses: actions/setup-dotnet@v4
```
rhysparry: Current setup-dotnet version is 5.

```suggestion
      - uses: actions/setup-dotnet@v5
```

rhysparry (approved):
Nothing blocking, but a few possible improvements. Action versions are probably worth fixing.

gb-8 (approved):
Claude found some things (see below).

I'm not sure how problematic the hanging test issue is. Is there a cost implication? Or a throughput implication?

Other than that, they don't seem crucial, so I'm ✅ and leaving it up to you.


--- 

PR Review: #1245 — Run Windows-only tests on a GitHub Actions runner (on demand)
                                                                                                                                                                                                                                                                                                                          
  Author: Jim Pelletier | Base: main | Files: 3 | +110 / -0                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                          
  ---                                                                                                                                                                                                                                                                                                                     
  Overview                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                          
  A neat piece of dev tooling that closes a real pain point: [WindowsTest] tests skip entirely on macOS/Apple Silicon, leaving a slow push-to-CI feedback loop as the only option. This PR adds a manual workflow_dispatch GitHub Actions job and a Claude skill (run.sh + SKILL.md) to dispatch it and stream results.   
  The approach is well-scoped — dispatch-only, no accidental push/pull_request triggers, filter is mandatory.                                                                                                                                                                                                             
  
  ---                                                                                                                                                                                                                                                                                                                     
  What works well                                                   
                 
  - Security: passing the filter through env: rather than inlining ${{ inputs.filter }} into the run: block is exactly right — prevents script injection.
  - set -euo pipefail and ${1:?...} in run.sh are good defensive defaults.                                                                                                                                                                                                                                                
  - SKILL.md is unusually thorough — the "Common mistakes" section and the workflow_dispatch-on-default-branch gotcha are the kind of thing that would bite repeatedly without documentation.                                                                                                                             
  - actions/setup-dotnet@v4 with global-json-file: global.json correctly pins the SDK version.                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                          
  ---                                                                                                                                                                                                                                                                                                                     
  Issues and suggestions                                                                                                                                                                                                                                                                                                  
                                                                    
  1. Race condition in run.sh (medium risk)
```
                                                                                                                                                                                                                                                                                                                          
  sleep 5
  RID="$(gh run list --workflow windows-test.yml --branch "$REF" --limit 1 --json databaseId --jq '.[0].databaseId')"                                                                                                                                                                                                     
    
```                                                                                                                                                                                                                                                                                                                      
  sleep 5 then take the most-recent run ID is fragile in two ways:                                                                                                                                                                                                                                                        
  - GitHub is occasionally slow to register a dispatch → the listed run may be a previous one on the same branch.                                                                                                                                                                                                         
  - A parallel dispatch (another dev, a retry) → you watch the wrong run.                                                                                                                                                                                                                                                 
                                                                         
  gh workflow run can emit the new run URL with --json if you capture stderr, but the cleaner fix is to record the time before dispatch and filter by createdAt:                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                          
```
  BEFORE=$(date -u +%Y-%m-%dT%H:%M:%SZ)                                                                                                                                                                                                                                                                                   
  gh workflow run windows-test.yml -f filter="$FILTER" --ref "$REF"                                                                                                                                                                                                                                                       
  sleep 8                                                           
  RID="$(gh run list --workflow windows-test.yml --branch "$REF" --limit 5 \                                                                                                                                                                                                                                              
    --json databaseId,createdAt \                                   
    --jq "[.[] | select(.createdAt > \"$BEFORE\")] | .[0].databaseId")"                                                                                                                                                                                                                                                   

```                                                                    
  2. No timeout-minutes on the workflow job                                                                                                                                                                                                                                                                               
                                                                    
  A hanging test (or a test that unexpectedly runs more than the filtered set) will consume the GitHub Actions default 6-hour limit. Consider:                                                                                                                                                                            
                                                                    
```
  jobs:                                                                                                                                                                                                                                                                                                                   
    windows-test:                                                                                                                                                                                                                                                                                                         
      runs-on: windows-latest
      timeout-minutes: 30                                                                                                                                                                                                                                                                                                 

```                                                                    
  3. No permissions block

  The other workflow in this repo (approve-renovate-pull-request.yml) explicitly declares permissions. Even if only reads are needed, a restrictive permissions: block is good practice and communicates intent:                                                                                                          
  
```
  permissions:                                                                                                                                                                                                                                                                                                            
    contents: read                                                  

```                                                                                                                                                                                                                                                                                                                          
  4. --ref "$REF" silently fails if workflow isn't on main
                                                                                                                                                                                                                                                                                                                          
  SKILL.md documents this limitation well, but run.sh will fail with a confusing gh error if run before the PR is merged. A guard at the top of run.sh would make the failure message actionable:                                                                                                                         
  
```
  # Verify the workflow exists on the default branch                                                                                                                                                                                                                                                                      
  DEFAULT=$(gh repo view --json defaultBranchRef --jq '.defaultBranchRef.name')                                                                                                                                                                                                                                           
  if ! gh api "repos/{owner}/{repo}/contents/.github/workflows/windows-test.yml?ref=$DEFAULT" &>/dev/null; then
    echo "ERROR: windows-test.yml is not yet on '$DEFAULT' — merge this PR first." >&2                                                                                                                                                                                                                                    
    exit 1                                                                                                                                                                                                                                                                                                                
  fi                                                                                                                                                                                                                                                                                                                      

```                                                                                                                                                                                                                                                                                                                          
  ---                                                               
  Minor nits
            
  - run.sh line 22: # Give GitHub a moment to register the run, then find and watch it. — the comment implies this is expected to be reliable; worth softening ("may need a longer sleep on a busy runner").
  - The SKILL.md references "project memory" for the QEMU/edk2 story — fine for internal use, but that memory file isn't in this PR. Not a blocker.                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                          
  ---                                                                                                                                                                                                                                                                                                                     
  Summary                                                                                                                                                                                                                                                                                                                 
                                                                    
  Solid PR. The core design (dispatch-only, mandatory filter, env-based injection guard) is correct and the documentation is genuinely good. The main things worth fixing before merge are the 4 items above.

.claude/skills/run-windows-tests/run.sh
jimmyp: Yeah... claude is there a better way to do this than a sleep?

.claude/skills/run-windows-tests/SKILL.md
jimmyp: Claude remove this

.claude/skills/run-windows-tests/SKILL.md
jimmyp: Claude make it clear this is an example test name, you dont need to pass this specifically. Tell the user (which is you) how to construct this parameter

.claude/skills/run-windows-tests/run.sh
```diff
@@ -15,11 +15,29 @@ set -euo pipefail
 FILTER="${1:?usage: run.sh \"<filter>\"  e.g. Name~WhenGrandchildHoldsRedirectedPipes}"
 REF="$(git rev-parse --abbrev-ref HEAD)"
 
+# workflow_dispatch can only see the workflow once it's on the default branch. Fail with an
+# actionable message instead of a confusing gh error if it isn't there yet.
+DEFAULT="$(gh repo view --json defaultBranchRef --jq '.defaultBranchRef.name')"
+if ! gh api "repos/{owner}/{repo}/contents/.github/workflows/windows-test.yml?ref=$DEFAULT" >/dev/null 2>&1; then
+  echo "ERROR: windows-test.yml is not on the default branch ('$DEFAULT') yet, so workflow_dispatch can't see it. Merge this branch to '$DEFAULT' first." >&2
+  exit 1
+fi
```
jimmyp: Delete this

PR #1246, opened 2026-06-05
Remove Calamari volume mount subpath

APErebus:
# Background

The `subPath` is broken on AKS and it's really unnecessary tbh.

# Results

This means the entire calamari image is loaded at the `/calamari` mount path, howwever, there is only a single folder in each docker image, so it just means we end up with paths like `/calamari/Calamari/Calamari` 😂 

Related to MD-2016

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


flin-8 (approved):
LGTM

PR #1247, opened 2026-06-15
Remove max polling connection count limiter

liam-mackie:
To support self-hosted customers with large, multi-space agents, the maximum polling connection count was a hindrance rather than a good safety net. This PR removes it.

### DIFF
--- a/source/Octopus.Tentacle/Communications/HalibutInitializer.cs
+++ b/source/Octopus.Tentacle/Communications/HalibutInitializer.cs
@@ -115,8 +115,6 @@ namespace Octopus.Tentacle.Communications
             }
         }
                 
-        const int MaximumPollingConnectionCount = 8;
-                
         uint GetPollingConnectionCount()
         {
             //Open multiple polling connections if the env var is set to a non-zero/negative number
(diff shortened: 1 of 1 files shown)


source/Octopus.Tentacle/Communications/HalibutInitializer.cs
```diff
@@ -115,7 +115,8 @@ void AddPollingEndpoints()
             }
         }
                 
-        const int MaximumPollingConnectionCount = 8;
+        // We don't expect anyone to hit this, but the theoretical cap is the number of allocatable outbound ports
+        const int MaximumPollingConnectionCount = 65535;
```
zentron:
```suggestion
        const int MaximumPollingConnectionCount = 10000;
```

zentron: 10K TCP connections should be enough for anyone right :)

zentron (approved):
Just dialing the max down from the total saturation, while still being likely a ridiculous number

source/Octopus.Tentacle/Communications/HalibutInitializer.cs
liam-mackie: >10K TCP connections should be enough for anyone right :)

Famous last words, though I agree.

zentron (approved):

PR #1252, opened 2026-06-18
Pin K8s agent golang builder to 1.25.11

NickJosevski:
# Background

The Kubernetes agent's \`bootstrapRunner\` is compiled with the \`golang:1.25\` builder image, which floats to whatever the latest 1.25 patch is at build time. This pins it explicitly to **\`golang:1.25.11\`** (latest 1.25 patch, released 2026-06-02).

This is a manual stop-gap while the new K8s-agent Renovate config (#1250) can't yet raise PRs — its bot account (\`team-yosemite-bot\`) currently only has **read** access to this repo, so its pushes 403.

# Results

Pins the builder toolchain to a patched release, picking up the security fixes across 1.25.7–1.25.11, notably in **1.25.11**:
- **CVE-2026-42504** (\`mime\`) — CPU exhaustion from malicious MIME headers
- **CVE-2026-27145** (\`crypto/x509\`) — quadratic cert verification cost with large DNS SAN lists
- **CVE-2026-42507** (\`net/textproto\`) — error-message injection

Note: this is the **build-time** toolchain only (the image that compiles \`bootstrapRunner\`), not the shipped \`runtime-deps\` base image. \`go.mod\`'s \`go 1.25\` language-version directive is intentionally left unchanged.

## Before
\`\`\`dockerfile
FROM golang:1.25 AS bootstraprunnerbuilder
\`\`\`
## After
\`\`\`dockerfile
FROM golang:1.25.11 AS bootstraprunnerbuilder
\`\`\`

# How to review this PR

Quality :heavy_check_mark: — two-line change to the builder \`FROM\` tag in both \`Dockerfile\` and \`dev/Dockerfile\`. Green means go.

### DIFF
--- a/docker/kubernetes-agent-tentacle/Dockerfile
+++ b/docker/kubernetes-agent-tentacle/Dockerfile
@@ -1,7 +1,7 @@
 ARG RuntimeDepsTag
 
 
-FROM golang:1.25 AS bootstraprunnerbuilder
+FROM golang:1.25.11 AS bootstraprunnerbuilder
 
 ARG TARGETARCH
 ARG TARGETOS
--- a/docker/kubernetes-agent-tentacle/dev/Dockerfile
+++ b/docker/kubernetes-agent-tentacle/dev/Dockerfile
@@ -1,7 +1,7 @@
 ARG RuntimeDepsTag
 
 
-FROM golang:1.25 AS bootstraprunnerbuilder
+FROM golang:1.25.11 AS bootstraprunnerbuilder
 
 ARG TARGETARCH
 ARG TARGETOS


APErebus:
Is the idea that we now explicitly update the dependencies, rather than let it just be whatever is the latest at build time?

NickJosevski:
Ahh yes I see what you mean, I'm trying to work backwards from customers reporting CVEs on this, and depending on their tooling it may be picking up that we're not on the latest, and that there are CVEs (as listed) that are only patched in 1.25.11 or 1.26

the bot EMEA set up doesn't have access yet (still with SecOps) but it was going to propose 1.26 as the next bump

APErebus:
Given how relatively infrequently we update Tentacle, it probably makes sense to be explicit about the version so if we need to bump for a patch build, we can explicitly update it.

I'm fine for this change to go ahead

NickJosevski:
this isn't what the customer CVE scanners picked up (or shouldn't be), it was transitive deps here: https://github.com/OctopusDeploy/OctopusTentacle/pull/1253

closing

PR #1205, opened 2026-03-31
Fix CVE vulnerabilities in transitive NuGet dependencies (#1204)

pm7y:
**_Are you a customer of Octopus Deploy? Please contact [our support team](https://octopus.com/support) so we can triage your PR, so that we can make sure it's handled appropriately._**

# Background

#1204

The `octopusdeploy/tentacle` container image bundles transitive NuGet dependencies with known CVEs in `/opt/octopus/tentacle/Tentacle.deps.json`. These are flagged by container vulnerability scanners (Azure Defender for Cloud, Trivy, Snyk, etc.) and cannot be remediated by image consumers.

This PR pins 6 vulnerable transitive dependencies to their fixed versions by adding explicit `PackageReference` entries to `Octopus.Tentacle.csproj`.

# Results

| CVE | Severity | Package | Before | After |
|-----|----------|---------|--------|-------|
| CVE-2024-38095 | High | System.Formats.Asn1 | 6.0.0 | 6.0.1 |
| CVE-2023-29331 | High | System.Security.Cryptography.Pkcs | 6.0.1 | 6.0.3 |
| CVE-2017-11770 | High | System.Security.Cryptography.X509Certificates | 4.1.0 | 4.3.2 |
| CVE-2018-8292 | Medium | System.Net.Http | 4.1.0 | 4.3.4 |
| CVE-2019-0657, CVE-2019-0980, CVE-2019-0981 | Medium | System.Private.Uri | 4.3.0 | 4.3.2 |
| CVE-2019-0820 | Medium | System.Text.RegularExpressions | 4.3.0 | 4.3.1 |

## Before

Container vulnerability scanners flag 8 CVEs (3 High, 5 Medium) from `/opt/octopus/tentacle/Tentacle.deps.json`.

## After

All 8 CVEs resolved. No functional changes — only transitive dependency version pins.

# How to review this PR

This is a low-risk dependency pin change. The packages are already present as transitive dependencies; we are only overriding their versions to pick up security fixes.

Key points to verify:
- `System.Formats.Asn1` and `System.Security.Cryptography.Pkcs` are conditioned on `'$(TargetFrameworkIdentifier)' != '.NETFramework'` because they only target netstandard2.1/net6.0+ and would break the net48 build
- The remaining 4 packages support all target frameworks and are unconditional
- `System.Security.Cryptography.X509Certificates` is pinned to 4.3.2 (not the NVD-listed 4.1.2) because `System.Net.Http 4.3.4` depends on `>= 4.3.0`

[...]

PR #1255, opened 2026-06-18
Enable NuGet transitive dependency auditing (NuGetAuditMode=all)

NickJosevski:
# Background

While remediating the transitive `System.Security.Cryptography.Xml` 8.0.2 CVEs (found by an image scan, fixed in a separate PR), it became clear that **nothing in the build flags vulnerable transitive packages**. The .NET 8 SDK enables NuGet auditing by default, but only in `direct` mode — it audits direct `PackageReference`s and ignores the transitive graph, which is where these CVEs lived.

This PR turns on **`NuGetAuditMode=all`** so `dotnet restore` audits the full dependency graph and emits `NU1901`–`NU1904` warnings for known-vulnerable packages.

# Results

Added to `source/Octopus.Tentacle/Octopus.Tentacle.csproj`:
```xml
<PropertyGroup>
  <NuGetAudit>true</NuGetAudit>
  <NuGetAuditMode>all</NuGetAuditMode>
  <NuGetAuditLevel>low</NuGetAuditLevel>
  <WarningsNotAsErrors>$(WarningsNotAsErrors);NU1901;NU1902;NU1903;NU1904</WarningsNotAsErrors>
</PropertyGroup>
```

A fresh `dotnet restore` (net8.0) now reports — and notably surfaces **5 vulnerable transitive packages**, not just the one we went looking for:

| Package | Severity |
|---|---|
| System.Security.Cryptography.Xml 8.0.2 | high ×2 (the CVEs we're fixing) |
| System.Net.Http 4.1.0 | high |
| System.Private.Uri 4.3.0 | high ×2 + moderate |
| System.Security.Cryptography.X509Certificates 4.1.0 | high |
| System.Text.RegularExpressions 4.3.0 | high |

(The 4.x findings are exactly the packages PR #1205 set out to pin — this is independent confirmation of that work.)

## Important: this does NOT break the build
The project sets `TreatWarningsAsErrors`, so audit findings would otherwise become **errors** and fail CI. `WarningsNotAsErrors` keeps `NU190x` as **warnings** — informational visibility, not a hard gate. Verified: `dotnet restore` exits **0** with the warnings printed.

# How to review this PR

Quality :heavy_check_mark: — single `<PropertyGroup>`. To see it work: `dotnet restore source/Octopus.Tentacle/Octopus.Tentacle.csproj -p:TargetFramework=net8.0 --force` and watch the `NU1903` warnings. If you later want this to be a **gate**, drop the `WarningsNotAsErrors` line (or scope it) so findings fail the build.

### DIFF
--- a/source/Octopus.Tentacle/Octopus.Tentacle.csproj
+++ b/source/Octopus.Tentacle/Octopus.Tentacle.csproj
@@ -16,6 +16,18 @@
 		<Nullable>enable</Nullable>
 	</PropertyGroup>
 
+	<!-- NuGet security auditing. The .NET 8 SDK default (NuGetAuditMode=direct) only audits
+	     direct PackageReferences; 'all' also audits transitive dependencies, surfacing things
+	     like the transitive System.Security.Cryptography.Xml 8.0.2 CVEs at restore time.
+	     Audit findings (NU1901-NU1904) are kept as warnings so they don't trip the project's
+	     TreatWarningsAsErrors and break the build - this is informational, not a hard gate. -->
+	<PropertyGroup>
+		<NuGetAudit>true</NuGetAudit>
+		<NuGetAuditMode>all</NuGetAuditMode>
+		<NuGetAuditLevel>low</NuGetAuditLevel>
+		<WarningsNotAsErrors>$(WarningsNotAsErrors);NU1901;NU1902;NU1903;NU1904</WarningsNotAsErrors>
+	</PropertyGroup>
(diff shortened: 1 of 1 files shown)


todthomson (approved):
Approved, with thanks. 🥇

re: `WarningsNotAsErrors`
I am not opposed to you just failing the build for CVEs, thought that's just my opinion, but you could ask in EFT Requests in Slack if you agree. 😄

PR #1259, opened 2026-06-18
fix(docker): make Tentacle images OCI-compliant (FD-492)

todthomson:
**Linear:** [FD-492 — Image octopusdeploy/tentacle not OCI Compliant](https://linear.app/octopus/issue/FD-492/image-octopusdeploytentacle-not-oci-compliant)

## Background

Customer FD-492 (Zendesk 209584, ARUP Laboratories) reported that the `octopusdeploy/tentacle` Docker image is **not OCI compliant**: Podman rejects it when it is used as a base image (`FROM octopusdeploy/tentacle …`).

The root cause is **not** Dockerfile content and **not** the Debian 11→12 base-image upgrade (#1218 / EFT-3311) — that PR does not touch the manifest format, and building the old vs new image locally shows no difference. The actual problem is **inconsistent/mixed manifest media types** produced by the build/push pipeline. Docker tolerates this; strict OCI clients like Podman do not.

### What Octopus Server actually does (verified in the OctopusDeploy repo)

The triage note claimed Server "fixed the same issue in Jan 2026." I checked the Server repo history directly and that framing is **not accurate**. There is **no `oci-mediatypes`/`--provenance` flag combination anywhere in Server's history** (`git log -S` finds zero hits). Instead, Server builds its Linux image **single-architecture (`linux/amd64`) with the classic Docker builder** (`DockerTasks.DockerBuild` + `DockerTasks.DockerPush`, `nuke-build/Build.DockerImages.cs`), producing one self-consistent Docker schema2 manifest with no manifest list and no attestation index. Server avoids the problem by **never producing a multi-arch/attestation manifest**, rather than by fixing one with flags.

### Why this PR uses buildx flags instead of copying Server

The Tentacle **Kubernetes agent** image is genuinely **multi-arch** (`linux/arm64,linux/amd64`) and therefore *must* be built with buildx and a manifest list — it can't adopt Server's single-arch classic-build approach without dropping arm64. For a multi-arch buildx image, the correct equivalent remedy is to force a single OCI media type and drop the attestation index.

## Manifest evidence (verified against the published registry)

**`octopusdeploy/tentacle:latest` (the customer's image) — confirmed two distinct OCI violations:**

1. The OCI index mixes manifest formats:

   | Platform | Manifest mediaType |
   |---|---|
   | linux/amd64 | `application/vnd.oci.image.manifest.v1+json` (OCI — buildx) |
   | windows/amd64 | `application/vnd.docker.distribution.manifest.v2+json` (Docker schema2 — classic `docker build`/`push`) |

   A strict OCI consumer rejects an OCI index that references a Docker-schema2 manifest.

2. Even the linux/amd64 manifest mixes **layer** media types — `application/vnd.docker.image.rootfs.diff.tar.gzip` ×4 (inherited `debian` base layers) + `application/vnd.oci.image.layer.v1.tar+gzip` ×8 (new layers). This is the literal "mixes two layer format standards", normalized by `oci-mediatypes=true`.

**`kubernetes-agent-tentacle:latest` (this repo's image) — current published index contains 4 manifests:** arm64, amd64, **plus two `unknown/unknown` `attestation-manifest` entries** (buildx provenance). Its per-platform layers are already uniformly OCI, so here the defect is purely the attestation index — removed by disabling default attestations (`BUILDX_NO_DEFAULT_ATTESTATIONS=1`). (Per-platform layer types verified all-OCI.)

## Changes

1. **Build the Kubernetes agent image with consistent OCI media types** (`build/Build.Pack.cs`)
   - On push, use `--output type=image,oci-mediatypes=true,push=true` and `BUILDX_NO_DEFAULT_ATTESTATIONS=1` so the whole manifest uses a single OCI media type with no attestation index.
   - The local `--load` path (used by tests/dev) is unchanged.

2. **Modernize image labels** across all four Dockerfiles
   - Replace the deprecated `org.label-schema.*` namespace (retired 2016) with the standard `org.opencontainers.image.*` annotation keys.
   - Use the SPDX identifier `Apache-2.0` for the license.

## ⚠️ Follow-up required (out of scope for this repo)

The customer-reported image, **`octopusdeploy/tentacle`** (`docker/linux/Dockerfile`), is **amd64-only on Linux** but published in a multi-platform tag alongside the Windows image, and is built in the **TeamCity pipeline**, not in this repo's NUKE build. Closing FD-492 for that image requires, in TeamCity: (a) build/push the Linux image with `oci-mediatypes=true` so its layers are uniformly OCI, and (b) assemble the multi-platform tag so the Windows manifest is also OCI (otherwise the OCI index keeps referencing a Docker-schema2 manifest). This PR fixes the Kubernetes agent image (built here) and the shared label metadata.

## Testing notes

- `dotnet build build/_build.csproj` succeeds.
- Verify the pushed manifest with `docker buildx imagetools inspect --raw <tag>`: the index should contain only the two platform manifests (no `unknown/unknown` attestation manifests), and every layer `mediaType` should be `application/vnd.oci.image.layer.v1.tar+gzip`. Plus a `podman pull` / `FROM` smoke test.

---
**Note (review follow-up):** the attestation switch is standardised on `BUILDX_NO_DEFAULT_ATTESTATIONS=1` to match the `docker buildx bake` step in OctopusDeploy/TeamCity-Configuration#112 (identical across buildx `build` and `bake`, covers all default attestations, and avoids a post-PATH CLI flag).

---
**OCI annotations (review follow-up):** in addition to the Dockerfile `LABEL`s (which set `org.opencontainers.image.*` on the image *config*), the k8s agent build now also stamps those annotations onto the image **index** and each **platform manifest** via the buildx exporter `annotation-index.*` / `annotation.*` options — the spec-preferred location registries/OCI tooling read. Verified locally on a multi-arch build. The linux `octopusdeploy/tentacle` `:latest` index is assembled by a separate `docker manifest create` step (TeamCity-Configuration), so its index annotations are deferred to the manifest-step follow-up (switching to `docker buildx imagetools create --annotation`).

### DIFF
--- a/build/Build.Pack.cs
+++ b/build/Build.Pack.cs
@@ -769,9 +769,17 @@ partial class Build
                 };
 
                 // annotation-index.* lands on the image index, annotation.* on each platform manifest.
+                // Guard: annotation values are folded into the comma-separated buildx --output option
+                // list, so a comma in a value would corrupt it (the keys are fixed and comma-free).
                 var output = "type=image,oci-mediatypes=true,push=true";
                 foreach (var (key, value) in annotations)
+                {
+                    if (value.Contains(','))
+                        throw new InvalidOperationException(
+                            $"OCI annotation '{key}' value must not contain a comma; it would break the buildx --output option list: '{value}'");
+
                     output += $",annotation-index.{key}={value},annotation.{key}={value}";
+                }
 
                 settings = settings
                     .SetOutput(output)
--- a/build/Build.Pack.cs
+++ b/build/Build.Pack.cs
@@ -749,6 +749,8 @@ partial class Build
                 // onto both the image index and each platform manifest. The Dockerfile LABELs
                 // set the same metadata on the image *config*; annotations are the spec-preferred
                 // location that registries and OCI tooling read from the manifest/index.
+                // Keep these values in sync with the org.opencontainers.image.* LABELs in
+                // docker/kubernetes-agent-tentacle/Dockerfile.
                 var annotations = new Dictionary<string, string>
                 {
                     ["org.opencontainers.image.title"] = "Octopus Deploy Kubernetes Agent Tentacle",
--- a/build/Build.Pack.cs
+++ b/build/Build.Pack.cs
@@ -744,8 +744,30 @@ partial class Build
                 // is used (rather than --provenance=false) so the same switch applies to both
                 // `docker buildx build` here and `docker buildx bake` in the TeamCity Linux
                 // image build (see OctopusDeploy/TeamCity-Configuration).
+                //
+                // We also stamp the OCI image-spec annotations (org.opencontainers.image.*)
+                // onto both the image index and each platform manifest. The Dockerfile LABELs
+                // set the same metadata on the image *config*; annotations are the spec-preferred
+                // location that registries and OCI tooling read from the manifest/index.
+                var annotations = new Dictionary<string, string>
+                {
+                    ["org.opencontainers.image.title"] = "Octopus Deploy Kubernetes Agent Tentacle",
+                    ["org.opencontainers.image.vendor"] = "Octopus Deploy",
+                    ["org.opencontainers.image.url"] = "https://octopus.com",
+                    ["org.opencontainers.image.source"] = "https://github.com/OctopusDeploy/OctopusTentacle",
+                    ["org.opencontainers.image.licenses"] = "Apache-2.0",
+                    ["org.opencontainers.image.description"] = "Octopus Kubernetes Agent Tentacle instance with auto-registration to Octopus Server",
+                    ["org.opencontainers.image.version"] = FullSemVer,
+                    ["org.opencontainers.image.created"] = DateTime.UtcNow.ToString("O"),
+                };
+
+                // annotation-index.* lands on the image index, annotation.* on each platform manifest.
+                var output = "type=image,oci-mediatypes=true,push=true";
+                foreach (var (key, value) in annotations)
+                    output += $",annotation-index.{key}={value},annotation.{key}={value}";
+
                 settings = settings
-                    .SetOutput("type=image,oci-mediatypes=true,push=true")
+                    .SetOutput(output)
                     .AddProcessEnvironmentVariable("BUILDX_NO_DEFAULT_ATTESTATIONS", "1");
             }
             else
--- a/build/Build.Pack.cs
+++ b/build/Build.Pack.cs
@@ -737,13 +737,16 @@ partial class Build
             if (push)
             {
                 // FD-492: Force a single, consistent OCI media type across the whole image
-                // manifest and disable provenance attestations. Without this, buildx can emit
+                // manifest and disable default attestations. Without this, buildx can emit
                 // a manifest that mixes Docker and OCI layer media types (and an attestation
                 // index), which strict OCI clients such as Podman reject - in particular when
-                // the image is used as a base image (FROM ...).
+                // the image is used as a base image (FROM ...). BUILDX_NO_DEFAULT_ATTESTATIONS
+                // is used (rather than --provenance=false) so the same switch applies to both
+                // `docker buildx build` here and `docker buildx bake` in the TeamCity Linux
+                // image build (see OctopusDeploy/TeamCity-Configuration).
                 settings = settings
                     .SetOutput("type=image,oci-mediatypes=true,push=true")
-                    .AddProcessAdditionalArguments("--provenance=false");
+                    .AddProcessEnvironmentVariable("BUILDX_NO_DEFAULT_ATTESTATIONS", "1");
             }
             else
             {
--- a/build/Build.Pack.cs
+++ b/build/Build.Pack.cs
@@ -727,8 +727,13 @@ partial class Build
             if (includeDebugger)
                 tag += "-debug";
 
+            // Capture one timestamp and reuse it for both the BUILD_DATE build arg (which feeds
+            // the Dockerfile's org.opencontainers.image.created LABEL) and the created annotation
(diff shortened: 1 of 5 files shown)


todthomson:
## Code review

Reviewed at current state (5 commits; 5 files). Substantive change is `Build.Pack.cs` (k8s agent image); the 4 Dockerfiles are label modernization + keep-in-sync comments.

### What it does
- **`Build.Pack.cs`** (k8s agent image): on push, emits `type=image,oci-mediatypes=true,push=true` with OCI `org.opencontainers.image.*` annotations on the index + each platform manifest, and sets `BUILDX_NO_DEFAULT_ATTESTATIONS=1`. Load path unchanged. A single `buildDate` is reused for the `BUILD_DATE` build-arg and the `created` annotation.
- **4 Dockerfiles**: `org.label-schema.*` → `org.opencontainers.image.*` (SPDX `Apache-2.0`), plus keep-in-sync pointer comments.

### Correctness — solid (compile-verified, behaviour verified locally)
- `oci-mediatypes=true` + `annotation-index.*`/`annotation.*` → annotations confirmed on **both** the index and per-platform manifests (incl. values with spaces).
- `BUILDX_NO_DEFAULT_ATTESTATIONS=1` via `AddProcessEnvironmentVariable` → no attestation index; cleaner than a post-PATH `--provenance=false`, and aligned with #112/#113.
- `buildDate` scoping correct; compiles 0/0.
- Works with the k8s TeamCity step (which `--use`s a `docker-container` builder; `type=image,push=true` requires exactly that).

### Issues / notes
- **`--output` CSV comma fragility — now guarded.** Annotations are folded into the comma-separated `--output` option string, so a comma in a value would silently corrupt it. Safe today (no value has a comma), but the `description` was the realistic future candidate. Addressed: the build now throws a clear error if any annotation value contains a comma. (NUKE's `DockerBuildxBuild` has no `--annotation` setter, so the `--output` CSV is the pragmatic mechanism; the guard future-proofs it.)
- **(info) Annotations only on the push path** — expected; the load/dev path doesn't need them. The config `created` LABEL is still set on both paths via the build-arg.
- **(info) Metadata duplication** — mitigated via the keep-in-sync comments.

### Conventions / tests
- Comments are excellent — they explain the *why* (media-type mixing, the env-var choice, the annotation rationale) and point at the sibling LABELs.
- No automated tests — inherent for a buildx-invocation change; covered by post-merge `imagetools inspect` verification (expect 2 platform manifests, no attestation entries, all-OCI layers, annotations present).

### Verdict
Clean, correct, minimal, compile-verified. No blockers. The one substantive review point (`--output` CSV comma fragility) is now guarded; everything else is informational.

bec-callow-oct (approved):

LukeButters (approved):
I guess this makes sense, I didn't know about OCI until today.

build/Build.Pack.cs
```diff
@@ -727,14 +727,68 @@ void BuildAndPushOrLoadKubernetesTentacleContainerImage(bool push, bool load, st
...
+                // is used (rather than --provenance=false) so the same switch applies to both
+                // `docker buildx build` here and `docker buildx bake` in the TeamCity Linux
+                // image build (see OctopusDeploy/TeamCity-Configuration).
+                //
+                // We also stamp the OCI image-spec annotations (org.opencontainers.image.*)
+                // onto both the image index and each platform manifest. The Dockerfile LABELs
+                // set the same metadata on the image *config*; annotations are the spec-preferred
+                // location that registries and OCI tooling read from the manifest/index.
+                // Keep these values in sync with the org.opencontainers.image.* LABELs in
+                // docker/kubernetes-agent-tentacle/Dockerfile.
```
matt-richardson, line 758: Claude is over commenting - can we trim it down to things that future people will need when looking at the code.

Reasoning for "why are we making the changes" are much better as comments on a PR IMHO.

build/Build.Pack.cs
```diff
@@ -727,14 +727,68 @@ void BuildAndPushOrLoadKubernetesTentacleContainerImage(bool push, bool load, st
...
+                // list, so a comma in a value would corrupt it (the keys are fixed and comma-free).
+                var output = "type=image,oci-mediatypes=true,push=true";
+                foreach (var (key, value) in annotations)
+                {
+                    if (value.Contains(','))
+                        throw new InvalidOperationException(
+                            $"OCI annotation '{key}' value must not contain a comma; it would break the buildx --output option list: '{value}'");
+
+                    output += $",annotation-index.{key}={value},annotation.{key}={value}";
+                }
```
APErebus, line 782: We are controlling the annotations above and `FullSemVer` and `buildDate` will never have comma's, so this is a completely unnecessary guard + comment

todthomson:
Closing this: the customer issue it targeted (FD-492 / LEV-1001) has since been fixed another way.

**What fixed it:** #1301 moved the Linux Tentacle image base to Ubuntu 24.04, whose base layers are already OCI-typed, so the published manifest no longer mixes Docker and OCI layer media types. #1304 dropped Windows containers, so the `octopusdeploy/tentacle` index is now `linux/amd64` only.

**Verified:** `octopusdeploy/tentacle:latest` (12.2.4485) has 13/13 layers of type `application/vnd.oci.image.layer.v1.tar+gzip`, and `buildah bud` with `FROM` that image succeeds. We've asked the customer to retest; we'll reopen if it still fails for them.

**Why not merge the k8s part anyway:** the `kubernetes-agent-tentacle` layers were already all-OCI. The only remaining difference is the two `unknown/unknown` attestation entries in the index, which Podman, containerd and Kubernetes ignore, and nobody has reported a problem with them. The value doesn't justify the review cost.

Thanks @bec-callow-oct, @LukeButters, @matt-richardson and @APErebus for the reviews :slightly_smiling_face:

PR #1263, opened 2026-06-19
Add support for setting polling connection count on tentacles. Tentacle Workers default to max(5, core count).

### LINKED ISSUE #1267: Support configuring the number of polling connections on Tentacles
## Problem

Halibut performs one RPC per TCP connection. For polling Tentacles this means:

* A single long-running RPC can starve other RPCs, causing them to queue and potentially fail.
* Throughput does not scale with the Tentacle's hardware — e.g. a 4-core Tentacle can still only process one RPC at a time.

## Proposal

Allow the number of polling connections a Tentacle opens to each Octopus Server to be configured and persisted, so polling Tentacles can open multiple connections and scale with their hardware.

* New persisted setting `Tentacle.Communication.PollingConnectionCount`.
* Resolution at startup: `TentaclePollingConnectionCount` environment variable > persisted config value > default of 1, coerced to `1 .. 512`.
* Sensible registration defaults (polling registrations only, when not already set): workers default to `Math.Max(5, ProcessorCount)`, deployment targets default to 1.
* Configurable via the `configure --pollingConnectionCount` option, a new `set-polling-connection-count` command, and the Linux Docker container's `PollingConnectionCount` environment variable.

LukeButters:
# Background

Halibut can do one RPC per TCP connection, this is problem for Polling Tentacles since:
* A single long running RPC can cause other RPCs to be starved and fail.
* It does not scale with the Tentacle e.g. a 4 core tentacle can still only do one RPC at a time.

With this change going forward newly configured tentacle workers will have the number of Polling connections set to either 5 or the number of processors at configuration time (whatever is larger). Tentacle targets will remain with a connection count of 1, but this can be configured.

## PollingConnectionCount Setting
Additionally this adds a new configuration setting:
```
<set key="Tentacle.Communication.PollingConnectionCount">16</set>
```
To the Tentacle configuration xml to set (and persist) the number of polling connections.

Note that:
* The environment variable `TentaclePollingConnectionCount` overrides the value in config.
* The actual number of connections is set to be within `0 < NumberOfPollingConnections <= 512`.

## New pollingConnectionCount option added to the configure command.

Add an option to the `configure` command: `--pollingConnectionCount=N` sets and persists the polling connection count. Values less than 1 are rejected with a controlled failure.

## New set-polling-connection-count command.

The number of polling connections to use can be set by:
`Tentacle set-polling-connection-count --instance=MyInstance --pollingConnectionCount=N`

## Updated limits for Max Polling Connection Count

Updates the value from 10K to a fixed cap of `512`, which is a more sensible ceiling on the number of polling connections a single Tentacle should open.

# Results

This lets polling Tentacles open more than a single TCP connection to the Octopus Server, so that long-running RPCs no longer starve other RPCs and throughput scales with the Tentacle's core count. Workers default to a higher connection count (at least 5, or the processor count) while deployment targets default to 1, and operators can override the count via the `configure` command, the new `set-polling-connection-count` command, or the `TentaclePollingConnectionCount` environment variable.

Verified end-to-end against a local Octopus: a registered worker opened the expected number of live polling connections, and the `TentaclePollingConnectionCount` environment variable correctly overrode the persisted config value.

Fixes: https://github.com/OctopusDeploy/OctopusTentacle/issues/1267

Fixes EFT-3416

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.Tentacle.Tests/Commands/SetPollingConnectionCountCommandFixture.cs
+++ b/source/Octopus.Tentacle.Tests/Commands/SetPollingConnectionCountCommandFixture.cs
@@ -1,0 +1,57 @@
+using System;
+using FluentAssertions;
+using NSubstitute;
+using NUnit.Framework;
+using Octopus.Tentacle.Commands;
+using Octopus.Tentacle.Configuration;
+using Octopus.Tentacle.Configuration.Instances;
+using Octopus.Tentacle.Core.Diagnostics;
+using Octopus.Tentacle.Startup;
+
+namespace Octopus.Tentacle.Tests.Commands
+{
+    [TestFixture]
+    public class SetPollingConnectionCountCommandFixture : CommandFixture<SetPollingConnectionCountCommand>
+    {
+        StubTentacleConfiguration tentacleConfiguration;
+
+        [SetUp]
+        public override void SetUp()
+        {
+            base.SetUp();
+
+            tentacleConfiguration = new StubTentacleConfiguration();
+            var log = Substitute.For<ISystemLog>();
+            var selector = Substitute.For<IApplicationInstanceSelector>();
+            selector.Current.Returns(info => new ApplicationInstanceConfiguration(null, null!, null!, null!));
+            Command = new SetPollingConnectionCountCommand(
+                new Lazy<IWritableTentacleConfiguration>(() => tentacleConfiguration),
+                selector,
+                log,
+                Substitute.For<ILogFileOnlyLogger>());
+        }
+
+        [Test]
+        public void ShouldSetThePollingConnectionCount()
+        {
+            Start("--count=5");
+            tentacleConfiguration.PollingConnectionCount.Should().Be(5);
+        }
+
+        [Test]
+        public void ShouldFailWhenNoCountProvided()
+        {
+            Action start = () => Start();
+            start.Should().Throw<ControlledFailureException>();
+            tentacleConfiguration.PollingConnectionCount.Should().BeNull();
+        }
+
+        [Test]
+        public void ShouldFailWhenCountIsLessThanOne()
+        {
+            Action start = () => Start("--count=0");
+            start.Should().Throw<ControlledFailureException>();
+            tentacleConfiguration.PollingConnectionCount.Should().BeNull();
+        }
+    }
+}
--- a/source/Octopus.Tentacle.Tests/Commands/SetPollingConnectionCountCommandFixture.cs
+++ b/source/Octopus.Tentacle.Tests/Commands/SetPollingConnectionCountCommandFixture.cs
@@ -34,7 +34,7 @@ namespace Octopus.Tentacle.Tests.Commands
         [Test]
         public void ShouldSetThePollingConnectionCount()
         {
-            Start("--count=5");
+            Start("--pollingConnectionCount=5");
             tentacleConfiguration.PollingConnectionCount.Should().Be(5);
         }
 
@@ -49,8 +49,16 @@ namespace Octopus.Tentacle.Tests.Commands
         [Test]
         public void ShouldFailWhenCountIsLessThanOne()
         {
-            Action start = () => Start("--count=0");
+            Action start = () => Start("--pollingConnectionCount=0");
             start.Should().Throw<ControlledFailureException>();
+            tentacleConfiguration.PollingConnectionCount.Should().BeNull();
+        }
+
+        [Test]
+        public void ShouldFailWithAFriendlyErrorWhenCountIsNotANumber()
+        {
+            Action start = () => Start("--pollingConnectionCount=notanumber");
+            start.Should().Throw<ControlledFailureException>().WithMessage("*not a valid whole number*");
             tentacleConfiguration.PollingConnectionCount.Should().BeNull();
         }
     }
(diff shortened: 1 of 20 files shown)


source/Octopus.Tentacle/Communications/HalibutInitializer.cs
```diff
@@ -115,30 +114,43 @@ void AddPollingEndpoints()
             }
         }
                 
-        // We don't expect anyone to hit this, but the theoretical cap is the number of allocatable outbound ports
-        const int MaximumPollingConnectionCount = 10000;
+        // Limit Polling connections to something within the realm of reasonable.
+        // We allow polling connection count to scale with core count, if someone has a 32-core machine, 128 connections might not be unreasonable. 
+        static readonly int MaximumPollingConnectionCount = Math.Max(32, Environment.ProcessorCount * 4);
```
APErebus: I wonder if we need to be so protective? Why not just set a fixed limit of `512` or something and leave it up to the user to do that if they want?

source/Octopus.Tentacle/Communications/HalibutInitializer.cs
```diff
@@ -115,30 +114,43 @@ void AddPollingEndpoints()
...
+                connectionCount = (uint)MaximumPollingConnectionCount;
+            }
+            
+            if(connectionCount <= 0) {
                     log.InfoFormat("The requested polling connection count must be greater than 0, setting to 1");
                     connectionCount = 1;
-                    break;
             }
 
             log.InfoFormat("Starting {0} polling connections", connectionCount);
```
APErebus: I wonder if this log line belongs higher up. Doesn't quite make sense here 🤔

source/Octopus.Tentacle/Commands/ConfigureCommand.cs
```diff
@@ -80,6 +80,15 @@ public ConfigureCommand(
                 }
                 VoteForRestart();
             }));
+            Options.Add("pollingConnectionCount=", "The number of polling connections this Tentacle should open to each Octopus Server it polls. Only applies to polling Tentacles.", v => QueueOperation(delegate
+            {
+                var pollingConnectionCount = int.Parse(v);
```
APErebus: Probably should be a tryparse for safety? Or do we want it to explode 🤔

APErebus (approved):
Love it!

source/Octopus.Tentacle/Communications/HalibutInitializer.cs
LukeButters: Yeah I guess I was thinking of limiting low core count tentacles from having insane levels of concurrency, while also allowing the core count to just keep scaling.

Perhaps just setting 512 is easier to reason about and good enough

source/Octopus.Tentacle/Commands/ConfigureCommand.cs
LukeButters: It now explodes with a friendly message.

source/Octopus.Tentacle.Tests/Commands/SetPollingConnectionCountCommandFixture.cs
```diff
@@ -0,0 +1,65 @@
...
+            selector.Current.Returns(info => new ApplicationInstanceConfiguration(null, null!, null!, null!));
+            Command = new SetPollingConnectionCountCommand(
+                new Lazy<IWritableTentacleConfiguration>(() => tentacleConfiguration),
+                selector,
+                log,
+                Substitute.For<ILogFileOnlyLogger>());
+        }
+
+        [Test]
+        public void ShouldSetThePollingConnectionCount()
```
xwipeoutx, line 35: Need a test for a super big number, test the clamping? Is the intention here that you can set it to whatever, but it'll just clamp it at runtime (found in tests below)?

source/Octopus.Tentacle/Communications/HalibutInitializer.cs
xwipeoutx: I'm not opposed to allowing advanced users to foot-gun, if it's easier to reason about.  Makes docs a bit saner, too.

source/Octopus.Tentacle/Communications/HalibutInitializer.cs
xwipeoutx: Agreed - this method is side-effect-free.  It's not starting anything. For some reason I'm ok with the other log messages though :D

xwipeoutx (approved):
PR description has

> The environment variable `todo` overrides the value on config.

What are your plans for advertising this change? Public docs etc.

In any case, approving, this is a good change.

source/Octopus.Tentacle.Tests/Commands/SetPollingConnectionCountCommandFixture.cs
LukeButters: Yeah I just did runtime clamping. Not sure if I have strong opinions

source/Octopus.Tentacle/Communications/HalibutInitializer.cs
LukeButters: Setting to 512

source/Octopus.Tentacle/Communications/HalibutInitializer.cs
LukeButters: moved

PR #1277, opened 2026-08-04
Stop corrupt script logs failing script status reads

akirayamamoto:


akirayamamoto:
Superseded by #1278, which drops the reader change and keeps this to diagnostics plus the two fixes.

PR #1278, opened 2026-08-04
Make a corrupt script log diagnosable

akirayamamoto:
# Background

The Tentacle nightly failed with an unhandled `JsonReaderException` out of `ScriptLog.GetOutput`. The real failure was one test in [Integration Test: net8.0 on Windows 2012 R2 #9.2.4244-main-nightly](https://build.octopushq.com/buildConfiguration/TeamFireAndMotion_OctopusTentacle_TentacleVLatest_net80_IntegrationTestNet80onWindows2012r2/23575913); the chain build just relayed it. That test has passed 39 of its last 40 runs, so this is a flake, not a regression.

I could not work out why the log was corrupt. The test deletes its workspace at the end of the method and nothing prints that file anywhere, so there was nothing left by the time I looked.

So this PR adds the diagnostics we were missing. Next time it happens there should be enough to go on.

# Results

- When the log cannot be parsed, the error now says how big it was and whether a write arrived after it closed. Before it just quoted the JSON token that tripped the parser, which told you nothing about how the log got that way. This is the only change a user sees.
- A failing integration test prints the script log into the build output. Nothing printed it before, and the workspace was gone by teardown.
- Closing the log takes the same lock as every write. It could previously close midway through a write and leave half an entry on disk.
- Writes arriving after the log closed are refused rather than attempted, so they cannot leave a partial entry either.
- Failed writes get logged. The try/catch was around the wrong code, so this had been failing silently.
- Abandoning a script logs whether its task finished and what state it was in, which tells us if a live writer was left on the log.

A corrupt log still fails the step, exactly as it does today. The behaviour is unchanged.

## Before

The parse error, with nothing to explain it:

```
Newtonsoft.Json.JsonReaderException : JsonToken EndArray is not valid for closing JsonType None. Path '', line 1, position 1450.
   at Newtonsoft.Json.JsonReader.ValidateEnd(JsonToken endToken)
   at Octopus.Tentacle.Core.Services.Scripts.Logging.ScriptLog.GetOutput(Int64 afterSequenceNumber, Int64& nextSequenceNumber)
```

## After

Same failure, now saying what state the log was in:

```
Could not parse the script log. It is 56 bytes, and a write was refused after it closed. Parser reported: JsonToken EndArray is not valid for closing JsonType None. Path '', line 1, position 53.
```

And from a real build, forced to fail on a Linux agent to check this reaches the build output:

```
Abandoning the script for task 1312161c...; its task had not finished, status WaitingForActivation
Could not write script output to log, for task 1312161c.... This script's log is closed, so any
further output it produces will be discarded.
System.ObjectDisposedException: The script log writer has been disposed...
### SCRIPT LOG .../script-logs/0-1419/l8MFdaAcskqTMUz3sjNxaw.log (832 bytes) ###
["stdout","##octopus[stdout-verbose]","2026-08-04T08:28:09.2980966+00:00"]["stdout","Acquiring isolation mutex RunningScript with FullIsolation in 1312161c...","2026-08-04T08:28:09.2987699+00:00"]...
### END SCRIPT LOG ###
```

That middle warning is the interesting one. It fired on a real agent, so the abandoned script really does try to write after its log has closed. That was silent before this.

# How to review this PR

The claim worth checking is that behaviour does not change. `ShouldHandleIncompleteLine` is the guard for it. I did not touch the existing corrupt-log handling, so that test still asserts the same message it always did.

Two production changes to think about: `Writer.Dispose` now takes the lock, and writes after the log closes are refused. The rest is log lines and test code.

`ScriptLog` is shared by net48 and net8.0. I could only run net8.0 locally, so the net48 side rests on the build.

# Follow ups

Right now the log is only text in the build output. Copying it into `artifacts/trace-logs` would publish it as a proper artifact, since these configs already have a rule for that path. Needs a build to prove it works, so not in this PR.

### DIFF
--- a/source/Octopus.Tentacle.Core/Services/Scripts/Logging/ScriptLog.cs
+++ b/source/Octopus.Tentacle.Core/Services/Scripts/Logging/ScriptLog.cs
@@ -25,7 +25,14 @@ namespace Octopus.Tentacle.Core.Services.Scripts.Logging
 
         public IScriptLogWriter CreateWriter()
         {
-            return new Writer(logFile, fileSystem, sync, sensitiveValueMasker);
+            return new Writer(this);
+        }
+
+        string MaskSensitiveValues(string rawMessage)
+        {
+            string? maskedMessage = null;
+            sensitiveValueMasker.SafeSanitize(rawMessage, s => maskedMessage = s);
+            return maskedMessage ?? rawMessage;
         }
 
         public List<ProcessOutput> GetOutput(long afterSequenceNumber, out long nextSequenceNumber)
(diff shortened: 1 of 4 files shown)


akirayamamoto:
I am closing this. This flaky test seems to be very rare and I am not the most familiar with this repo to proceed with this investigation. It could be useful to look at the next time it fails.

PR #1279, opened 2026-08-04
DO NOT MERGE: CI verification for #1278 script log capture

akirayamamoto:
Throwaway branch. Do not review, do not merge, will be closed shortly.

This forces `WhenPowerShellNeverStarts_AndWorkspaceIsDeletedBeforeScriptRuns` to fail on a real build agent, so I can confirm the script log capture added in #1278 actually reaches the build output. Verified locally, not yet on TeamCity.

Expect `Integration Test: net8.0 on Windows 2012 R2` and `Integration Test: Windows` to go red on purpose.


akirayamamoto:
Verification done. All three signals reach the TeamCity build output on a real agent: the abandon log line, the post-dispose write warning, and the captured script log. Notably the warning fired, confirming an orphaned task really does write after the log closes. Closing; no changes here are intended for main.

PR #1284, opened 2026-08-13
Take the lock when closing a script log

akirayamamoto:
# Background

`ScriptLog` guards access to one log file with a single lock. Reads take it and writes take it, but closing does not. Closing is not inert either, since it completes any JSON array still open, so in principle a close could overlap a write that is partway through.

I came across this while looking at a nightly that failed reading a corrupt script log, [build 23575913](https://build.octopushq.com/buildConfiguration/TeamFireAndMotion_OctopusTentacle_TentacleVLatest_net80_IntegrationTestNet80onWindows2012r2/23575913). That has happened once and I could not tie it to this, so it is more where I came across the gap than a reason to merge.

# Results

Closing now takes the same lock as writes, so a close and a write cannot overlap.

Nothing else changes. A corrupt log still fails the caller as it does today, and writes that arrive after the log has closed are untouched.

# How to review this PR

Small enough to read in full.

Two things I would welcome a second opinion on. There is no `disposed` flag, so a double dispose calls through twice, which looked harmless when I tried it. And `Dispose` now waits on a write, though it runs on a background task rather than a request thread.


source/Octopus.Tentacle.Core/Services/Scripts/Logging/ScriptLog.cs
```diff
@@ -153,9 +153,12 @@ string MaskSensitiveValues(string rawMessage)
 
             public void Dispose()
             {
-                json.Close();
-                writer.Dispose();
-                writeStream.Dispose();
+                lock (sync)
```
sburmanoctopus, line 156: It's interesting that this is the same lock object as the parent `ScriptLog`.

I don't know if disposing the json or writer objects cause changes to the underlying file. Maybe they do. If so, then yes, we would want disposing and reading (in `ScriptLog`) to wait for each other.

sburmanoctopus (approved):

source/Octopus.Tentacle.Core/Services/Scripts/Logging/ScriptLog.cs
akirayamamoto: Yes, they do. I checked with a small repro: closing the writer with an array still open finishes it off on disk, because `AutoCompleteOnClose` defaults to true. Disposing the `StreamWriter` also flushes whatever is buffered.

akirayamamoto:
I will hold this until I get [GH access](https://apps.octopushq.com/accessrequest/01a06997-b719-7a3b-80ff-5437da69d8cd) so I can merge myself.

PR #1293, opened 2026-09-08
LEV-1830: Skip Tentacle 5.0.x compatibility tests when the host has no OpenSSL 1.x

### LINKED ISSUE #1294: Debian 11 (Bullseye) reached EOL on 31 August 2026 and we can no longer build and test docker/linux/Dockerfile
### Team

- [x] I've assigned a team label to this issue

### What happened?

Debian 11 (Bullseye) reached EOL on 31 August 2026.

Subsequently, from 5 September 2026, our internal TeamCity `docker/linux/Dockerfile` wasn't able to be built, because the upstream Debian 11 dependencies are no longer available, which has forced us to reconsider the base of our Linux Tentacle Dockerfile.

Based on customer usage (and internal decision making e.g. to align our base with Microsoft's .NET base images in the future) we have determined that we should move our Linux container base to Ubuntu.

Ubuntu 22.04 (Jammy Jellyfish) is (roughly speaking) based on Debian 11, so it's the closest thing to move to first, which is why we've moved first to Ubuntu 22.04.

We plan to move to Ubuntu 24.04 next, then we may move to something like `[mcr.microsoft.com/dotnet/runtime-deps](https://github.com/dotnet/dotnet-docker/blob/main/README.runtime-deps.md)` after that, post the Tentacle .NET 10 upgrade.

### Reproduction

n/a

### Error and Stacktrace

```shell
n/a
```

### More Information

n/a

### Workaround

n/a

todthomson:
# Summary (Tod-written)

Fixes: https://github.com/OctopusDeploy/OctopusTentacle/issues/1294.

Stacked on top of: https://github.com/OctopusDeploy/OctopusTentacle/pull/1292.

[Tim Nicholas upgraded the Ubuntu TeamCity agent from `22.04` to `24.04`](https://octopusdeploy.slack.com/archives/C02SY496BAM/p1788739316594879).

This is fine (and to be expected, as per our developer platform upgrade plans), but we missed that announcement, as we weren't in [#feed-build-platform-change-log](https://octopusdeploy.slack.com/archives/C02SY496BAM).

Ubuntu 24.04 doesn't provide OpenSSL 1.x (aka `libssl1.1`).

Thus, when we try to run Tentacle `5.0.x` compatibility tests on Ubuntu `24.04`, they will fail, because the Tentacle cannot boot up when it can't access `libssl1.1`.

This is because Tentacle `5.0.x` was compiled against .NET Core 3.1, which requires  `libssl1.1`, which (as above) is not provided on Ubuntu 24.04.

This PR stops Tentacles `5.0.x` being added to the compatibility tests list whenever `libssl1.1` is not available.

# Details (Claude-written)

Fixes [LEV-1830](https://linear.app/octopus/issue/LEV-1830/integration-tests-tentacle-50x-compatibility-cases-fail-on-linux-x64).

`Integration Test: Linux (not sudo)` is failing **38 tests on every branch** — `main` included — which red-lights `Gate: Integration Tests` and `Chain: Full Chain` on every PR in the repo.

## Cause

All 38 failures are backwards-compatibility cases for Tentacle `5.0.4`, `5.0.12` and `5.0.15`. **Zero failures on the current version.** Those old binaries are downloaded onto the agent and run directly:

```
/root/.local/share/TentacleBinaryCache.../net8.0/5.0.15/Tentacle create-instance ...
[Tentacle] No usable version of libssl was found          ×11, then SIGABRT (exit 134)
```

Tentacle 5.0.x is a .NET Core 3.1 build: it loads OpenSSL 1.x at start-up and aborts when it can't find one. The agent (`nautilus-linux-7964510`, linux-x64) now has only OpenSSL 3. Nothing in the repo changed — last green `main` was main-nightly `24107896` (6 Sep), first red `24129127` (7 Sep).

## Fix

`TentacleVersions` already excluded these exact three versions for MacOS and linux-arm64. This extends that same mechanism to cover "the host has no OpenSSL 1.x", probing with `NativeLibrary.TryLoad` so the dynamic linker's own search path is used rather than guessing at library locations.

Excluded versions are removed from the test case source by `TentacleConfigurationsAttribute`, so the cases stop being *generated* rather than being generated and failing.

The two near-duplicate version-5 conditions become one predicate, so the download list and the case source can't drift — a version in the case source but missing from `AllTestedVersionsToDownload` makes `TentacleFetcher` throw. It must stay declared above both public fields, since static field initialisers run in textual order.

### Behaviour is unchanged everywhere except the broken case

| Host | 5.0.x | |
|---|---|---|
| Windows | runs | unchanged |
| MacOS | excluded | unchanged |
| linux-arm64 | excluded | unchanged |
| linux-x64 **with** OpenSSL 1.x | runs | unchanged |
| linux-x64 **without** OpenSSL 1.x | excluded | **was: ran and aborted** |

## Verification

- Builds clean, 0 warnings.
- The detection code, published self-contained and run in containers, returns `True` on `ubuntu:20.04` with `libssl1.1` installed (`libssl.so.1.1`) and `False` on `ubuntu:22.04` with only `libssl3` (`libssl.so.3`).
- **The skip itself is not verified locally.** This checkout is arm64, where all three versions are already excluded, so the new branch never executes here. Proving the gate goes green needs CI.

## This is a mitigation, not the root-cause fix

The underlying fragility is that `Integration Test: Linux (not sudo)` runs on a **bare agent with no pinned container**, so it silently inherits whatever OS the `nautilus-linux` pool happens to be on. Its two siblings in the same project already pin one:

| Config | Pinned image |
|---|---|
| `Integration Test: Linux (not sudo)` | **none — bare agent** |
| `Integration Test: Ubuntu 22.04 LTS net8.0` | `mcr.microsoft.com/dotnet/runtime-deps:8.0-jammy` |
| `Integration Test: Ubuntu 24.04 LTS net8.0` | `mcr.microsoft.com/dotnet/runtime-deps:8.0-noble` |

It has the Docker-wrapper params (`plugin.docker.imagePlatform`, `plugin.docker.pull.enabled`) but no `plugin.docker.imageId`, so the wrapper never containerises the step. Fixing that is a `TeamCity-Configuration` change and belongs in its own PR.

Note the two changes **compose** rather than compete: pinning an image that carries OpenSSL 1.x makes `CanLoadOpenSsl1x` true, so the 5.0.x cases run again and this skip becomes a no-op there. Pinning `8.0-jammy` as-is would *not* fix the failure on its own — jammy ships only `libssl.so.3`, so the 5.0.x cases would still abort. Either the pinned image provides OpenSSL 1.x, or this skip is what keeps the suite honest.

## Alternative considered

Installing `libssl1.1` from Ubuntu 20.04 onto the agent. Rejected: it puts an EOL, unpatched OpenSSL on a shared build agent, and it is an agent-image change rather than a repo change — so it would silently regress the next time the agent image is rebuilt, which is most likely what caused this.

Worth a separate product decision: whether compatibility testing still needs to reach Tentacle 5 at all. The code comment dates `5.0.4` to 9 Sep 2019 and .NET Core 3.1 went EOL in Dec 2022. Dropping those three versions outright would remove the EOL-OpenSSL dependency permanently, but that call isn't made here.

### DIFF
--- a/docker/linux/Dockerfile
+++ b/docker/linux/Dockerfile
@@ -23,6 +23,7 @@ RUN apt-get update && \
         ca-certificates \
         libicu70 \
         libssl3 && \
+    apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
 ARG BUILD_NUMBER
--- a/docker/linux/Dockerfile
+++ b/docker/linux/Dockerfile
@@ -2,22 +2,6 @@ FROM ubuntu:22.04
 ENV ASPNETCORE_URLS=http://+:80 DOTNET_RUNNING_IN_CONTAINER=true
 
 # .NET 8 runtime dependencies.
-#
-# libc6, libgcc-s1, libgssapi-krb5-2, libstdc++6 and zlib1g were dropped from
-# this list: ubuntu:22.04 already ships them at the newest version jammy
-# offers, so naming them installed nothing.
-#
-# The three below are NOT redundant, and the distinction matters:
-#   ca-certificates  not in the base image at all
-#   libicu70         not in the base image at all
-#   libssl3          in the base, but at 3.0.2-0ubuntu1.26 while jammy-security
-#                    has 3.0.2-0ubuntu1.29. Naming it is what pulls that
-#                    security update in, so it has to stay.
-#
-# Beware when trimming this further: a package that is a no-op today stops
-# being one the moment the base image lags behind jammy-security. The smoke
-# tests in testing/docker-linux/build-and-test-linux-docker-image.sh assert all
-# eight are present in the built image regardless of what installed them.
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
         ca-certificates \
--- a/docker/linux/Dockerfile
+++ b/docker/linux/Dockerfile
@@ -1,15 +1,28 @@
 FROM ubuntu:22.04
 ENV ASPNETCORE_URLS=http://+:80 DOTNET_RUNNING_IN_CONTAINER=true
+
+# .NET 8 runtime dependencies.
+#
+# libc6, libgcc-s1, libgssapi-krb5-2, libstdc++6 and zlib1g were dropped from
+# this list: ubuntu:22.04 already ships them at the newest version jammy
+# offers, so naming them installed nothing.
+#
+# The three below are NOT redundant, and the distinction matters:
+#   ca-certificates  not in the base image at all
+#   libicu70         not in the base image at all
+#   libssl3          in the base, but at 3.0.2-0ubuntu1.26 while jammy-security
+#                    has 3.0.2-0ubuntu1.29. Naming it is what pulls that
+#                    security update in, so it has to stay.
+#
+# Beware when trimming this further: a package that is a no-op today stops
+# being one the moment the base image lags behind jammy-security. The smoke
+# tests in testing/docker-linux/build-and-test-linux-docker-image.sh assert all
+# eight are present in the built image regardless of what installed them.
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
         ca-certificates \
-        libc6 \
-        libgcc-s1 \
-        libgssapi-krb5-2 \
         libicu70 \
-        libssl3 \
-        libstdc++6 \
-        zlib1g && \
+        libssl3 && \
     rm -rf /var/lib/apt/lists/*
 
 ARG BUILD_NUMBER
@@ -43,7 +56,7 @@ RUN /install-scripts/install-docker.sh
 # Install Tentacle
 COPY _artifacts/deb/tentacle_${BUILD_NUMBER}_amd64.deb /tmp/
 RUN apt-get update && \
-    apt install ./tentacle_${BUILD_NUMBER}_amd64.deb && \
+    apt-get install -y --no-install-recommends ./tentacle_${BUILD_NUMBER}_amd64.deb && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/* && \
     ln -s /opt/octopus/tentacle/Tentacle /usr/bin/tentacle
--- a/docker/linux/Dockerfile
+++ b/docker/linux/Dockerfile
@@ -1,13 +1,13 @@
-FROM debian:11-slim
+FROM ubuntu:22.04
 ENV ASPNETCORE_URLS=http://+:80 DOTNET_RUNNING_IN_CONTAINER=true
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
         ca-certificates \
         libc6 \
-        libgcc1 \
+        libgcc-s1 \
         libgssapi-krb5-2 \
-        libicu67 \
-        libssl1.1 \
+        libicu70 \
+        libssl3 \
         libstdc++6 \
         zlib1g && \
     rm -rf /var/lib/apt/lists/*
--- a/docker/linux/install-scripts/install-docker.sh
+++ b/docker/linux/install-scripts/install-docker.sh
@@ -7,21 +7,49 @@ set -eux
 # Add the apt sources for Docker (they're not part of the stock Ubuntu distro).
 apt-get update
 
+# Only ca-certificates and curl are needed to add the repository. Three
+# packages that used to be installed here are not:
+#   apt-transport-https - a transitional stub on jammy ("transitional package
+#                         for https support"); apt 2.4 ships
+#                         /usr/lib/apt/methods/https itself.
+#   gnupg               - was only needed for `gpg --dearmor`. apt reads the
+#                         ASCII-armoured key directly via Signed-By and verifies
+#                         it with gpgv, which apt itself depends on.
+#   lsb-release         - replaced by /etc/os-release, which is always present.
 apt-get install -y --no-install-recommends \
-    apt-transport-https \
     ca-certificates \
-    curl \
-    gnupg \
-    lsb-release
+    curl
 
-curl -fsSL https://download.docker.com/linux/ubuntu/gpg | gpg --dearmor -o /usr/share/keyrings/docker-archive-keyring.gpg
+# This follows Docker's documented setup for Ubuntu: the ASCII-armoured key
+# under /etc/apt/keyrings, referenced from a deb822 .sources file rather than a
+# one-line .list. deb822 is supported by apt 2.4 on jammy.
+# https://docs.docker.com/engine/install/ubuntu/
+install -m 0755 -d /etc/apt/keyrings
+curl -fsSL https://download.docker.com/linux/ubuntu/gpg -o /etc/apt/keyrings/docker.asc
+chmod a+r /etc/apt/keyrings/docker.asc
 
-echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/docker-archive-keyring.gpg] https://download.docker.com/linux/ubuntu \
-  $(lsb_release -cs) stable" | tee /etc/apt/sources.list.d/docker.list > /dev/null
+# UBUNTU_CODENAME first so this still resolves on Ubuntu derivatives, which set
+# VERSION_CODENAME to their own release name. Deliberately unguarded: with
+# `set -u` an unset codename fails the build rather than writing empty Suites.
+cat > /etc/apt/sources.list.d/docker.sources <<EOF
+Types: deb
+URIs: https://download.docker.com/linux/ubuntu
+Suites: $(. /etc/os-release && echo "${UBUNTU_CODENAME:-$VERSION_CODENAME}")
+Components: stable
+Architectures: $(dpkg --print-architecture)
+Signed-By: /etc/apt/keyrings/docker.asc
+EOF
 
 
 # Install Docker and its runtime dependencies.
 # https://github.com/docker/docker/blob/master/project/PACKAGERS.md#runtime-dependencies
+#
+# iproute2 is here for `ip`, which dockerd-entrypoint.sh uses in _tls_san to
+# collect the container's addresses. dockerd itself does not need it, but
+# without it a dind TLS certificate is issued with DNS SANs only - verified as
+# DNS:docker,DNS:<hostname>,DNS:localhost and no IP: entries - so a client
+# connecting to the daemon by IP address fails hostname verification. Upstream
+# docker:dind ships iproute2 for the same reason.
 
 apt-get update
 apt-get install -y \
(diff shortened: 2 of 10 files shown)


todthomson:
# Local Claude Code - Code Review

Scope: the delta between this branch and `lev/tod/LEV-1803-move-linux-tentacle-Dockerfile-from-debian-to-ubuntu` only — i.e. the two commits that are actually LEV-1830. Verified throughout with `dotnet build Octopus.Tentacle.Tests.Integration -f net8.0` (0 warnings, 0 errors). The `net48` leg of `CommonTestUtils` can't be built from macOS — the csproj only lists that TFM on Windows — so CI still has to confirm it.

The approach is sound. `dlopen`-by-soname beats guessing at distro paths, and collapsing the two gates into one `Version5IsRunnableHere` predicate removes a real class of bug: the old code duplicated the OS/arch condition across `GetAllTestedVersionsToDownload` and `GetVersionsUnsupportedByCurrentOperatingSystemAndArchitecture`, and `TentacleFetcher` throws if those two ever disagree.

Eight findings, all discussed and resolved below. Seven produced changes; one was my error.

---

## 1. ~~PR base~~ — withdrawn, my mistake

I flagged that #1293's diff mixes in #1292's work with no signal. The description already opens with "Stacked on top of: #1292". Struck.

Residue is cosmetic only: `gh pr view 1293` still reports base `main`, 8 files, +1164/−46, so the *diff view* mixes the two even though the prose doesn't. Retargeting the base would fix that and GitHub auto-retargets to `main` on merge. Optional, not done.

## 2. Nothing signalled at runtime when Tentacle 5 coverage disappeared — **fixed**

The description reasons about this at the design level (the skip and a pinned image compose; `8.0-jammy` alone wouldn't fix it; dropping v5 is a separate product call). The runtime half was missing.

The excluded set used to be a pure function of OS + arch — readable off the source, invariant per agent. It is now a function of what's installed when the tests run. When `libssl1.1` is absent, every 5.0.x case vanished from the run: not skipped, not logged, just a smaller green count. That is the same failure mode this PR exists to work around — the agent moved 22.04 → 24.04 and it surfaced as 38 SIGABRTs rather than as a notice.

`WarmTentacleCache.OneTimeSetUp` now warns once, before the download loop:

```csharp
var excludedVersions = TentacleVersions.VersionsUnsupportedByCurrentOperatingSystemAndArchitecture;
if (excludedVersions.Count > 0)
{
    logger.Warning("Skipping Tentacle versions {Versions} - they are not runnable on this host ({RuntimeIdentifier}, Linux without OpenSSL 1.x: {IsLinuxWithoutOpenSsl1x})",
        string.Join<Version>(", ", excludedVersions),
        RuntimeInformation.RuntimeIdentifier,
        CommonTestUtilsPlatformDetection.IsLinuxWithoutOpenSsl1x);
}
```

`{RuntimeIdentifier}` is in there because without it the line on macOS/arm64 reads "not runnable … OpenSSL 1.x loadable: True", which is self-contradictory — a live demonstration of finding 3.

## 3. The probe claimed a fact it hadn't established — **fixed**

`CanLoadOpenSsl1x` returned `true` on Windows and macOS, where nothing was probed; `true` was standing in for "not applicable, don't let this exclude you". Safe for the only caller (which short-circuits both platforms first), but a trap for the next one — I walked into it writing the log line above.

Inverted to `IsLinuxWithoutOpenSsl1x`, which is true *only* where the fact was actually established. Non-Linux platforms fall on the `false` side and read correctly with no doc caveat. Caller becomes `return !PlatformDetection.IsLinuxWithoutOpenSsl1x;`.

## 4. The public array aliased the private one — **fixed**

`GetVersionsUnsupportedByCurrentOperatingSystemAndArchitecture()` returned `Version5Releases` itself, so the public field and the private one were the same instance. `readonly` protects the reference, not the contents. The old code allocated fresh per call, so the aliasing was new — theoretical (one `.Contains()` consumer), but gratuitous.

Both are now `IReadOnlyList<Version>`, which closes it rather than papering over it with a copy. `AllTestedVersionsToDownload` is untouched: it's `ToArray()`d from a local list, so it never aliased anything.

## 5. `#if` polarity — **fixed**

The `#else` returned `true` unconditionally, i.e. the *unsafe* answer for any future non-net8 TFM that isn't .NET Framework — on Linux that schedules v5 tests that then abort. Now `#if NETFRAMEWORK` → `false` / `#else` → probe, which names the actual reason (no `NativeLibrary` on .NET Framework) and fails closed.

## 6. Missing trailing newline — **fixed**

`PlatformDetection.cs` had no `\n` at EOF. Resolved as a side effect of 3.

## 7. Probe fired from the type initialiser — **fixed**

As a static field initialiser, the `dlopen` calls ran on first touch of *any* `PlatformDetection` member — `IsRunningOnWindows` included, which is used all over the test projects. On Linux that loaded OpenSSL 1.1 into a process already holding OpenSSL 3 via .NET 8, on behalf of callers with nothing to do with Tentacle versions.

Now a `Lazy<bool>`: same caching and same public shape, but nothing is loaded unless something asks about OpenSSL. The `dlopen` mechanism itself is kept — it's the faithful reproduction of what the runtime does, and co-resident 1.1/3.x is fine (1.1 uses versioned symbols, handle is `Free`d immediately).

## 8. The soname list was missing one the runtime tries — **fixed**

I originally raised this as low-confidence speculation about a bare `libssl.so` fallback. Checked it against [`opensslshim.c` on `dotnet/corefx` `release/3.1`](https://github.com/dotnet/corefx/blob/release/3.1/src/Native/Unix/System.Security.Cryptography.Native/opensslshim.c#L49-L99). `OpenLibrary()` tries, in order:

| # | soname | was in the probe? |
|---|---|---|
| 1 | `libssl.so.$CLR_OPENSSL_VERSION_OVERRIDE` | no (env-var driven) |
| 2 | `libssl.so.1.1` | yes |
| 3 | **`libssl.so.1.0.2`** — Debian 9's bumped soname | **no** |
| 4 | `libssl.so.1.0.0` | yes |
| 5 | `libssl.so.10` — Fedora-derived | yes |

So the bare-`libssl.so` speculation was **wrong** — no such fallback exists. But `libssl.so.1.0.2` was a real omission: on a host carrying only that soname the probe says "no OpenSSL 1.x" and skips v5, while the Tentacle would have started fine. Over-skipping, so it failed in the safe direction, but it was coverage lost for no reason.

Added, in the runtime's own order, with the source cited in the comment and the `CLR_OPENSSL_VERSION_OVERRIDE` omission called out as deliberate.

Incidentally this confirms the mechanism end to end: the shim is an `__attribute__((constructor))` that prints `No usable version of libssl was found` and then `abort()`s — exactly the exit 134 in the description.

---

*Reviewed locally with Claude Code (Opus 5) against `lev/tod/LEV-1803-move-linux-tentacle-Dockerfile-from-debian-to-ubuntu`. Fixes applied to the working tree, not committed.*

docker/linux/install-scripts/install-docker.sh
```diff
@@ -4,24 +4,52 @@ set -eux
 # This script is adapted from https://github.com/docker-library/docker/blob/master/19.03/dind/Dockerfile
 
 
-# Add the apt sources for Docker (they're not part of the stock Debian distro).
+# Add the apt sources for Docker (they're not part of the stock Ubuntu distro).
 apt-get update
 
+# Only ca-certificates and curl are needed to add the repository. Three
+# packages that used to be installed here are not:
+#   apt-transport-https - a transitional stub on jammy ("transitional package
```
xwipeoutx: No need to comment what we don't need.  I think this whole comment block can be removed

docker/linux/install-scripts/install-docker.sh
```diff
@@ -56,16 +85,23 @@ dos2unix /usr/local/bin/dind
...
-  # https://forums.docker.com/t/failing-to-start-dockerd-failed-to-create-nat-chain-docker/78269
-  update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
-  update-alternatives --set iptables /usr/sbin/iptables-legacy
-else
-  # There can be issues with some (newer) operating systems are moving away from iptables to nftables, like RHEL.
-  # See https://octopusdeploy.slack.com/archives/CNHBHV2BX/p1677028476905509 for more details
-  update-alternatives --set ip6tables /usr/sbin/ip6tables-nft
-  update-alternatives --set iptables /usr/sbin/iptables-nft
-fi
+# Newer operating systems (RHEL in particular) have moved from iptables to
```
xwipeoutx: Given our `Dockerfile` points at Ubuntu, I don't think we need these considerations and comments and things about what other OSs support, right?

docker/linux/Dockerfile
```diff
@@ -1,15 +1,29 @@
-FROM debian:11-slim
+FROM ubuntu:22.04
 ENV ASPNETCORE_URLS=http://+:80 DOTNET_RUNNING_IN_CONTAINER=true
+
+# .NET 8 runtime dependencies.
+#
+# libc6, libgcc-s1, libgssapi-krb5-2, libstdc++6 and zlib1g were dropped from
```
xwipeoutx: More redundant commenting IMHO

source/Octopus.Tentacle.CommonTestUtils/PlatformDetection.cs
```diff
@@ -8,5 +8,44 @@ public static class PlatformDetection
...
+            if (!IsRunningOnNix) return true;
+
+            // dlopen by soname, so the dynamic linker's own search path is used rather
+            // than a guess at where the distribution puts its libraries. This is the
+            // same resolution the .NET Core 3.1 runtime performs.
+            foreach (var soname in new[] { "libssl.so.1.1", "libssl.so.1.0.0", "libssl.so.10" })
+            {
+                if (!NativeLibrary.TryLoad(soname, out var handle)) continue;
+
+                NativeLibrary.Free(handle);
```
xwipeoutx, line 35: For any future travellers, `dlopen` is ref counted, so this `Free` will not cause problems if it's already loaded.

todthomson:
@claude review once, and ignore the changes that are already part of the base PR: https://github.com/OctopusDeploy/OctopusTentacle/pull/1292/ (which you have already reviewed!).

docker/linux/Dockerfile
todthomson: Agree, fixed.

docker/linux/install-scripts/install-docker.sh
todthomson: Agree, fixed.

docker/linux/install-scripts/install-docker.sh
todthomson: Agree, fixed.

source/Octopus.Tentacle.CommonTestUtils/PlatformDetection.cs
todthomson: Resolving, as this is just a comment.

xwipeoutx (approved):
Approved - I've checked most of this via #1292

PR #1295, opened 2026-09-09
Checking certificate

### LINKED ISSUE #1296: Tentacle should confirm operator intent before trusting an unvalidated Octopus Server certificate
When a Tentacle is registered or configured to poll over a WebSocket address, it first probes
the Octopus Server over TLS to learn the certificate thumbprint it should trust from then on.
That probe deliberately accepts whatever certificate is presented, because the thumbprint cannot
be read from a handshake we refuse to complete — this is the moment the trust relationship is
established, so there is nothing to validate the server against yet.

Today the thumbprint learned this way is pinned with no further checks and no indication to the
operator. We are tightening this so that trusting a certificate which does not stand up on its
own becomes a deliberate, visible choice rather than the default.

## What's changing

Registration now takes account of standard certificate validation (chain of trust and host name)
before pinning a thumbprint:

- **The certificate validates** — the thumbprint is trusted exactly as it is today. No prompt,
  no change in behaviour.
- **The certificate does not validate** — Tentacle reports the thumbprint and why validation
  failed, then asks the operator to confirm before trusting it.
- **There is nobody to ask** (scripted or pipeline registration) — the command fails with a
  message naming the thumbprint and explaining how to proceed deliberately.
- **A new `--dangerously-skip-certificate-validation` argument** trusts the certificate without
  prompting, for unattended registration against a server whose certificate cannot be validated
  — a self-signed certificate, for example.

Affected commands: `register-with`, `register-worker`, `register-k8s-cluster`,
`register-k8s-target`, `register-k8s-worker`, and `poll-server`.

This applies only to WebSocket addresses (`--server-web-socket=wss://...`). Registration over the
Octopus Server's communications port takes its thumbprint from the authenticated Octopus API
rather than from the probe, and is unaffected.

## Breaking change

Unattended registration over a WebSocket address against a server whose certificate does not
chain to a trust anchor installed on the Tentacle machine will now fail instead of silently
trusting whatever it is given.

To keep those registrations working, either:

- install the issuing CA certificate on the Tentacle machine so the server's certificate
  validates, which is the better option because the thumbprint is then genuinely authenticated; or
- verify the thumbprint against the one shown by your Octopus Server and add
  `--dangerously-skip-certificate-validation` to the command.

Registrations against a server with a certificate from a public or internal CA are unaffected, as
are all interactive registrations, which now simply prompt.

sburmanoctopus:
# Background

When registering or configuring a polling Tentacle against a WebSocket address, `OctopusServerChecker` probes the Octopus Server over TLS to learn the thumbprint to trust. The probe accepts any certificate — it has to, since the thumbprint can't be read from a handshake we refuse to complete — but the harvested thumbprint was then pinned with no further checks and nothing shown to the operator.

Establishing trust this way is intentional and documented, and exploiting it needs an attacker already positioned between the Tentacle and the Server, so this isn't treated as a vulnerability. 

# Results

Fixes LEV-1806

Fixes https://github.com/OctopusDeploy/OctopusTentacle/issues/1296

The probe now records the `SslPolicyErrors` alongside the thumbprint, and registration acts on it:

- Certificate validates → trusted as before, no prompt.
- Doesn't validate → operator is shown the thumbprint and the failure reason, and must confirm.
- No operator to ask (`Console.IsInputRedirected`) → fails with a message naming the thumbprint and
  how to proceed.
- New `--dangerously-skip-certificate-validation` → trusts without prompting, for unattended runs.

Only the WebSocket path changes. Comms-port registration takes its thumbprint from the authenticated
Octopus API, so it never relied on the probe.

**Breaking change:** unattended WebSocket registration against a server whose certificate doesn't
chain to a local trust anchor now fails instead of silently trusting it. Intended for the next set
of breaking changes.

Also fixes a real defect: `ServicePointManager.ServerCertificateValidationCallback` is process-wide
and was left installed after the probe. It's now restored in a `finally`.

## Testing
We wrote tests for this functionality. This did mean having to pass down classes through the layers so we could mock these objects.

We manually ran the "register-with" command with a couple of these scenarios to ensure registration behaved as expected.

# Pre-requisites

- [x] I have read How we use GitHub Issues for help deciding when and where it's appropriate to make an issue.
- [x] I have considered informing or consulting the right people, according to the ownership map.

### DIFF
--- a/source/Octopus.Tentacle/Commands/PollCommand.cs
+++ b/source/Octopus.Tentacle/Commands/PollCommand.cs
@@ -24,11 +24,13 @@ namespace Octopus.Tentacle.Commands
         readonly IOctopusClientInitializer octopusClientInitializer;
         readonly ISystemLog log;
         readonly IApplicationInstanceSelector selector;
+        readonly IServerCertificateTrustConfirmation serverCertificateTrustConfirmation;
         readonly ApiEndpointOptions api;
         int? serverCommsPort = null;
         string serverWebSocketAddress = null!;
         string serverCommsAddress = null!;
         bool reuseThumbprint;
+        string? serverWebSocketThumbprint;
 
         public PollCommand(Lazy<IWritableTentacleConfiguration> configuration,
                            ISystemLog log,
@@ -36,7 +38,8 @@ namespace Octopus.Tentacle.Commands
                            Lazy<IOctopusServerChecker> octopusServerChecker,
                            IProxyConfigParser proxyConfig,
                            IOctopusClientInitializer octopusClientInitializer,
-                           ILogFileOnlyLogger logFileOnlyLogger)
+                           ILogFileOnlyLogger logFileOnlyLogger,
+                           IServerCertificateTrustConfirmation serverCertificateTrustConfirmation)
             : base(selector, log, logFileOnlyLogger)
         {
             this.configuration = configuration;
@@ -45,6 +48,7 @@ namespace Octopus.Tentacle.Commands
             this.octopusClientInitializer = octopusClientInitializer;
             this.log = log;
             this.selector = selector;
+            this.serverCertificateTrustConfirmation = serverCertificateTrustConfirmation;
 
             // purposefully not adding this with AddOptionSet because
             // we only want to validate it if reuseThumbprint is false.
@@ -54,6 +58,7 @@ namespace Octopus.Tentacle.Commands
             Options.Add("server-comms-port=", "The comms port on the Octopus Server; the default is " + DefaultServerCommsPort + ". If specified, this will take precedence over any port number in server-comms-address.", s => serverCommsPort = int.Parse(s));
             Options.Add("server-web-socket=", "When using active communication over websockets, the address of the Octopus Server, eg 'wss://example.com/OctopusComms'. Refer to http://g.octopushq.com/WebSocketComms", s => serverWebSocketAddress = s);
             Options.Add("reuse-server-thumbprint", "Reuse the Server Thumbprint from the first trusted server instance currently configured", _ => reuseThumbprint = true);
+            Options.Add("server-web-socket-thumbprint=", "When using active communication over websockets, the thumbprint of your server's websockets certificate, eg 'AB1C2D...'. This is the SSL certificate configured wherever TLS is terminated for the websockets endpoint, not the Octopus Server's Tentacle Communications certificate. When supplied the certificate is trusted only if its thumbprint matches, which allows registering against an endpoint whose certificate cannot otherwise be validated (for example a self-signed certificate) without prompting.", s => serverWebSocketThumbprint = s);
         }
 
         protected override void Start()
@@ -68,9 +73,16 @@ namespace Octopus.Tentacle.Commands
             {
                 api.Validate();
             }
+            else if (!string.IsNullOrWhiteSpace(serverWebSocketThumbprint))
+            {
+                throw new ControlledFailureException("Option --server-web-socket-thumbprint cannot be used with --reuse-server-thumbprint, which takes the thumbprint from an already trusted server instead of checking the one the server presents.");
+            }
 
             if (!string.IsNullOrEmpty(serverWebSocketAddress) && !string.IsNullOrEmpty(serverCommsAddress))
                 throw new ControlledFailureException("Please specify a --server-web-socket, or a --server-comms-address - not both.");
+
+            if (!string.IsNullOrWhiteSpace(serverWebSocketThumbprint) && string.IsNullOrWhiteSpace(serverWebSocketAddress))
+                throw new ControlledFailureException("Option --server-web-socket-thumbprint can only be used with --server-web-socket.");
 
             var serverAddress = GetAddress();
 
@@ -100,7 +112,7 @@ namespace Octopus.Tentacle.Commands
             //if we are on a polling tentacle with a polling proxy set up, use the api through that proxy
             var proxyOverride = proxyConfig.ParseToWebProxy(configuration.Value.PollingProxyConfiguration);
 
-            var sslThumbprint = octopusServerChecker.Value.CheckServerCommunicationsIsOpen(serverAddress, proxyOverride);
+            var serverCheckResult = octopusServerChecker.Value.CheckServerCommunicationsIsOpen(serverAddress, proxyOverride);
 
             log.Info($"Configuring Tentacle to poll the server at {api.ServerUri}");
 
@@ -108,7 +120,7 @@ namespace Octopus.Tentacle.Commands
 
             var repository = new OctopusAsyncRepository(client);
 
-            var serverThumbprint = await GetServerThumbprint(repository, serverAddress, sslThumbprint);
+            var serverThumbprint = await GetServerThumbprint(repository, serverAddress, serverCheckResult);
 
             var alreadyConfiguredServerInCluster = GetAlreadyConfiguredServerInCluster(serverThumbprint);
 
@@ -148,14 +160,17 @@ namespace Octopus.Tentacle.Commands
             return pollingServerConfiguration;
         }
 
-        async Task<string> GetServerThumbprint(IOctopusAsyncRepository repository, Uri serverAddress, string sslThumbprint)
+        async Task<string> GetServerThumbprint(IOctopusAsyncRepository repository, Uri serverAddress, OctopusServerCommunicationsCheckResult serverCheckResult)
         {
             if (serverAddress != null && ServiceEndPoint.IsWebSocketAddress(serverAddress))
             {
-                if (sslThumbprint == null)
-                    throw new ControlledFailureException($"Could not determine thumbprint of the SSL Certificate at {serverAddress}");
-                return sslThumbprint;
+                serverCertificateTrustConfirmation.EnsureCertificateIsTrusted(serverAddress, serverCheckResult, serverWebSocketThumbprint);
+
+                return serverCheckResult.Thumbprint;
             }
+
+            // Every other address takes its thumbprint from the authenticated Octopus API rather than from the probe,
+            // so there is nothing here for the operator to vouch for.
             return (await repository.CertificateConfiguration.GetOctopusCertificate()).Thumbprint;
         }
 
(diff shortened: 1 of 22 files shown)


xwipeoutx:
This needs a commit with the comment `+semver: major` - or it needs to be in the PR title. See the readme.

source/Octopus.Tentacle/Communications/ConsolePrompt.cs
```diff
@@ -0,0 +1,23 @@
+using System;
+
+namespace Octopus.Tentacle.Communications
+{
+    public class ConsoleCertificateTrustPrompt : ICertificateTrustPrompt
```
xwipeoutx: [nit] given the question comes in via the method, I think this is a general prompt, not a certificate trust prompt

source/Octopus.Tentacle/Communications/OctopusServerChecker.cs
```diff
@@ -30,57 +31,71 @@ public string CheckServerCommunicationsIsOpen(Uri serverAddress, IWebProxy? prox
...
+            var policyErrors = SslPolicyErrors.None;
+
+            var previousValidationCallbackToRestore = ServicePointManager.ServerCertificateValidationCallback;
             ServicePointManager.ServerCertificateValidationCallback = (sender, certificate, chain, errors) =>
             {
                 thumbprint = (certificate as X509Certificate2)?.Thumbprint;
+                policyErrors = errors;
+
+                // We have to accept the certificate to complete the handshake, otherwise we never get to read its thumbprint. 
                 return true;
```
xwipeoutx, line 43: I would rather this part not `return true` if the cert failed and they haven't put in the `--skip...` CLI.  Fail as early as we can.

source/Octopus.Tentacle/Commands/RegisterMachineCommandBase.cs
```diff
@@ -201,15 +206,22 @@ async Task RegisterMachine(IOctopusSystemAsyncRepository systemRepository, IOcto
...
                     throw new Exception($"Could not determine thumbprint of the SSL Certificate at {serverAddress}");
-                return sslThumbprint;
+                }
+
+                serverCertificateTrustConfirmation.EnsureCertificateIsTrusted(serverAddress, serverCheckResult, skipCertificateValidation);
+
+                return serverCheckResult.Thumbprint;
             }
 
+            // Every other address takes its thumbprint from the authenticated Octopus API rather than from the probe,
```
xwipeoutx, line 226: I don't really understand the _why_ here - why is web sockets different? Feels like bugs and vulnerabilities waiting to happen.

There's probably a reason, and I'd like us to nab it now while we're in major-version-and-security-land

source/Octopus.Tentacle/Commands/PollCommand.cs
```diff
@@ -54,6 +58,7 @@ public PollCommand(Lazy<IWritableTentacleConfiguration> configuration,
             Options.Add("server-comms-port=", "The comms port on the Octopus Server; the default is " + DefaultServerCommsPort + ". If specified, this will take precedence over any port number in server-comms-address.", s => serverCommsPort = int.Parse(s));
             Options.Add("server-web-socket=", "When using active communication over websockets, the address of the Octopus Server, eg 'wss://example.com/OctopusComms'. Refer to http://g.octopushq.com/WebSocketComms", s => serverWebSocketAddress = s);
             Options.Add("reuse-server-thumbprint", "Reuse the Server Thumbprint from the first trusted server instance currently configured", _ => reuseThumbprint = true);
+            Options.Add(ServerCertificateTrustConfirmation.SkipValidationArgumentName, ServerCertificateTrustConfirmation.SkipValidationArgumentDescription, _ => skipCertificateValidation = true);
```
xwipeoutx: I don't like this constant, made it way harder to track down the name

source/Octopus.Tentacle/Communications/ServerCertificateTrustConfirmation.cs
```diff
@@ -0,0 +1,61 @@
...
+        readonly ISystemLog log;
+        readonly ICertificateTrustPrompt prompt;
+
+        public ServerCertificateTrustConfirmation(ISystemLog log, ICertificateTrustPrompt prompt)
+        {
+            this.log = log;
+            this.prompt = prompt;
+        }
+
+        public void EnsureCertificateIsTrusted(Uri serverAddress, OctopusServerCommunicationsCheckResult checkResult, bool skipValidation)
```
xwipeoutx: Hmm, I know the linear ticket said skip validation, but I think we can be a little stronger here and say `--server-certificate-thumbprint=<X>` instead.  My concern is automated scripts (say a k8s cluster of elastic workers) all having `--dangerously-skip-certificate-validation` by default, leaving any of them open for this MITM.

source/Octopus.Tentacle/Commands/RegisterMachineCommandBase.cs
xwipeoutx: I threw Claude at it - the websocket endpoint cert could very well be terminated before it even gets to server - like, IIS or caddy or nginx would be doing that.  The thumbprint reported in Server is specifically the comms thumbprint.  They're just different.

Asking the user to just "confirm it's the same as octopus server" is going to mislead them.  I think we need to be clearer about it.

source/Octopus.Tentacle/Communications/ServerCertificateTrustConfirmation.cs
xwipeoutx: Of course we can still show them the thumbprint we received, so they can still copy/paste it in for the unattended case.  And we can still `y/N` prompt them in the attended case (with a warning to check the thumbprint of course)

xwipeoutx: Hmm it's also becoming clear to me that this flag is only for websockets - so it should be named as such.

We have `--server-web-socket=wss://` I reckon we also need `--server-web-socket-thumbprint=<X>` to cover all this, and we're good, yeah?

There's the extra fun case of when they use `ws://` instead of `wss://` - do we even allow that?

xwipeoutx (commented):
I think we need to reshape this a fair bit - further scrutiny and AI interrogation is showing me the original suggested fix had some gaps.

Happy to jump on a call.

source/Octopus.Tentacle/Communications/ConsolePrompt.cs
sburmanoctopus: Agreed. Changed

source/Octopus.Tentacle/Communications/OctopusServerChecker.cs
sburmanoctopus: I looked into this, and making it return false means we end up throwing an exception rather than returning a nice message.

I'm also not a massive fan of putting the user prompt in here either. 

We had a discussion, and decided to leave it as is.

source/Octopus.Tentacle/Commands/PollCommand.cs
sburmanoctopus: I was on the fence on this one. If we remove the constants, it means some repeats, but it fits everything else.

I'm happy to remove them 🙂

source/Octopus.Tentacle/Communications/ServerCertificateTrustConfirmation.cs
sburmanoctopus: Sounds good. I'll go with `--server-web-socket-thumbprint=<X>`

source/Octopus.Tentacle/Commands/RegisterMachineCommandBase.cs
sburmanoctopus: Rewording was done on the prompts and error messages.

source/Octopus.Tentacle/Commands/PollCommand.cs
```diff
@@ -58,7 +58,7 @@ public PollCommand(Lazy<IWritableTentacleConfiguration> configuration,
...
             Options.Add("reuse-server-thumbprint", "Reuse the Server Thumbprint from the first trusted server instance currently configured", _ => reuseThumbprint = true);
-            Options.Add(ServerCertificateTrustConfirmation.SkipValidationArgumentName, ServerCertificateTrustConfirmation.SkipValidationArgumentDescription, _ => skipCertificateValidation = true);
+            Options.Add("server-web-socket-thumbprint=", "When using active communication over websockets, the thumbprint of your server's websockets certificate, eg 'AB1C2D...'. This is the SSL certificate configured wherever TLS is terminated for the websockets endpoint, not the Octopus Server's own certificate. When supplied the certificate is trusted only if its thumbprint matches, which allows registering against an endpoint whose certificate cannot otherwise be validated (for example a self-signed certificate) without prompting.", s => serverWebSocketThumbprint = s);
```
xwipeoutx: > "not the Octopus Server's own certificate"

In all likelihood, it is the same SSL cert for HTTPS as WSS - this is kinda misleading.

perhaps "not the Octopus Server's Tentacle Communications certificate"?

See [this docs page](https://octopus.com/docs/security/octopus-tentacle-communication/custom-certificates-with-octopus-server-and-tentacle/) - which is clearly about the TCP transport not websockets

source/Octopus.Tentacle/Communications/ServerCertificateTrustConfirmation.cs
```diff
@@ -5,50 +5,56 @@ namespace Octopus.Tentacle.Communications
...
+
             log.Warn(problem);
 
             if (!prompt.CanPrompt)
             {
                 throw new ControlledFailureException(
-                    $"{problem} {CannotConfirmExplanation} " +
-                    $"Verify the thumbprint against the one shown by your Octopus Server, then re-run this command with --{SkipValidationArgumentName} to trust it.");
+                    $"{problem} Because this certificate could not be validated, it is not possible to confirm that it belongs to your Octopus Server rather than to an attacker positioned between this machine and the server. " +
+                    "Check that this thumbprint matches your server's websockets certificate - the one configured wherever TLS is terminated for the websockets endpoint, such as IIS, HTTP.SYS or a reverse proxy, and not the Octopus Server's own certificate. " +
```
xwipeoutx: I'd just leave out the "and not..." bit.

xwipeoutx (commented):
A few suggested wording changes, and comment.

I won't approve until I see a major version rev.

source/Octopus.Tentacle/Commands/PollCommand.cs
sburmanoctopus: Changed to `not the Octopus Server's Tentacle Communications certificate` in all locations that had `not the Octopus Server's own certificate`

source/Octopus.Tentacle/Communications/ServerCertificateTrustConfirmation.cs
sburmanoctopus: Done

xwipeoutx (approved):
Approved - as discussed, let's not squash-merge this one, I'm not sure the major version bump survives it

PR #1299, opened 2026-09-14
Support timing out waiting for remote scripts to cancel, allowing us to abandon the remote script which is left running on Tentacle.

LukeButters:
## Summary

Older versions of tentacle have issues in which cancellation way not work. Since some operations can be abandoned, this PR allows the TentacleClient to walk away from scripts that refuse to cancel.

It works by allowing callers to set `TimeSpan scriptCancellationTimeoutBeforeAbandoning`, once the ExecuteScript CancelltationToken is triggered AND `scriptCancellationTimeoutBeforeAbandoning` time has passed since the first attempt to cancel the script THEN the TentacleClient returns (throwing an `OperationCancelledException`) without waiting for the Remote Script to complete.

Fixes: LEV-1900

## Test plan
- [x] `dotnet test Octopus.Tentacle.Client.Tests` — all 38 tests pass, including the 2 new ones
- [x] `dotnet build` on `Octopus.Tentacle.Client`, `Octopus.Tentacle.Client.Tests`, and `Octopus.Tentacle.Tests.Integration` — all succeed
- [ ] New integration test (`WhenALongRunningScriptCannotBeCancelled_TheClientAbandonsObservingAfterTheTimeout`) — **could not be run locally**: even the pre-existing, unmodified `CanRunScript` integration test hangs/times out in this sandboxed macOS environment (likely a real-Tentacle-process/networking sandbox restriction, unrelated to this change). Needs verification in CI/TeamCity.


source/Octopus.Tentacle.Client/Observability/NonThrowingTentacleClientObserver.cs
```diff
@@ -60,5 +61,17 @@ public void ExecuteScriptCompleted(ClientOperationMetrics clientOperationMetrics
...
+
+        public void ScriptCancellationTimedOut(ScriptCancellationTimedOutEvent scriptCancellationTimedOutEvent)
+        {
+            try
+            {
+                inner.ScriptCancellationTimedOut(scriptCancellationTimedOutEvent);
+            }
+            catch
+            {
+                // Swallow - no logger is available for this notification and the decorator must not throw.
```
gb-8: Logger can easily be made available. We should log this.

gb-8 (approved):
✅ Pre-approved. I left a comment, and there were existing copilot comments which seem valid and need addressing before merging.

source/Octopus.Tentacle.Client/Observability/NonThrowingTentacleClientObserver.cs
LukeButters: done

PR #1305, opened 2026-09-21
Stop deriving the Tentacle.config encryption key from /etc/machine-id on Linux

gb-8:
# Background

[LEV-1171](https://linear.app/octopus/issue/LEV-1171/use-a-fresh-etcmachine-id-for-tentacle-in-docker) (Security, from support case FD-601).

On Linux, `ProtectionLevel.MachineKey` values in `tentacle.config` (the Tentacle certificate with its private key, and the proxy passwords) were AES-encrypted with the 32 hex characters of `/etc/machine-id` used directly as the key, and one fixed IV for every value. The official Linux image bakes a populated `/etc/machine-id` into the image, so every container started from a given tag shares the encryption key, and anyone with a copy of `tentacle.config` from a container can pull the same public image and decrypt the certificate. On any host the machine-id is world-readable by design, and systemd's `machine-id(5)` says applications must never use it directly as a key.

# Results

Design, in four commits so the Docker one can be dropped independently:

1. **Crypto + migration + tests.** Values are now always encrypted with the random per-machine key Tentacle already knew how to generate (`LinuxGeneratedMachineKey`), with a fresh IV per value, stored as `$OctopusMachineKeyV1$` + base64(iv || ciphertext). Values without the prefix were written by earlier versions and are decrypted exactly the way those versions did it (machine-id key, then generated key), so every existing install keeps working after upgrade. A prefixed value is only ever tried with the generated key, so a wrong key can no longer "succeed" by chance; legacy decryption additionally rejects output that is not valid UTF-8, which closes most of the 1-in-256 wrong-key/valid-padding gap. `tentacle agent` re-encrypts the three protected settings on start when they are in the old format (`ProtectedSettingsMigrator`), logging a warning instead of failing when it cannot (read-only config). Reads never rewrite. The key file is created mode 0600, an existing world-readable one is tightened once, and it now follows `TentacleMachineConfigurationHomeDirectory` like the instance registry, so non-root installs can write it (the old fixed path is still read for legacy values). Windows/DPAPI and the Kubernetes agent's ConfigMap store are untouched.
2. **Linux package tests** assert the new format, key file mode and cross-process readability on every distro; a new `testing/linux-packages/test-upgrade-from-previous.sh` upgrades from the latest released package and checks identity survives and migration runs exactly once.
3. **Docker image** ships with an empty `/etc/machine-id` and no `/var/lib/dbus/machine-id`; `docker/readme.md` documents persistence and recovery.
4. **Docker image tests** cover the machine-id removal, per-container keys, and stop/start plus destroy/re-create on a persisted `/etc/octopus` volume keeping the same thumbprint without re-registering.

Things to decide (from the handover's open questions):

- **Downgrade:** after migration, values are prefixed and an older Tentacle cannot read them. Recovery is `tentacle new-certificate` + re-trust, or restoring the pre-upgrade `tentacle.config`. Same class of break as any config format change; needs a release note.
- **Docker upgrade with a persisted volume (commit 3):** a container whose `/etc/octopus` was written by an older image and never started on a version that re-encrypts cannot read its certificate on the new image, because the old image's machine-id is gone. Today this already breaks whenever the Ubuntu base layer's machine-id changes between builds; with commit 3 it is deterministic but happens at most once, after which the key lives on the volume. The legacy-decrypt error message now names this scenario and the recovery. If that trade-off is not acceptable, drop commit 3 and the machine-id assertions in commit 4.
- **Non-root `show-thumbprint` against a root-owned instance** now needs read access to a 0600 root-owned key file, where the machine-id was world-readable before.

## Before

```xml
<set key="Tentacle.Certificate">2qbSDqJXa0/Bz1SzohM21m3dg3mIPBjj4mgb...</set>
```
Decryptable by anyone with the image's `/etc/machine-id`.

## After

```xml
<set key="Tentacle.Certificate">$OctopusMachineKeyV1$Jm3k...</set>
```
```
Re-encrypted the protected setting 'Tentacle.Certificate' with the machine key generated for this machine.
```
Key in `/etc/octopus/machinekey`, `-rw-------`, different in every container.

# Reducing risk

The risk in this change is an upgraded Tentacle that can no longer read its own certificate. Everything below is aimed at that, in order of how directly it proves it.

## Automated tests added

**Upgrade compatibility (the proof that existing installs keep working)**

`LegacyCiphertextCompatibilityFixture` pins ciphertext produced by the pre-change encryptor (commit 7b1504e, captured by running that code, not regenerated with this code) for five plaintexts: short text, a serialised bool, a serialised int, unicode with symbols, and a certificate-sized value. Each is checked against the encryptor exactly as `MachineKeyEncryptor.CreateLinuxEncryptor` composes it in production:
- values written with the machine-id key still decrypt
- values written with the generated key on a host without a machine-id still decrypt
- values written with the generated key still decrypt after a machine-id has since appeared on the host
- every legacy value is reported as needing re-encryption
- once re-encrypted, a value decrypts with no machine-id at all and with a different machine-id
- an install that relocated its configuration home still reads values written with the key at `/etc/octopus/machinekey`, and writes its new key into the relocated home
- a value encrypted on this machine cannot be read by another machine that has the same image machine-id but a different generated key (the attack this fixes)

**Encryptor behaviour** (`LinuxMachineKeyEncryptorFixture`, rewritten)
- round trip, including the empty string
- new values carry the version prefix and are not reported as legacy; unprefixed values are
- the same plaintext encrypts to different ciphertext each time (per-value IV)
- encrypting never loads a legacy key
- decrypting a prefixed value never loads a legacy key
- decrypting a prefixed value with the wrong key throws and does not fall back to a legacy key that could have decrypted it
- tampered and truncated prefixed values throw
- a corrupt or unloadable key makes encryption throw with the underlying reason in the message
- legacy values decrypt with the legacy keys, in order, first success wins, later keys untouched
- a legacy value no legacy key can decrypt throws
- legacy values are only tried with the legacy list, never the primary key (documents why production puts the generated key in both)
- legacy decryption with a wrong key that happens to produce valid padding is rejected as invalid UTF-8

**Generated key file** (`LinuxGeneratedMachineKeyFixture`, new; the old file of that name tested the machine-id key and is renamed `LinuxMachineIdKeyFixture`)
- a missing file is generated in the `base64(key).base64(iv)` format earlier versions read, via the owner-only write, with a 32-byte key
- two machines get different keys
- an existing file loads without writing anything
- an existing world-readable file is tightened and logged at Info; an already-restricted one logs nothing
- a permission change that fails (read-only volume) is logged at Verbose and the key still loads; permissions are only checked once per instance
- corrupt files (bad base64, no separator, empty) throw naming the file
- a legacy-only location does not create a key when the file is missing
- path resolution: `/etc/octopus/machinekey` by default, the relocated machine configuration home when set, `$TentacleHome` for the Kubernetes agent, and an explicit path over all of them

**Generated key file on the real filesystem** (`LinuxGeneratedMachineKeyFileFixture`, Linux only)
- a new key file is created mode 0600 with the expected contents
- the same key loads next time
- a file left 0644 by an earlier version is tightened to 0600 without changing the key
- an already 0600 file is left alone, including its modification time

**Store re-encryption** (`XmlFileKeyValueStoreReEncryptionFixture`)
- a legacy string is re-encrypted in place, saved, and reads back unchanged from a fresh store
- serialised bool, int, enum and object values keep their serialised form so typed reads still work
- a value already in the current format, a missing value and a blank value are left alone and the file is not touched
- the list of protected settings is exactly the certificate and the two proxy passwords, so unprotected values are never rewritten
- a legacy value that cannot be decrypted throws and leaves the file untouched

**Start-up migration** (`ProtectedSettingsMigratorFixture`)
- all three protected settings written by the old scheme are re-encrypted and read back with the original plaintext, with one Info line each
- unprotected settings are untouched
- running again is a no-op (file identical, nothing logged); a fresh install is a no-op
- settings not present are skipped
- a setting that cannot be rewritten logs one warning and the remaining settings are still processed
- a store that cannot re-encrypt (the Kubernetes ConfigMap store) and a missing writable configuration are ignored without error

**Agent** (`RunAgentCommandFixture`)
- re-encryption runs before Halibut starts
- an exception from re-encryption does not stop the agent starting

**Command line, end to end** (`TentacleCommandLineTests`, Linux only)
- `create-instance` then `new-certificate` writes a prefixed certificate, creates the key file in the machine configuration home with mode 0600, and a second process running `show-thumbprint` reads back the recorded thumbprint

**Existing fixtures made hermetic**: `XmlFileKeyValueStoreFixture`, its round-trip fixtures, and `ProxyConfigurationCommandFixture` now inject `TestMachineKeyEncryptor` (real DPAPI on Windows, in-memory key elsewhere) so no unit test creates or depends on `/etc/octopus/machinekey`.

**Package and image scripts**
- `linux-packages/test-scripts/test-linux-package.sh` (runs on every distro in the matrix): after install, `create-instance` and `new-certificate`; the certificate is prefixed, `/etc/octopus/machinekey` is mode 600, and `show-thumbprint` in a new process matches the recorded thumbprint
- `testing/linux-packages/test-upgrade-from-previous.sh` (new, ubuntu:24.04 and redhat/ubi9): install the latest released package, configure certificate and both proxy passwords, confirm they are in the old format, upgrade to the built package, `show-thumbprint` unchanged and nothing rewritten, run the agent and confirm exactly three re-encryption log lines, all values prefixed, thumbprint unchanged, key file 600, and a second agent run logs nothing
- `testing/docker-linux/build-and-test-linux-docker-image.sh` smoke: `/etc/machine-id` empty and `/var/lib/dbus/machine-id` absent; a configured instance has a prefixed certificate, a 600 key file, and `show-thumbprint` matches; two containers generate different keys
- same script, e2e: the listening Tentacle keeps `/etc/octopus` on a named volume; after `docker stop`/`start` and after `docker rm` and re-create on that volume the thumbprint is unchanged, the server reports it healthy, and it did not register a second time

## Manual test plan

Run against a real Octopus Server. "Healthy" means the health check passes and a script step deploys.

*Fresh installs*
1. deb on Ubuntu 24.04 and rpm on RHEL 9: install, create instance, new certificate, register listening and polling, deploy. Certificate prefixed in `tentacle.config`, key file `-rw-------`, `show-thumbprint` matches the Server.
2. Fresh Linux container, listening and polling, with `-v tentacle:/etc/octopus`: register and deploy. A second container from the same image has a different key file. `docker stop && start` stays healthy. `docker rm` and re-create on the same volume stays healthy with the same thumbprint and no re-registration.
3. Kubernetes agent fresh Helm install as target and worker: healthy; `$TentacleHome/machinekey` exists, 0600 where the storage class enforces modes; OpenShift-style random UID still starts.

*Upgrades (the critical part)*
4. In-place package upgrade: install the current release, register and deploy, set a proxy password and a polling proxy password, upgrade, restart. Still healthy without re-trust, thumbprint unchanged, all three values prefixed, proxied polling still connects, migration logged once and silent on the next restart.
5. Server-driven "Upgrade Tentacle" from the Octopus UI on a Linux target running the previous version: healthy afterwards.
6. Host without `/etc/machine-id` (for example `debian:bookworm-slim`) on the previous version, so it used the generated key: upgrade decrypts via the legacy generated-key source, then migrates.
7. Docker upgrade with a persisted `/etc/octopus` from the current `octopusdeploy/tentacle` tag to this image. Expected: fails to read the certificate because the old image's machine-id is gone; confirm the new error message names the scenario and that `tentacle new-certificate` plus re-trust recovers. Record the result for the release note.
8. Kubernetes agent upgrade from the current chart: healthy, file-based config migrated, ConfigMap values unaffected.
9. Read-only config: a container with `-v tentacle:/etc/octopus:ro` after being configured read-write on the previous image starts, logs the migration warning, and stays healthy.

*Negative and edge cases*
10. Downgrade after migration: confirm the failure is understandable and `new-certificate` plus re-trust recovers. Feed into release notes.
11. Non-root `show-thumbprint` against a root-owned instance: fails on the 0600 key file where the machine-id was world-readable. Decide whether acceptable and document.
12. Corrupt `machinekey`: the error names the file and the agent does not crash-loop silently.
13. Windows sanity: unit tests on net48 and net8.0-windows, one fresh install registers and deploys.
14. `show-configuration` and `configure --reset-trust` still work after migration.

# How to review this PR

- Start with `LegacyCiphertextCompatibilityFixture`: the ciphertext in it was captured by running the *pre-change* encryptor (commit 7b1504e) against a fixed machine-id and key file, and the new code must decrypt it. That is the upgrade-safety proof.
- Then `LinuxMachineKeyEncryptor` (format, no fall-through for prefixed values) and `ProtectedSettingsMigrator` + `FlatDictionaryKeyValueStore.ReEncryptIfLegacy` (idempotent, failure-tolerant migration).
- `testing/linux-packages/test-upgrade-from-previous.sh` is the end-to-end upgrade check.

Verification status, honestly: the sandbox this was written in cannot reach feedz.io, so the full solution was not built here. The changed production files and all new/changed fixtures were compiled with warnings-as-errors for net8.0, net8.0-windows and net48 in a harness that links the real source files, and the 142 crypto/store/migration tests pass on net8.0 (including the real-filesystem 0600 tests). Not run: the full `Octopus.Tentacle.Tests` suite, the integration test added to `TentacleCommandLineTests`, the package/Docker scripts (no Docker available), and anything on Windows. CI should be treated as the first full run; the manual test plan above is still to do.

Quality :heavy_check_mark:
Focus areas: review the compatibility fixture and the migration failure handling; exploratory testing of an in-place upgrade on a Linux target with proxy passwords set; a Docker upgrade with a persisted `/etc/octopus` volume.

# Pre-requisites

https://claude.ai/code/session_01DQxynCbgxJiV6cYB6Byau7


source/Octopus.Tentacle.Core/Util/OctopusPhysicalFileSystem.cs
```diff
@@ -250,6 +250,56 @@ public void WriteAllText(string path, string contents)
             File.WriteAllText(path, contents);
         }
 
+#if NETFRAMEWORK
```
gb-8, line 253: Don't think we need this.

gb-8:
Closed and regenerated with fresh commits here:
https://github.com/OctopusDeploy/OctopusTentacle/pull/1307

Because Claude made the original commits under its own account, and CLAassistant doesn't like that.

